PR #1012, opened 2015-07-02
Use long string syntax to generate `config.lua` file.

### LINKED ISSUE #1006: Use long string syntax to generate config.lua file.
```
Now on Windows when config.lua file generate we can get string like
`php_dir = "c:\php"`
but this is not valid Lua string.
I suggest generate strings like
`php_dir = [[c:\php]]`

```

Original issue reported on code.google.com by `alexeyme...@gmail.com` on 29 Jun 2015 at 6:48

moteus:
Close #1006


richbreton:

PR #1016, opened 2015-07-02
Fix. Escape `<>` in all fields in CDR xml.

moteus:
When using redirect cdr also can contain e.g. `sip_redirect_contact_0` / `sip_redirected_to`.
I use `preg_replace_callback` to proceed all fields.
May be it more correct use `urlencode` because if we get xml with non UTF8 char we lost CDR.
```php
$xml_string = preg_replace_callback("/<([^><]+)>(.*?)<\/\g1>/",
	function ($matches) {
		// var_dump($matches);
		return '<' . $matches[1] . '>' .
			urlencode($matches[2]).
		'</' . $matches[1] . '>';
	},
	$xml_string
);
```

### DIFF
--- a/app/xml_cdr/v_xml_cdr_import.php
+++ b/app/xml_cdr/v_xml_cdr_import.php
@@ -80,7 +80,7 @@
 			global $debug;
 
 		//fix the xml by escaping the contents of <sip_full_XXX>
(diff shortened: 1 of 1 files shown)


richbreton:

PR #1017, opened 2015-07-02
Change. Does not use `default` dialplan/directory.

moteus:
I think it more consistent that it allows edit domain in single place.
Also on my Windows machine while installing all users/context creates not in `default` directory.
And I have problem because remove users did not work, `default.xml` file contain `$${v_domain}` variables.
Also when crete extension by default there real domain name not `default`.
I am not sure. May be better fix to support `default` domain but I see no reason to do that.


richbreton:

markjcrane:
Can you update the pull request as it currently has conflicts.

moteus:
update

PR #1021, opened 2015-07-07
Fix. Show registrations to sip profile with force-XXX-domain.

moteus:
When on profile set `force-register-db-domain` realm on xmlstatus is just IP but user is like 101@domain.name.


moteus:
To reproduce this also needs to use number aliases for extensions

richbreton:

PR #1023, opened 2015-07-07
Fix. Use extension to generate dial-string in xml_handler

### LINKED ISSUE #1022: Genarate dial-string for ext with alias
I have extensions whit number aliases.
I try use `xml_handler` to generate directory.
Problem whith this code
```Lua
dial_string = "{sip_invite_domain=" .. domain_name .. ",presence_id=" .. user .. "@" .. domain_name .. "}${sofia_contact(" .. user .. "@" .. domain_name .. ")}";
```
In this case `user` is a number alias but not user id.
The easest way to fix this is just use `${dialed_user}`
For now it works for me

moteus:
When uses number_alias contact should be searched by extension name.
Close #1022

### DIFF
--- a/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
+++ b/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
@@ -254,7 +254,7 @@
 								else
 									--set a default dial string
 										if (dial_string == null) then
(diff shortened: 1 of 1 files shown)


richbreton:

daniel-lucio:
This can be work-arounded with dialplan, by default Fusion does not look on aliases, but you can enable it

PR #1024, opened 2015-07-07
Update app_languages.php

zhulizhong:
translate the modules in chinese


markjcrane:
The pull request needs to be done to the open-source repository. As described here then we will set you up with a translation account. https://github.com/fusionpbx/open-source

PR #1030, opened 2015-07-09
Update contacts.php

daniel-lucio:
In some cases, the $_SESSION['groups'] array has a '' elemenl (null element). This makes PostgreSQL to break (Mysql/Mariadb are tolerant to this) at like 80 when doing the SQLsentence: .....where group_uuid in ('".implode("','", $user_group_uuids)."') ";  as it will create a query like: group_uuid in ('','XXXXXXX'.......) which it wont work.

### DIFF
--- a/app/contacts/contacts.php
+++ b/app/contacts/contacts.php
@@ -22,6 +22,7 @@
 
 	Contributor(s):
 	Mark J Crane <markjcrane@fusionpbx.com>
+	Luis Daniel Lucio Quiroz <dlucio@okay.com.mx>
 */
 require_once "root.php";
 require_once "resources/require.php";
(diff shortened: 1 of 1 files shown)


markjcrane:
Sorry this is a band-aid for an upgrade that needs to be done better.
select * from v_group_users the group_uuid should not be null
http://wiki.fusionpbx.com/index.php?title=Upgrade#Version_3.6_to_3.8
See this section "Version 3.6 to 3.8" ->   "Groups and Permissions"
Need to be logged in and run the following SQL and PHP and then maybe run Advanced -> Group Manager Restore default.

After your this patch will not be needed for a system that is upgraded properly.

PR #1053, opened 2015-07-22
Change. Use `require "resources.functions.config"` to load config file.

moteus:
Change. Move `script_dir` definition into config file.

This allows load config file only once.
It this ok then I also change all dofile that load modules such as database_handler/file_exists.

### DIFF
--- a/core/databases/app_defaults.php
+++ b/core/databases/app_defaults.php
@@ -24,6 +24,22 @@
 	Mark J Crane <markjcrane@fusionpbx.com>
 */
 
+global $IS_WINDOWS;
+
+if ($IS_WINDOWS == null) {
+	if (stristr(PHP_OS, 'WIN')) { $IS_WINDOWS = true; } else { $IS_WINDOWS = false; }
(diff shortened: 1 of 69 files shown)


markjcrane:
Reverted - My system couldn't follow the path and so it broke everything.

moteus:
Did you set `<param name="script-directory" value="$${base_dir}/scripts/?.lua"/>` in `lua.conf.xml`?

markjcrane:
But I likely  forgot to restart freeswitch. Your note should have gone in the commit message as this will break every system that upgrades until making that change and restarting.

What is the benefit to the change? It looks prettier than dofile but is there a performance gain?

moteus:
`dofile` load file each time but `require` load file only once. This is why I left handlers with dofile.
Also with `require` you can load other libraries (e.g. installed via LuaRocks) and code will be same.

markjcrane:
Tested again and it worked if you resubmit the pull request I will apply it.

PR #1066, opened 2015-07-28
Fix. Allow intercept only outbound calls.

moteus:
Fix. Use dbh only from config.
Fix. intercept channels also in `EARLY` state.

### DIFF
--- a/resources/install/scripts/intercept_group.lua
+++ b/resources/install/scripts/intercept_group.lua
@@ -111,7 +111,7 @@
 			sql = sql .. "AND extension = '"..caller_id_number.."'";
 			status = dbh:query(sql, function(row)
 				call_group = row.call_group;
-				freeswitch.consoleLog("NOTICE", "result "..call_group.."\n");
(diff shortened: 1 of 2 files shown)


resources/install/scripts/intercept_group.lua
```diff
@@ -119,39 +119,32 @@
...
 				extensions[x] = row.extension;
-				freeswitch.consoleLog("NOTICE", "result "..row.extension.."\n");
+				freeswitch.consoleLog("NOTICE", "[intercept_group] member "..row.extension.."\n");
 				x = x + 1;
 			end);
 
 		--connect to the database
-			if (file_exists(database_dir.."/core.db")) then
-				--dbh = freeswitch.Dbh("core:core"); -- when using sqlite
-				dbh = freeswitch.Dbh("sqlite://"..database_dir.."/core.db");
```
moteus: Is it intention or better remove this?
I saw such code in several places.
I use PgSQL, but after install FS I run it and it creates core.db
It may be quite tricky to figure out why inteecept mpt working without checkout source code

PR #1091, opened 2015-08-12
Fix. Allow intercept only outbound calls.

moteus:
Now it is possible intercept call from local extension to the world.

Fix. intercept channels also in EARLY state.

One of my softphones return 180 Ringing with SDP (no 183 Progress).
And FreeSWITCH shows such channel as `EARLY` not as `RINGING`)


moteus:
I think may be it worth keeping to direct intercept only (**[ext])?

PR #1093, opened 2015-08-12
Fix. Use number alias to forward call to VM when DND is on.

moteus:


moteus:
I will fix also Lua file today later.

moteus:
Please test work with memcache.
May be [here](https://github.com/fusionpbx/fusionpbx/blob/master/resources/install/scripts/do_not_disturb.lua#L146) also need use number_alias.
I never work with mod_memcache (I have problem with bulding it)

PR #1100, opened 2015-08-14
New resources dir inside the xml_cdr app

daniel-lucio:
This is the first test to do the billing lest intrusive, by adding the resource, i will put later code to include files there, and not in the code itself.  

This will benefit any 3th party developer.


markjcrane:
I think it would be better to create your own app/billing/cdr_import.php to import CDR in your own billing folder. That way your billing system can't break the CDR page like it did in the past and lost a project because of it.

PR #1116, opened 2015-08-20
Fix. Verify number-alias and user id.

moteus:
This replaces `inbound-reg-force-matching-username` profile option.
So if your extension has number-alias you can use
UserID=number-alias
AuthID=Extension

It also supports such check on INVITE not only on REGISTER.

### DIFF
--- a/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
+++ b/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
@@ -62,6 +62,20 @@
 			--all other directory actions: sip_auth, user_call 
 			--except for the action: group_call
 
+			local sip_auth_method = params:getHeader("sip_auth_method")
+			if sip_auth_method then
+				sip_auth_method = sip_auth_method:upper();
+			end
+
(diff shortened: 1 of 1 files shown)


resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
```diff
@@ -250,6 +264,15 @@
 								forward_no_answer_destination = row.forward_no_answer_destination;
 								do_not_disturb = row.do_not_disturb;
 
+							-- check matching UserID and AuthName
+								if sip_auth_method and (METHODS[sip_auth_method] or METHODS._ANY_) then
+									continue = (sip_from_user == user) and ((sip_from_number == user) or (sip_from_number == from_user))
```
moteus: I just think that this check should be 
`continue = (sip_from_user == user) and ((from_user == user) or (sip_from_number == from_user))`
to allow register with authid=extension for extension with number-alias

need tests

moteus: @markjcrane Did you test with this check?

PR #1125, opened 2015-08-25
Fix. Start record when b-leg answer.

moteus:
This fix problem when DID transfers to IVR.
After that user make direct-dial to extension.
IVR do transfer but a-leg already answered and
api_on_answer did not call.


moteus:
This also allows avoid record call that go to voicemail

PR #1129, opened 2015-08-26
Add. option to switch which value will be in presence_id extension or number_alias.

moteus:
Category: `Provision`
Subcategory: `number_as_presence_id`
Type: `boolean`
Default: `false`

To configure user dial_string use `NUMBER_AS_PRESENCE_ID` variable in `directory.lua` script.

Main incompatible changes is that presence_id for external number in call forward dial-string set to
extension itself not external number. This behavior consistent with follow me.
So you get same presence_id for follow_me/call_forward for same external number.
It allows work BLF/Intercept.

Also follow_me did not set presence_id for user because it should be already has one in dial-string.

I am not shure about name/category so please tell me your suggestions.

I do not use `settings` in `directory.lua` because it quite expensive way to get one option.
And with `load_balancing=true` directory did not use cache. I think it possible use settings when
`load_balancing=false`. Or write function that returns one specific option (e.g. `settings_get_value`).
### Todo
- [x] - Merge with #1147
- [x] - Fix. Allow only register with AuthID=ext and UserID=ext when using  `xml_handler["reg_as_number_alias"] = false`
- [x] - Fix. Allow only register with AuthID=ext and UserID=alias when using  `xml_handler["reg_as_number_alias"] = true`
- [x] - Test BLF for outbound calls with

``` Lua
xml_handler["number_as_presence_id"] = false;
xml_handler["reg_as_number_alias"] = true;
```


moteus:
One note. presence_id uses in intercept script to find out call.
So when user dial `**200` intercept script lookup presence_id like `200@domain`.
So to allow intercept works with number-alias presence_id have to be not extension but number alias.
(or make one more call like user_data in intercept extension)

markjcrane:
Committed changes to directory.lua in this commit.
https://github.com/fusionpbx/fusionpbx/commit/b807c7e15f9d1fb526bb030df7ea27ba1dc6d945

markjcrane:
number_as_presence_id added to default settings. 
https://github.com/fusionpbx/fusionpbx/commit/fef88f08730461de8ac3ce91c22218c65e2234da

moteus:
[This one](https://github.com/fusionpbx/fusionpbx/pull/1129/files?w=1#diff-1b7ce6b6eb1ad9b6eba1be104e3531d2R86) needs to be able do auth with `fs_path=true`.

Also `number_as_presence_id` also require other PHP code. In fact now it does not uses anywhere

PR #1139, opened 2015-09-03
Allow provisioning with internal addressbook and update snom320 tpl

joernott:
We wanted to populate the addressbook of our snom320 phones from the list of available extensions and found out, that the provided xml templates only provide an empty address book. Also, we didn't want to base the address book on the contacts. This commit adds another template variable internal_addressbook from the extensions table.

The snom320 template was modified to support the function keys (like the 7xx template) and populates the tbook array.

### DIFF
--- a/app/provision/resources/classes/provision.php
+++ b/app/provision/resources/classes/provision.php
@@ -661,6 +661,14 @@ include "root.php";
 					}
 					unset ($prep_statement);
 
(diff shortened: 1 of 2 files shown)


markjcrane:
Please follow the contributor instructions in https://github.com/fusionpbx/open-source

Issue #1065, opened 2015-07-26
Group Manager Permissions not working


anis-bedhiafi:
Creating a New group in Advanced > Group Manager and setting permissions to the new group are not working.

My configuration:
FusionPBX: 3.9.0 with Postgres 9 on Ubuntu 14.04 LTS
I used the install script from Github (https://raw.githubusercontent.com/fusionpbx/fusionpbx-scripts/master/install/ubuntu/install_fusionpbx.sh)

Here is how I do:

- Go to Advanced > Group Manager and create a new group (set it Global or for the current domain)
- Define Permissions for the new group
- Go to Advanced > Menu Manager and define the menu items for the group
- Go to Accounts > User Manager and create a new user and affect it to the new group.
- Log out and login with the new user
- Strange behavior, the new user can create extensions but cannot see them (only the superadmin can), follow me is not working in the new user account, Extension Summary, Conferences can be created but not shown.
NOTICE: I am sure that I defined all permissions for the above apps

Please help me on this one. What could be the problem? is it from my installation on Ubuntu server or Postgres or what? Am I missing something? Anyone who faced the same problem?

I opened a ticket for this before  #1028 but no answers
@richbreton can you help with this one?

anis-bedhiafi:
I even followed instaructions by word on https://github.com/fusionpbx/fusionpbx
using Debian Jessie PostgreSQL and Nginx but still facing the same problem while creating a new group with different permissions.

Please help what could be the problem?

moteus:
Hello.
First at all I am not expert on FusionPBX but just newbie user.
I have installed current master version on Windows and I can not reproduce problem.
What I do
 * create new group in domain (I have 2 domain in FPBX)
 * Check all permissions on `Extension` group in Group Manager
 * add this group to menu `Account->Extension`
 * Create new user and assign his to group.
 * Login as this user and I can see all existed Extensions as well as create new one.

anis-bedhiafi:
Thank you @moteus for the reply.
What about conference center can you create a new conference and see it using this new user?

moteus:
I can add new conf but I can not see it because of

```PHP
if (if_group("superadmin") || if_group("admin")) {
	//show all extensions
	$sql = "select * from v_conference_centers ";
	$sql .= "where domain_uuid = '".$_SESSION['domain_uuid']."' ";
}
else {
	//show only assigned extensions
	$sql = "select * from v_conference_centers as c, v_conference_center_users as u ";
[...]

PR #1146, opened 2015-09-07
Random ring groups

badcrc:
This adds the random strategy to the ring groups, its simply does an order by random in the SQL when selecting the members of the group. It should support mysql and postgresql (though I only tested it in psql).

Also included is a lua script for entering/leaving a ring group, the code is based on the fifo_member.lua and needs a dialplan entry like in the queues.

### DIFF
--- a/resources/install/scripts/app/ring_groups/index.lua
+++ b/resources/install/scripts/app/ring_groups/index.lua
@@ -158,6 +158,28 @@
 		--forward the ring group
 			session:execute("transfer", ring_group_forward_destination.." XML "..context);
 	else
+		--get the strategy of the ring group, if random, we use random() to order the destinations
+			sql = [[
+					SELECT 
+						r.ring_group_strategy
+					FROM 
+						v_ring_groups as r, v_ring_group_destinations as d
+					WHERE 
+						d.ring_group_uuid = r.ring_group_uuid 
+						AND d.ring_group_uuid = ']]..ring_group_uuid..[[' 
+						AND r.domain_uuid = ']]..domain_uuid..[[' 
+						AND r.ring_group_enabled = 'true'  
+					]];
+
+			--freeswitch.consoleLog("notice", "SQL:" .. sql .. "\n");
+			assert(dbh:query(sql, function(row)
+				if (row.ring_group_strategy == "random") then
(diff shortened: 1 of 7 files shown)


resources/install/scripts/app/ring_groups/index.lua
```diff
@@ -158,6 +158,43 @@
...
+					SELECT 
+						r.ring_group_strategy
+					FROM 
+						v_ring_groups as r, v_ring_group_destinations as d
+					WHERE 
+						d.ring_group_uuid = r.ring_group_uuid 
+						AND d.ring_group_uuid = ']]..ring_group_uuid..[[' 
+						AND r.domain_uuid = ']]..domain_uuid..[[' 
+						AND r.ring_group_enabled = 'true'  
+					]];
```
moteus: You do not need here join with `v_ring_group_destinations`
You can use just
```Lua
		sql = [[
				SELECT 
					r.ring_group_strategy
				FROM 
					v_ring_groups as r
				WHERE 
					AND r.ring_group_uuid = ']]..ring_group_uuid..[[' 
					AND r.domain_uuid = ']]..domain_uuid..[[' 
					AND r.ring_group_enabled = 'true'  
			]];
```

resources/install/scripts/app/ring_groups/index.lua
```diff
@@ -158,6 +158,43 @@
 		--forward the ring group
 			session:execute("transfer", ring_group_forward_destination.." XML "..context);
 	else
+		--first we check what version of SQL are we using
+			sql = "SELECT version() as version";
```
moteus: `database.type` in config already has database type (e.g. `pgsql`) I think you can use it.
Checkout [directory.lua](https://github.com/fusionpbx/fusionpbx/blob/be564502df73a93edd09b0a8b11a05f49382dfcd/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua#L172-L177)

badcrc:
I've updated the code with the tips from @moteus , thanks!

Also, I've included a couple of permissions fixed for issue #1065

badcrc:
Thanks! seems that SQLite has the random() as well, MySQL seems to be the stranger one here.

PR #1185, opened 2015-10-06
Fix. use timeout handler in Enterprise ring group

moteus:
see https://github.com/moteus/fusionpbx/commit/e5a0134ec6c36e44d0828301b20c9c194c25cd78#commitcomment-13601198


franktiemissen:
Tested this, works! Also intercept keeps working. Please Commit.

PR #1188, opened 2015-10-07
Add. Support block calls by called number

moteus:


moteus:
I think main idea is for call center who make outbound calls. If some one ask to do not call to him any more operator may add his number to block list.

DigiDaz:
The purpose of this feature was to extend the call block to work for outbound numbers too.

This is particularly useful in sales type environments and allows for a "do not call" list so people who have asked that they are not called again, cannot be repeatedly called.

DigiDaz:
Any reason why this is not being comitted?

mafoo:
i would like this feature too

daniel-lucio:
I think there is a better way to do this. I mean, non-solicitation lists are public (well, you need to pay for them, but they are some kind of public). An optional Lua script that verifies the outgoing call before bridging it would be better for law-enforcement.

I understand that the only legal way a call center stop calling is by adding your number to a non-solicitation government phone list (at least is the way it works in Mexico and Canada, I guess other countries have something similar). The script would need to be designed to accept different listing formats so anyone could add it's country way.

markjcrane:
If you want to get them merged then please fix the conflicts and re-open or submit a new pull request.

markjcrane:
Reason I didn't accept the pull request. I thought the change wasn't intuitive.

A new select box was added by this pull request.

In the database it was called call_block_number_type with the following options in the select box.
- caller
- called

I still find this wording confusing. But assuming it meant blocking outbound calls or blocking inbound calls. In which case type: inbound, and outbound would be more consistent with the rest of the theme in the interface.

Issue #686, opened 2015-06-30
ring group strategy rollover 


GoogleCodeExporter:
```
The problem is when the first ringing extension picked up, the other didn't 
ringing if second call coming in.

The problem is solved when I removed domain_name from line 421 of 
app/ring_groups/index.lua:

 extension_status = "show channels like "..destination_number.."@"..domain_name;

changed to 

 extension_status = "show channels like "..destination_number.."@";

Please check if this is the issue.
```

Original issue reported on code.google.com by `spw...@gmail.com` on 15 Aug 2014 at 7:12

badcrc:
I noticed this as well, I was having problems in this scenario:

- sequence ring group with the extension 100 and 101 on it
- the 101 tells me that somethings when the 100 is idle, the calls go directly to 101

I looked at the lua for the ring groups and checked how the matching is done. Turns out that doing the "show channels llike destination_number" also matches a lot of things (for example someone calling a number with an 100 on it) even with the aditional check for looking for the domain.

I'm currently trying this fix: exploding the show channels reply and looking for the "destionation_number@domain_name", seems to be working for now, I will send the fix after more testing!

PR #1197, opened 2015-10-14
This should do a better job detection active calls per extension in t…

badcrc:
…he ring group with skip_active

Related to issue #686 , this should check for destination_number@domain_name in the reply and should be more exact than "show channels like 100" (where 100 could be anywhere in any channel, not just the destination_number).

### DIFF
--- a/resources/install/scripts/app/ring_groups/index.lua
+++ b/resources/install/scripts/app/ring_groups/index.lua
@@ -38,6 +38,7 @@ local log = require "resources.functions.log".ring_group
 	require "resources.functions.explode";
 	require "resources.functions.base64";
 	require "resources.functions.file_exists";
+	require "resources.functions.explode";
 
 --get the variables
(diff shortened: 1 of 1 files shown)


resources/install/scripts/app/ring_groups/index.lua
```diff
@@ -443,12 +444,23 @@ local log = require "resources.functions.log".ring_group
 								cmd = "show channels like "..destination_number;
 								reply = trim(api:executeString(cmd));
 								--freeswitch.consoleLog("notice", "[ring group] reply "..cmd.." " .. reply .. "\n");
+								exploded_reply = {};
```
moteus: You do not need this line. You overwrite value in next line

badcrc: I wasn't sure about that tbh. I'm pretty new to lua so I just declared it first.

PR #1202, opened 2015-10-20
Fix. Use `api show channels like` did not work on fusion with BDR.

moteus:


badcrc:
I works :)

I'm using postgresql for the database.

PR #1203, opened 2015-10-22
Add. Ability to order the time conditions

badcrc:
When adding a new time condition it always gets generated as the last one, sometimes this isn't desired and the solution is to edit de dialplan manually and change the group for all the entries, which can be tedious if you have multiple conditions.

With this the "order" of the dialplan is also presented in the time condition form so can be edited.

PS. A drag&drop interface would be more intuitive so it doesn't presents the group_id to the final user but my javascript is kinda basic for doing that :/


markjcrane:
Adding order is a great feature to time conditions. I updated the placement of it slightly so it is a little more vertically condensed.

PR #1209, opened 2015-10-26
Fix. Call intercept - a scenario where an outgoing call can be intercepted

badcrc:
If Alice calls Bob's mobile and Bob returns a "183 Session progress" with SDP (for example if the mobile has no service) the call stays in EARLY callstate in the database.

If Charlie (from the same Call Group as Alice) dials "*8", the outgoing call from Alice gets chosen as an "interceptable" call and freeswitch bridges Alice and Charlie's legs, sending a "CANCEL - Reason: SIP;cause=200;text="Call completed elsewhere" to Bob's call leg.

This should fix it. We make sure that the direction of the call is outbound.

### DIFF
--- a/resources/install/scripts/intercept_group.lua
+++ b/resources/install/scripts/intercept_group.lua
@@ -186,6 +186,7 @@
 			end
 			sql = sql .. ") ";
 			sql = sql .. "and call_uuid is not null ";
+			sql = sql .. "and direction = 'outbound' ";
 			--if (domain_count > 1) then
 			--	sql = sql .. "and context = '"..context.."' ";
 			--end


moteus:
checckout [line number 176](https://github.com/fusionpbx/fusionpbx/blob/master/resources/install/scripts/intercept_group.lua#L176) :)
I already try to put this but Mark did not accept this.
Lets try again :)
+1

badcrc:
Ouch, didn't saw that :/

Is there any case where that is the desired effect?

moteus:
I do not know. May be it make sense for direct intercept (**ext) but I doubt.
I think I just could not explain this changes and Mark did not accept it because it brakes old behavior.

markjcrane:
which git pull are you referring to?

On Mon, Oct 26, 2015 at 11:39 AM, Alexey Melnichuk <notifications@github.com

PR #1242, opened 2015-11-18
local_hostname needs to use hostname

daniel-lucio:
hostname and switchname are not always the same. hostname warrants the server name, switchname doesnt. This will fix conference when you have db backend, no loadbalancing enabled and HA environment (active - pasive).

nasty workaround is to add an entry of switchname in /etc/hosts

### DIFF
--- a/resources/install/scripts/app/conference_center/index.lua
+++ b/resources/install/scripts/app/conference_center/index.lua
@@ -385,7 +385,7 @@
 			end
 
 		--check if someone has already joined the conference
-			local_hostname = trim(api:execute("switchname", ""));
+			local_hostname = trim(api:execute("hostname", ""));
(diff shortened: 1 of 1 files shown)


resources/install/scripts/app/conference_center/index.lua
```diff
@@ -385,7 +385,7 @@
 			end
 
 		--check if someone has already joined the conference
-			local_hostname = trim(api:execute("switchname", ""));
+			local_hostname = trim(api:execute("hostname", ""));
```
moteus: I think it should be switchname. channels table contain switchname.

markjcrane:
switchname is more versatile and moving to hostname would reduce functionality.

daniel-lucio:
http://inside-out.xyz/technology/i-found-a-bug-in-fusionpbx-about-high-availability-and-conferences-but-my-patch-was-rejected.html

PR #1250, opened 2015-11-19
Update y000000000028.cfg

blackc2004:


markjcrane:
I just updated the variable names adding the underscore.

PR #1255, opened 2015-11-26
Make class dialplan compatible with global dialplans

daniel-lucio:
There are 2 issues here:
1. the session variable is not updated if it exists, taking out the if will force reconstruction each time the broadcast is sent. this warrants always to use last set of dialplans, as we dont know if session is kept open or not for a long time
2. the were clause was ignoring global dialplans, i have copy the were clause (more or less) from the xml handler.


markjcrane:
The dialplan context ${domain_name} was needed for the xml handler. Its not really needed for this query.

PR #1256, opened 2015-11-26
Migrate common text from /*/app_languages.php

mafoo:
- [x] First pass /app/*
- [x] First pass /core/*
- [x] First pass /themes/*
- [x] add an app that can analyze missing definitions
- [x] Migrate existing /core/app_languages.php to /resources


markjcrane:
The 'core' directory should have nothing outside of a directory. If it is for core then it would belong primary 'resources' directory.

mafoo:
This should be ready for integration as stage one is complete.
stage two will be multi commit of each language rename

PR #1260, opened 2015-11-27
freeswitch.email() function has change behaivour

daniel-lucio:
before 1.6.5, first two arguements could be "", and freeswitch would use system value. Since 1.6.5 they are mandatory, they must be any valid email regexp.  As the variable voicemail_mail_to has a valid one, we use it.


moteus:
Please test current master. it should works.

daniel-lucio:
Not fixing the master, it is a bugfix for 4

markjcrane:
Conflict on this so I just manually made the change to the 4.0 branch.

daniel-lucio:
Thanks.

PR #1263, opened 2015-11-28
Change. Rename v_mailto.bat to mailto.bat

moteus:


daniel-lucio:
v_ prefix on scripts are historical. SO when you update it wont break

PR #1271, opened 2015-11-30
Enhance install based on feedback

mafoo:
All Done, reworked the way installation settings are passed around to use a central class to marshal all settings. Use the logic in this class to differentiate between first time install and upgrade
@todo
- [X] show all assumed paths
- [x] Under windows warn user to create system DSN if postgresql
- [x] Under windows create odbc connection automagically


markjcrane:
Under windows create odbc connection automagically - Unlikely with a php file.

PR #1279, opened 2015-12-04
Convert install_switch upgrade routines to app/switch

mafoo:
- [x] migrate file functions to functions.php
- [x] migrate copy_scripts to app/switch/app_defaults.php
- [ ] migrate /resources/install/scripts to \*/\*/resources/scripts
  - [x] move all scripts to /app/switch/resources/scripts
  - [x] update app/switch/app_defaults.php to search all apps for scripts to copy
  - [ ] distribute scripts to correct applications
- [ ] Remove upgrade switch functions from core/upgrade


markjcrane:
I moved copy scripts to app/scripts/resources/classes within a new class called scripts with two new methods copy_files and write_config. These are called by app/scripts/app_defaults.php and can be run when running advanced -> upgrade -> app defaults. I have not tested for new installs so install_switch class I did not remove the copy scripts and the method that creates the config.lua.

markjcrane:
Your update with the resources/functions.php with the better recursive_copy and recursive_delete would be a great improvement to the functions.php.

markjcrane:
Sorry when I started the code I did I had read this code and thought it was committed. I should have committed it then completed the work based on it.

mafoo:
I think this is now defunct as i have made a separate PR to cover the migration of file functions.

PR #1287, opened 2015-12-07
Added switch version

mafoo:
converted version label to text[*]
renamed labels


markjcrane:
Translation labels already done for those elements.

PR #1310, opened 2015-12-09
Update index.lua

blackc2004:
Adding USER NOT REGISTERED forward options to call control


moteus:
One note. This type of forwarding doeas not works for ring group.
Which may be surprise for some one.

PR #1321, opened 2015-12-10
CDR Stats changes

blackc2004:
Add advanced search to CDR Stats page, change graph to be time based so you can select wider time margin and it scales correctly.

<48 hours: Shows in hours
48-168 hours: Shows in Month:Day
>168 hours: Shows in Months


blackc2004:
Apparently this created one pull request for two things. This also includes ability to download extensions via CSV

PR #1328, opened 2015-12-11
Fix. Append PROJECT_PATH to DOCUMENT_ROOT

stefanomarty:
To get the Advanced -> Update -> Source menu working:

chdir($_SERVER["DOCUMENT_ROOT"].PROJECT_PATH);

as:
$_SERVER["DOCUMENT_ROOT"] = /var/www
PROJECT_PATH = /fusionpbx

### DIFF
--- a/core/upgrade/index.php
+++ b/core/upgrade/index.php
@@ -53,7 +53,7 @@ if (sizeof($_POST) > 0) {
 
 	// run source update
 	if ($do["source"] && permission_exists("upgrade_source") && !is_dir("/usr/share/examples/fusionpbx")) {
-		chdir($_SERVER["DOCUMENT_ROOT"]);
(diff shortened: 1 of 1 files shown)


mafoo:
I don't think this is the solution to the problem as on Centos7 and Debian 8 DOCUMENT_ROOT gets resolved by root.php as the correct root folder (/usr/share/fusionpbx, and /var/www/fusionpbx respectively) also this approach means the same change would need to happen in app/system/*.php as it also uses the DOCUMENT_ROOT to find the git folder.
I think it would be worth checking with mark what document_root is supposed to resolve down to then adjust the root.php code to work correctly with the platform you are on.

markjcrane:
root.php changes were committed to reach the end result of this pull request.

PR #1334, opened 2015-12-15
XML Handler incorrectly handling 'date-time' time conditions

kpabijanskas:
This pull requests fixes XML Handler support for 'date-time' time condition. With this, date-time gets rendered as:

```
 <condition date-time="2015-12-15 15:00~2015-12-16 15:00">
```
 Without this, the time condition WOULD get redered as:
```
 <condition field="date-time" expression="2015-12-15 15:00~2015-12-16 15:00">
```
 which is incorrect and would always match false in FreeSWITCH.

 FreeSWITCH docs for this are here: https://freeswitch.org/confluence/display/FREESWITCH/Time+of+Day+and+Holiday+Routing#TimeofDayandHolidayRouting-Variables

 All other time condition variables documented seem to be included in FusionPBX'es XML Handler.

### DIFF
--- a/resources/install/scripts/app/xml_handler/resources/scripts/dialplan/dialplan.lua
+++ b/resources/install/scripts/app/xml_handler/resources/scripts/dialplan/dialplan.lua
@@ -187,6 +187,8 @@
 								condition_type = 'time';
 							elseif (dialplan_detail_type == "week") then 
 								condition_type = 'time';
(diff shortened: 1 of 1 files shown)


moteus:
Please view [this link](https://github.com/fusionpbx/open-source/blob/master/sign-cla.md)

PR #1336, opened 2015-12-18
Enhancements for dialplan class

mafoo:
Added global=’true’ to the extension part to allow a template dialplan rule to be defined as global (rather than template copy)


markjcrane:
Check exists is problema ticket however I liked the global true enough that I wrote the code myself and implemented following other if statements on this page like you did so the code is likely almost identical. Sorry too excited about this improvementioned to wait for the pull request to be fixed.

PR #1347, opened 2015-12-27
Update install_fusionpbx.php

shadowym:
Fix several problems causing mysql create to fail.

-All these fixes were tested on CentOS v7 using PHP v5.4 and MariaDB v5.5

* Without brackets the variables strings are not returned.  

* db_create_password variable is incorrect so I changed it to $this->global_settings->db_create_password().

*  db_password variable is incorrect so I changed it to $this->global_settings->db_password().

### DIFF
--- a/core/install/resources/classes/install_fusionpbx.php
+++ b/core/install/resources/classes/install_fusionpbx.php
@@ -410,21 +410,21 @@ include "root.php";
 					$connect_string;
 					if (strlen($this->global_settings->db_host()) == 0 && strlen($this->global_settings->db_port()) == 0) {
 						//if both host and port are empty use the unix socket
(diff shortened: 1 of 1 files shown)


shadowym:
Also

* Changed "CREATE USER" to "GRANT SELECT ON" which is the mysql way of doing "create user if not exists" so that the command does not fail if the user already exists which throws an exception and stops the install script

moteus:
Please view [this link](https://github.com/fusionpbx/open-source/blob/master/sign-cla.md)

markjcrane:
shadowym if you follow the instructions on https://github.com/fusionpbx/open-source we will accept the code changes otherwise the git pull will be closed.

markjcrane:
Once you follow the instructions as previously given I will re-open the the pull request and merge it. Otherwise I will consider removing MySQL support from the script since its broken and I preferred database is PostgreSQL.

PR #1349, opened 2015-12-29
bdr manger, emails, device download

blackc2004:
A simple BDR manager to show LAG and Retain bytes, email show all button, and device download option.


markjcrane:
BDR interface is a good feature but this it the wrong repository for it. The reason is that some people install sqlite and a few mysql. And then most don't have BDR installed when using PostgreSQL.

This is the right location.
https://github.com/fusionpbx/fusionpbx-apps

blackc2004:
Ok! I'll move it. Thanks.

PR #1350, opened 2015-12-29
device download, show all emails

blackc2004:
Device download options

Show all emails button

Fix for show all on advanced CDR searchs


markjcrane:
Please submit a cleaned up pull request that is focused on one main topic per pull request.

PR #1373, opened 2016-01-05
fix directory setting url typo

blackc2004:


blackc2004:
fix directory setting url typo and also fix the consistency of LDAP settings in all yealink files so that they have the same variable names.

PR #1385, opened 2016-01-13
yealink remote phone book 

blackc2004:
start work on the yealink remote phone book support for local extensions.


markjcrane:
Nice!

PR #1403, opened 2016-01-21
Enhance app/ring_groups

mafoo:
Enhancements to ringback selection on ring_groups

added facilities to select default_ringback (as defined in vars)
prefix moh selectors with moh: (configurable via app_languages)
import all ringbacks from vars for selection
default for new ring groups is default_ringback


mafoo:
RE-written to use opt groups instead should be ready for merge

mafoo:
closing this off as it is to be restructured into several components

PR #1414, opened 2016-01-28
Fix. Path to records in music on hold.

moteus:


mafoo:
The problem is this will then get domain specific moh and not global as well

moteus:
This is only for records. And records are always domain specific.

mafoo:
Oh yeah! doh I read it wrong as the same class handles moh and recordings listing

PR #1417, opened 2016-02-01
Bugfix-conference center rooms

mafoo:
Fix for when a conference room has no admins

If a conference room had no admins it would become impossible to see it
in the GUI
This fix makes it possible for a admin or super admin to view rooms in
this state.
- [x] convert to use permission "conference_room_view_all"

### DIFF
--- a/app/conference_centers/app_config.php
+++ b/app/conference_centers/app_config.php
@@ -58,6 +58,10 @@
 		$apps[$x]['permissions'][$y]['groups'][] = "admin";
 		$apps[$x]['permissions'][$y]['groups'][] = "user";
 		$y++;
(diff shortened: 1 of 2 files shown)


daniel-lucio:
instead of doing if (if_group("admin") or if_group("superadmin")) {, check if there is a grou permision, that is the correct way

markjcrane:
He is correct permission is the correct way and if_group("admin") and if_group("superadmin") are to be rarely in favor of permissions.

mafoo:
added permission conference_room_view_all, I will bare in mind when adjust existing code that is using is_group to convert it at the same time over to permission instead

markjcrane:
Please rebuild the pull request with the new conference_room_view_all permissions.

PR #1418, opened 2016-02-01
Enhance Ringback

mafoo:
updated to include default_ringback option

ability to include default_ringback is coming with the GUI update for
ringbacks

### DIFF
--- a/resources/install/scripts/resources/functions/format_ringback.lua
+++ b/resources/install/scripts/resources/functions/format_ringback.lua
@@ -2,20 +2,26 @@
 --add the format_ringback function
 	function format_ringback (ringback) 
 		if (ringback == nil or ringback == "") then
-			--get the default ring back
+			--get the default music on hold
 				ringback = session:getVariable("hold_music");
-		elseif (ringback:match("%${.*}")) then
-			--strip the ${ and }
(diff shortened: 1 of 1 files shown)


markjcrane:
You haven't revealed  what default_ringback the empty string at this time is the default this appears to be a second default. So a further explanation or the gui part you hinted at us required.

resources/install/scripts/resources/functions/format_ringback.lua
```diff
@@ -2,20 +2,26 @@
...
-			--fallback to us-ring
-				if (ringback == "") then
-					ringback = session:getVariable("us-ring");
-				end
-			--convert to tone_stream
-				ringback = "tone_stream://" .. ringback .. ";loops=-1";
+		else
+			if (ringback == "default_ringback") then
+				--get the default ring back
+					ringback = session:getVariable("ringback");
```
moteus: is it possible get `nil` as result? i really do not check it at all

markjcrane:
We have a nil on the session as xml handler doesn't have sessions. So this new function broke call center I'm fixing it now. Since I have to remove session to fix this problem this pull request needs to be scrapped for now.

PR #1419, opened 2016-02-02
add tenant domain into the title

daniel-lucio:
if $_SESSION["theme"]["title.domain"]["boolean"] is set to true, the tenant domain will be added in the title


markjcrane:
I don't see how this is very useful considering the title goes in the tab title and is cut off.

daniel-lucio:
It is for comptability software. It cuts on the view but detected.  It was a sponsored add. No idea exactly how the software works.

PR #1432, opened 2016-02-05
IVR menu

moteus:
Simplify code so it easy to read and maintain.

Fix some bugs

1. Repeat execute only one action.
  * Create IVR with 2 optioins. First is `streamfile.lua ....` and second is `transfer ...`
  * Call to IVR and select first option.
  * After that if you try select second one IVR still will execute only first option.
2. Execute several actions
  * Create IVR with 2 options with same key. First is `phrase ..` with order 0 and second is `transfer ...` with order 1
  * Set `Max Failures` to at least 2
  * Call to IVR and select option.
  * FS will play phrase and then return to main menu.
  * If you reach to `Max Failures` tries only after that FS do transfer.
3. Infinity loop
  * Create IVR with Direct dial=true
  * Set `Max Failures` to at least 2
  * Call to IVR and try dial invalid numbers.
  * IVR never end until you stop.


mafoo:
Tested on FS 1.6.6,  Fusion 4.1 (commit 3ce7924c02ac1ff6c43dcb1c4440fbd1ec9c1302), OS CentOS 7

PR #1437, opened 2016-02-11
fix for when a path has been set via vars.xml

mafoo:
if vars.xml has a path set before the FT install runs we currently splat the vars.xml so the path gets lost.
This patch checks vars.xml for any \w+_dir vars and adds them into the app_defaults to get rebuilt.
- [ ] Tested
- [x] Rebase

### DIFF
--- a/app/vars/app_defaults.php
+++ b/app/vars/app_defaults.php
@@ -245,6 +245,26 @@ EOD;
 	if ($domains_processed == 1) {
 
 		$result = json_decode($vars, true);
+
+	// detect and add paths present in existing vars.xml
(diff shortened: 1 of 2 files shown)


markjcrane:
Give an example of a directory with _dir in advanced variables only thing I see is internal_ssl_dir or external_ssl_dir. Do you have any others?

mafoo:
DigiDaz has this issue on the ARM builds where the compiler is not setting the script_dir at compile time so is using vars.xml to set it, this patch is to support that configuration or anyone else needing to override the compiled paths in this way.

DigiDaz:
Its not just ARM, I'm currently testing on x86_64, same thing

mafoo:
this will be suspended until Free switch clarify if you are supposed to set dirs this way

PR #1438, opened 2016-02-11
check if the user has overridden dirs in vars.xml

mafoo:
This is unsupported, cause a Exception if it is used


mafoo:
yes it is as it checks if someone has overridden directories, which they shouldn't (but it is not obvious you should not) so we need them to fix it before continuing the install

markjcrane:
[a-zA-Z_]+[a-zA-Z]_dir)= may be too generic it is possible to have false positives when settings something with _dir in the variable name that is not one of freeswitch's directories.

PR #1449, opened 2016-02-25
Wrong variable in freeswitch.email

daniel-lucio:
according to https://wiki.freeswitch.org/wiki/Mod_lua#freeswitch.email, it should be from, not to. This will prevent problems with hardened mail systems.


markjcrane:
This has been fixed in Master for a while but directly copy changes to 4.0 as the changes were more extensive

PR #1486, opened 2016-03-11
Fix a few email issues

blackc2004:
1.  Fix resend button so that if the message fails again it doesn't duplicate the entry
2. Fix email_delete so that if you are doing "show all" it actually deletes the email entry
3. Add an email_cron.php so that you can call email resends via cron job.


mafoo:
You may want to rebase your commit against fusionpbx/master, this will leave the Merge cleaner for changes. If you are not sure how to do this i will hopefully have a guide up in a few days how to do it with TortoiseGit

PR #1513, opened 2016-03-27
add 'regex'-functionality via dialplanmanager (edit extension only)

schraegervogel:
In the dialplanmanager it is possible to select 'Regular expression', but in dialplan.lua is no way to get it out of the DB and write it to XML. With these changes no it's possible.

I hope this solves Issue 623


markjcrane:
If you want to update this pull request against the latest code then it will be committed. Sorry this wasn't done sooner before the conflict. Its its updated will re-open this pull request.

PR #1514, opened 2016-03-28
Fix. add `PROJECT_PATH` to menu item url

moteus:
P.S. Also class `menu` contain type `email` for menu item.

### DIFF
--- a/themes/default/template.php
+++ b/themes/default/template.php
@@ -1659,7 +1659,15 @@
 								if ($submenu) {
 									echo "<ul class='dropdown-menu'>\n";
 									foreach ($menu_parent['menu_items'] as $index_sub => $menu_sub) {
-										$mod_a_2 = ($menu_sub['menu_item_link'] != '') ? $menu_sub['menu_item_link'] : '#';
+										$mod_a_2 = $menu_sub['menu_item_link'];
+										if($mod_a_2 == ''){
+											$mod_a_2 = '#';
+										}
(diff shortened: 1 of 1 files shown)


stefanomarty:
Thanks, I have the same problem, all the menu links miss the PROJECT_PATH part, while the links on the pages are ok.

With the new menu stile (black background) I cannot see the icon in the top right angle of the page, the one which let you change the domain: link is there and ok, probably the icon is black and disappears in the background.

reliberate:
The *three-bar* icon you are referring to is no longer present in the new (default) theme.  Access to the Domain Selector is available through the domain name displayed near the right end of the menu. If you do not see the domain name displayed, modify the theme > domain_color > text value in Default Setting.

reliberate:
You'll also want to make sure that theme > domain_visible > text is set to 'true', in Default Settings.

stefanomarty:
Thanks!

_domain_visible > text_ was already true, while _theme > domain_color_ was #000000 (same as background). I made it #ffffff and the domain name is now clearly visible.

I liked the _three-bar_ icon, kinda more intuitive (yep, I'm on a Mac :-), anyway the domain name link works as well.

reliberate:
The _three-bar_ Domain Selector icon was nice, but it would have conflicted visually with the traditional _menu _ button that is displayed (only) on mobile devices, as the menu is now using Bootstrap.  So, the decision was made to drop it, and just use the domain name to open the Domain Selector.

PR #1519, opened 2016-03-29
Fix. Generate same `config.lua` on install as on Upgrade.

moteus:


markjcrane:
I fixed the issue by getting rid of _conf. and then ensuring that there is a default value.
We now have  expire.sofia = "3600"; and expire.acl = "3600";

PR #1520, opened 2016-03-29
Fix. Install with MySQL

moteus:


markjcrane:
Looks like this was handled by a different pull request.

PR #1531, opened 2016-03-31
allow outside transcription service to be called

blackc2004:
Add in a function to allow for calling an outside transcription service. One would create a file called /app/emails/email_translation.php with a function get_transcription which accepts WAV/MP3 file path. Within this function you would call outside voice to text transcription service such as ATT-API. Function would return the result. This would then get appended to the email body.

### DIFF
--- a/app/edit/app_defaults.php
+++ b/app/edit/app_defaults.php
@@ -70,7 +70,7 @@ if ($domains_processed == 1) {
 		}
 
 	//add the missing default settings
-		foreach ($missing as $row) {
+		if (count($missing) > 0) foreach ($missing as $row) {
 			//add the default settings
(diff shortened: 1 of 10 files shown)


blackc2004:
Are there any issues with this pull request that's keeping it from being merged?

moteus:
Your PR includes changes not related ot transcription but also to some Provision.
I think you should leave only related changes and make anothe PR if you need.

blackc2004:
Yes, I tried when submitting that to be a separate pull, but couldn't get it to work. I'm not sure at this point how to create a new pull for that. Can this just be accepted as is and I will work in the future on creating multiple pulls?

PR #1535, opened 2016-04-05
Allow use odbc to store files (record/voicemail) in database.

moteus:
Tests:
* ivr_menu.lua
  * [x] play greeting from db. (remove from recordings/... and call to IVR)

* ivr.conf.lua
  * [x] save file while building config

* recordings.lua
  * [x] record file via (*732)

* streamfile.lua 
  * [x] make IVR option to play greeting

* Voicemail
  * [x] record greeting
  * [x] listen greeting when call to extension
  * [x] listen greeting while choose before set
  * [x] leave mesage to unregistered extension
  * [x] listen new message
  * [x] set greeting and listen greeting
  * [x] send message to email
  * [x] record name

* directory.lua
  * [x] serch user

* fax queue
  * [x] save file after fax send

* fax dierct (without queue)
  * [ ] save file after fax send

Why need `base64/read` role:
`freeswitch.Dbh` with odbc works not correct. FS uses `SQLDescribeCol` to get column size and then allocate array and read from DB this amount of data. But ODBC driver does not return full size but it return some size (8k for pgsql 32k for sybase). So user should call `SQLGetData` multiple times until there data but FS does not do this and as result we get only first 8k of data.


moteus:
Fix confilict

markjcrane:
What I didn't like in this commit was this.
database.backend.base64 = 'luasql'\n";

The database engine used is hard coded in the script it should be in config.lua.

moteus:
This is default value. You can redefine it in local.lua file.
On my test system I have
```lua
database.backend = {
  main            = freeswitch and 'native' or 'odbc';
  base64          = 'odbc';
  ['base64/read'] = 'odbc';
}
```

markjcrane:
If you fix the merge conflicts then will accept the pull request.

PR #1543, opened 2016-04-12
Fix. Display `Switch Uptime` on User Dashboard.

moteus:


reliberate:
Thanks, moteus.  Good catch.

PR #1551, opened 2016-04-17
Update app_config.php

zhulizhong:
add chinese


markjcrane:
Once you have signed the contributor agreement then could accept your translation.

PR #1553, opened 2016-04-18
Add. optional intercept only inbound/outbound calls.

moteus:
Note.
  * `inbound` is call to some extension
  * `outbound` is call from some extension to external number.

Currently there no way to find call from extension to extension as `outbound`

### DIFF
--- a/resources/install/scripts/intercept.lua
+++ b/resources/install/scripts/intercept.lua
@@ -25,6 +25,7 @@
 
 --user defined variables
 	local extension = argv[1];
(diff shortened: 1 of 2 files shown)


markjcrane:
Where is direction:find being defined?

moteus:
Iam not sure I understand your question.
`direction` is declare in 
```lua
local direction = argv[2] or extension and 'inbound' or 'all';
```
`direction:find` is call `string.find` method for `direction

PR #1557, opened 2016-04-26
Mariadb 10.1, Freeswitch 1.6, Ubuntu Xenial LTS, Ldaps, warnings enabled, cli install/upgrade

hcoin:
and, a few other fixes/cleanups.

Useful project, intuitive GUI.  Thanks!


moteus:
But it breaks work with Php7. I run test system with apache 2.4 and PHP 7 and Fusion works correctly
I do not think it good idea break this

hcoin:
With nginx, xenial, php5 and php7 installed-- not one fusionpbx screen appears.  Remove 7, or apply the fix and leave 7, and it works.

hcoin:
I suggest for php7 devs, change instances of php5 -> php7 in the fix.

moteus:
Do you use php 7 for nginx?

hcoin:
nginx -> php5/fpm -> fusionpbx.   Understand, the motivator here is ease of scale, reducing chances for time-eating install glitches.  Can't assume anything about other sites, scripts, etc. there are.  So calls within fpbx to naked 'php' could easily find itself running a different version of php than the main site.  I suppose the long term fix would be to make all references to php refer to a variable.

moteus:
but you just need configure your system to use same php both on nginx and on cli. In other case it will breake if some one have only php7

hcoin:
Your idea of 'the same php on both nginx and cli' works when every application running needs the same version of php--- and the same person is willing to fix/manage fpbx upgrades to 'the latest php' at the same time as other apps on the same box (that need the later php) are being worked on (usually under time pressure).

If you want fpbx to scale, you can't assume 'php' gets you php5 or php7 ... or a version of php that will roll out some day and break fpbx because internally fpbx just points to 'php' and isn't tested.

Ease of scale calls for attention to such details.

moteus:
Then at least it should be configuriable but not hardcoded.

hcoin:
Yes.   Until the right way to set up such a variable is decided: Because php7 is new and php5.6 is stable, kindly accept the patch otherwise fpbx will be seen to 'not just work' for everyone with later versions of ubuntu, debian etc that likely will have both 5.6 and 7 installed.  Or, fpbx will break mysteriously when some dependency of an unrelated package causes a php7 on an already running fpbx system.

moteus:
It possible create one global variable in config.php and use it. Do not think it worth set in `Default settings`.

hcoin:
Because It is known at the time of release which version(s) of php fpbx were used for testing, I think the best approach would be:

1)  At install time, and
2)  At upgrade time:
3)  Detect the highest php version installed on the machine that the current release of fpbx was tested to use, then:
4)  set a variable to specify exactly that version everywhere in fpbx from then on.

I think that's better than a config file approach.   Developers can always edit the 'allowed version' to add versions on development machines.

What I am certain of is this:  It should not be a matter outside the control of fpbx which version of php gets run by *every* fpbx component.

markjcrane:
I think its unlikely to have multiple versions of php on the same machine. Most people will not have this problem. This is an edge case. Hard coding it to a specific version of PHP isn't a good solution setting a place for a variable is far better.

Also voicemail_id cast as used in one of the commits is a weak solution as it would be better to set it as a numerical data type.

I will accept your commit and then fix these items.

markjcrane:
Fixing the notice and warnings has been a goal that has not be completed yet but I do agree its worth getting done which is one of the things I really like in your pull request.

hcoin:
Was a day I too thought having multiple versions of any interpreter would happen only on a developer's desk.   But with these later 'long term release' distros so common in 'real world' deployments (where folk don't upgrade what doesn't need fixing where they work), what happens is some 'security update' here or 'minor tool/new shopping cart pacakge' there causes new versions of python and php to be loaded without notice.   I have systems with three versions of php, two versions of python--- and they were originally installed with one each.    Whatever you do re: this matter, make as they say 'double dog' sure all calls to 'php' are version specific against known-tested releases.  Never re: ".*php$".  A reasonable change would be to detect what version of php fusionpbx finds itself running-- then call that one in the internal-sub invocations.

hcoin:
Re: fixes to allow fpbx to run with warnings enabled.  You are most welcome.  It wasn't high on my list of 'interesting work', but it was necessary to advertise confidence in the package.   And, a side effect of it was... I've read more or less every code file and have 'a grip' on your architecture intent.

So in the category of 'winning ugly'... does 'root.php' really need to be copied all over the place?  maybe some magic in the 'sites-enabled' file instead??

hcoin:
P.S. If ever your travels take you anywhere between Chicago and Des Moines, or southern Wisconsin let me know.

markjcrane:
This pull request I'm waiting on you moving your hcoin.md file or create a new one in in the contributors directory in https://github.com/fusionpbx/open-source. Please do it soon longer we wait the more likely your pull request will have a conflict.

I go to CluCon in Chicago in August every year.

hcoin:
I think I had git move it....


On 04/28/2016 01:33 AM, FusionPBX wrote:
>
> This pull request I'm waiting on you moving your hcoin.md file or 
> create a new one in in the contributors directory in 
> https://github.com/fusionpbx/open-source. Please do it soon longer we 
> wait the more likely your pull request will have a conflict.
>
> I go to CluCon in Chicago in August every year.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub 
> <https://github.com/fusionpbx/fusionpbx/pull/1557#issuecomment-215324289>
>


-- 



Harry G Coin
Quiet Fountain LLC
2118 Lundy Ln
Bettendorf, Iowa 52722

hcoin:
Re: multiple php's on one machine --- not so rare anymore.  Note:
https://www.symfony.fi/entry/php-7-included-in-ubuntu-16-04-lts-xenial-xerus

Everyone upgrading running fpbx prior will find php5 and 7 installed on the same machine.

hcoin:
Glad it worked out.   Sorry about the tabs and spaces!


On 04/28/2016 03:49 PM, FusionPBX wrote:
>
> Merged #1557 <https://github.com/fusionpbx/fusionpbx/pull/1557>.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub 
> <https://github.com/fusionpbx/fusionpbx/pull/1557#event-644870395>
>


-- 



Harry G Coin
Quiet Fountain LLC
2118 Lundy Ln
Bettendorf, Iowa 52722

PR #1564, opened 2016-04-28
Detect if there is at least one database driver

mafoo:


markjcrane:
Please resolve the conflict and I will commit the change.

PR #1577, opened 2016-05-04
Fix Yealink provisioning for contacts/groups

blackc2004:
Fix the Yealink provision templates to allow for directory_extensions,
groups and users.


markjcrane:
where group_uuid = '".$_SESSION['user_uuid']."' ";
user_uuid and group_uuid are not compatible.

blackc2004:
Mark, I already resolved that after you mentioned it on IRC and my pull request is now showing:
where group_uuid = '".$_SESSION['group_uuid']."' ";

Which should be correct.

PR #1586, opened 2016-05-06
Master

kb3vgw:
Here is all the pkgs src merged to head and works so deb pkgs can be made. I have a build script also .


markjcrane:
I think we need a different github repository for the debian packages.

markjcrane:
Will need to get this updated lets please contact me in the next week or two to take a look at a new version of the package.

PR #1588, opened 2016-05-06
Yealink directory listing for extensions, groups, users contacts

blackc2004:


blackc2004:
Includes the changes from PR 1582 with some fixes to the SQL.

Also merged all the directory_* options into ONE directory.xml

Will need to call the it with any of these:

http://mydomain/app/provision/?file=directory.xml&contacts=groups

http://mydomain/app/provision/?file=directory.xml&contacts=users

http://mydomain/app/provision/?file=directory.xml&contacts=extensions

http://mydomain/app/provision/?file=directory.xml&contacts=all

PR #1590, opened 2016-05-10
Enhance Ringback

mafoo:
created new class to look after generating the list of ringbacks
available
adjusted MOH class so legacy method is still possible (just not
recommend) and utilizes optgroups as well
converted ring_groups and ivr_menus to use new method

### DIFF
--- a/app/ivr_menus/ivr_menu_edit.php
+++ b/app/ivr_menus/ivr_menu_edit.php
@@ -28,6 +28,7 @@ require_once "root.php";
 require_once "resources/require.php";
 require_once "resources/check_auth.php";
 require_once "resources/classes/logging.php";
+require_once "resources/classes/ringbacks.php";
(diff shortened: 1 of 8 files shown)


markjcrane:
For new translations in app_languages.php use an online translator to come with with placed holders in that language where a human translator can come back and improve it.

I don't understand changing the language to null. If it is null then why use this method at all?
$text = $language->get(null, 'app/music_on_hold');

Method list_recordings should come from table v_recordings instead of the file system.

mafoo:
Translation done best I can

if you use $null for the first argument to text classes get function it tells it to auto detect the language see https://github.com/fusionpbx/fusionpbx/blob/master/resources/classes/text.php#L64-L69

i will take another look at the code that gets the recordings as all i did was reorganise the code and fix it rather than write from scratch

PR #1603, opened 2016-05-19
Flowroute SMS

blackc2004:
Basic start of Flowroute SMS. Can send and receive SMS from flowroute.


blackc2004:
Set flowroute API URL to: http://yourdomain/app/sms/sms_api.php

Add two default settings:
Category: sms
Sub Category: flowroute_access_key 
Sub Category: flowroute_secret_key

markjcrane:
Your is named app/sms. If this SMS app only works for flowroute then it should be called sms_flowroute unless its sufficiently abstracted enough to support other carriers as well then it would be fine to be called app/sms.

If it is to be only for flowroute then it should be committed to https://github.com/fusionpbx/fusionpbx-apps

blackc2004:
Mark, The idea here is that it can be extended to work with other carriers. For now it supports flowroute as that's the only carrier I can test with. 

You can do with it as you wish.

markjcrane:
Please in the future do not combine different things like sms and yealink changes. Each major feature change should be its own pull request. I accepted the pull request but will like move https://github.com/fusionpbx/fusionpbx-apps for an additional reason its not translated not even with an online translator and we are close to a release and a non translated feature isn't ready for a release.

blackc2004:
Sorry, but I'm not sure what you mean by yealink changes? There are no yealink changes in the files committed. 

Also what online translator?

markjcrane:
Sorry on Yealink changes github interface is showing recent changes on the pull requests which is confusing.

hbondoc:
I have configured and follow the instruction said in github , but i got this error when sending sms
 [WARNING] mod_sms.c:397 Context public not found
[DEBUG] mod_sms.c:477 SMS chatplan no actions found
[DEBUG] sofia_presence.c:225 Can't find registered user

PR #1607, opened 2016-05-21
Update extension.php to be able to write extension when save to disk preferred over lua xml handler

### LINKED ISSUE #1606: PostgeSQL error in v4+ when using save config/dialplan/directory to disk
The problem exists in app/extensions/resources/classes/extension.php and requires a trivial modification to the SQL statement in function xml on lines 175-179 when using postgerSQL. The datatype for v_voicemails.voicemail_id has changed to numeric but v_extensions.extension is still a text datatype, thus requires an explicit cast from numeric to text or vise versa in the where clause...
Otherwise the function fails silently and no extensions are written to the directory...

nyfon:
This is a pull request to fix #1606.
If you prefer the static directory files generation over lua xml handler it will fix the postgres datatype casting issues which wouldn't write v_{ext}.xml files...


markjcrane:
Nyfon your change looks great but you need to add yourself to https://github.com/fusionpbx/open-source/tree/master/contributors like other contributors have done. Then I will accept this change.

markjcrane:
623fa846e3ebd9ecf8f5f797a5f74be17ee2b895  to be safe as this pull request looks like it could bring other changes with it.

PR #1608, opened 2016-05-23
Fixed bug causing the DB field recording_file to be always empty

ronyshtamler:


markjcrane:
Merge conflict so updated manually record_file change by hand.

markjcrane:
What is the purpose to the add exec-set to vars.xml change? 
Have you tested with latest code to see if it is still needed?

markjcrane:
Please respond to the above questions so action can be taken if still needed.

ronyshtamler:
The exec-set is needed to allow a variable to be set by running a script. This allows setting the external IP address of machines in a cloud environment. In most cloud infrastructure, the external IP can be changed using the cloud dashboard. This allows two servers to share a DB and each have it's own IP address without using the switch-name.

PR #1609, opened 2016-05-23
Debian pkgs

kb3vgw:
Fixed deb pkgs when putting the files in /usr/share/examples they can not be chowned www-data:www-data


markjcrane:
Will consider an updated Debian package after the release.

PR #1614, opened 2016-05-25
BugFix-Default-MOH

mafoo:
Added basic searching for moh

added basic searching for the default moh directory
added app_defaults so moh is configured/repairable with app_defaults
changed to use PROJECT_ROOT
changed template replacement to be encased in comment so xml will
validate
changed to use preg_replace so leading spaces and tabs can be stripped
(to output clean xml)
local_stream_conf dos2unixed
made default search consistent (they should all be music/default or
music not a mixture)

### DIFF
--- a/app/music_on_hold/app_defaults.php
+++ b/app/music_on_hold/app_defaults.php
@@ -1,0 +1,36 @@
+<?php
+/*
+	FusionPBX
+	Version: MPL 1.1
+
+	The contents of this file are subject to the Mozilla Public License Version
+	1.1 (the "License"); you may not use this file except in compliance with
+	the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


markjcrane:
Great changes good job large improvement over what was there.

markjcrane:
Missed the forward slash on the beginning and last part of the pattern without it the xml file is empty.
$file_contents = preg_replace("/music\/default/", $default_moh_prefix, $file_contents);
$file_contents = preg_replace("/[\t ]*(?:)?/", $xml, $file_contents);

markjcrane:
I committed a fix for it.

mafoo:
Oups thank you, i keep forgetting the '/' inside preg_replace as i am used to using them as the string delimiter under perl

PR #1615, opened 2016-05-25
Added sticky GID permissions

mafoo:


mafoo:
This will only fix copying the group over. directories will still get created as user owner www-data
I am thinking that it might be better for me to add a new function make_dir that implements both the initial mkdir, but also performs the chown to freeswitch:{current_user} but the road block to doing that is normal users cant chown a fie/directory to someone else without sudos being updated

PR #1622, opened 2016-06-01
Update vars.xml

hblandford:
Add Australian ring tone


mafoo:
Just to let you know with https://github.com/fusionpbx/fusionpbx/pull/1590 now being accepted you should beable to choose au-ring as a ringback option on ivr/ring_groups

PR #1640, opened 2016-06-09
Update call_flow.lua

JoinFreeMe:
Allows to use an 'action' > 'set' > call_flow_sound_on=/{path_to_audio_file.wav}  and  call_flow_sound_off=/{path_to_new_audio_file.wav} in the dialplan of a call flow so that after a feature code has been dialled and the password has been entered a recording is played dependant upon the status of that call flow.

Example: (feature code *281 for call flow)
turn on call flow *281 > 1234# > my .wav file is played     "CALLS TO DIVERT"
turn off call flow *281 > 1234# > my .wav file is played     "CALLS TO OFFICE"

dial - enter password - status of call flow changes - recording plays dependant on true or false


moteus:
Looks familier :).
But really it is not ideal solution because call flow dialplan extension recreates each time you save it in Web interface so you have to add one more extension in dialplan or edit it each time.
Correct solution is add this fields to DB.

JoinFreeMe:
Working on the Web interface

moteus:
I mean when you save it in `App->Call Flow->Edit->Save` Fusion recreates extension in dialplan manager and it remove all changes you create in it.

JoinFreeMe:
I'll have a poke around, may request some help if that's alright?

PR #1642, opened 2016-06-09
Update call_flow_edit.php

JoinFreeMe:
Allows you to select a recording from a drop down menu so that when the status of a call flow is changed the recording can be played to tell the user what the current status is.

E.G 

*280
1234#
(plays "calls to divert")

*280
1234#
(plays "calls back to office")

Relying on entry's in the db in v_call_flows being "call_flow_sound_on" and "call_flow_sound_off"

### DIFF
--- a/app/call_flows/call_flow_edit.php
+++ b/app/call_flows/call_flow_edit.php
@@ -22,6 +22,7 @@
 
 	Contributor(s):
 	Mark J Crane <markjcrane@fusionpbx.com>
+	Lewis Hallam <lewishallam80@gmail.com>
 */
 require_once "root.php";
 require_once "resources/require.php";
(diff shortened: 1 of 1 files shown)


moteus:
I see few problems:
1. You have also edit [app_config](https://github.com/fusionpbx/fusionpbx/blob/master/app/call_flows/app_config.php#L131-L134) file to add new fields.
2. You buld wring drop-down list. You need to audio files and greetings. But you create dialplan ones. I do not know any good way to do this except copy/past e.g. from [here](https://github.com/fusionpbx/fusionpbx/blob/master/app/fax/fax_edit.php#L679-L789) and edit it. 
(@markjcrane is there exists better way?)
3. You add `call_flow_sound_off=/usr/local/freeswitch/recordings/..` to dialplan. You really do not need this. Better use data from DB in Lua script itself. and also do not use absolute path (I use Windows and this is definitely will not work for me :) ) So better just lets php put data to DB.

JoinFreeMe:
Shall start working on this tomorrow @moteus as there are a few things I need to get on with at the moment.

Thank you for the comments, I didn't even think about taking into consideration a windows environment

markjcrane:
Need to follow these instructions before we can consider accepting your pull requests.

PR #1648, opened 2016-06-10
Fix the xml_cdr class

daniel-lucio:
I found a double utilization of the variable $row, this was preventing the correct fill of the variable $this->array, causing everything to fail using this class.

### DIFF
--- a/app/xml_cdr/resources/classes/xml_cdr.php
+++ b/app/xml_cdr/resources/classes/xml_cdr.php
@@ -22,6 +22,7 @@
 
 	Contributor(s):
 	Mark J Crane <markjcrane@fusionpbx.com>
+	Luis Daniel Lucio Quiroz <dlucio@okay.com.mx>
 */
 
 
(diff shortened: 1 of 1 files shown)


markjcrane:
I try not to abbreviate much in the code to make it easier to read the code.

daniel-lucio:
It took me 4 hours of debugging to find the root cause. This is only a clean fix, i striped out all the coding i did to trace the variables and found why it was not working.

At least a thank you

markjcrane:
Thanks for identifying the bug that is a good thing and an approach to fix the bug.

Your fix worked fine. However after thinking about it a bit more and in the pursuit of perfection I think the best fix for this is to prevent this problem in the future by changing the method xml_array from using $row to $key. '$row' is what is a standard in the majority of the foreach functions that loop through an array. $key is more descriptive for the purpose the variable is used. I think by changing the variable name to something more intuitive it will prevent future bugs if a foreach is needed in this method in the future.
https://github.com/fusionpbx/fusionpbx/commit/11efb357125f0d3f30ddd91f99562bdd0db2ae0d

PR #1651, opened 2016-06-11
Update app_languages.php

JoinFreeMe:
Multilingual support for new features


AgregarDeveloper:
@markjcrane Can we revert this pull along with #1652 ? It has changed UTF-8 encoding, and broke almost all translations.

PR #1652, opened 2016-06-11
Update call_flow_edit.php

JoinFreeMe:
Allows the page to become multilingual


AgregarDeveloper:
@markjcrane Can we revert this pull along with #1651 ? It has changed UTF-8 encoding, and broke almost all translations in this file.

PR #1658, opened 2016-06-14
Fix. Display key type in `Profile` app.

moteus:


moteus:
Did you mess this.
Without this PR
screenshot_4

With this PR
screenshot_3

PR #1660, opened 2016-06-14
Fix. Remove duplicate options from internal profile.

moteus:


markjcrane:
Are these duplicates? I know they are settings that could be a disruptive if not disabled.

moteus:
They appear twice on web.
Same as caller-id-type. But it has different values so you can enable one value so it may be useful.

moteus:
screenshot_1

markjcrane:
New duplicates should be handled. I can see these are dangerous settings so I will comment them out instead of deleting them. Here is the reference to the commit 8909639101366f9ea43a717235a14eec90573205

PR #1665, opened 2016-06-15
Fix. Display correct info for extensions with number-alias.

moteus:
I think to use some think like 
```PHP
if (permission_exists('number_alias') ) {
  $sql .= "or (e.number_alias is not null and (caller_id_number = e.number_alias or destination_number = e.number_alias)) \n";
}
```
But not sure it correct way.


markjcrane:
The 'or' is incorrect and would break the query because it needs to be grouped by ( and ). I'll commit it then fix it.

PR #1666, opened 2016-06-15
Update app_languages.php

JoinFreeMe:
Filled in all language fields with the appropriate translation.


app/system/app_languages.php
```diff
@@ -91,74 +91,74 @@
...
-$text['label-path']['uk'] = "";
-$text['label-path']['de-at'] = "";
+$text['label-path']['es-cl'] = "Ruta del proyecto";
+$text['label-path']['pt-pt'] = "Caminho do projeto";
+$text['label-path']['fr-fr'] = "Chemin du projet";
+$text['label-path']['pt-br'] = "Caminho do projeto";
+$text['label-path']['pl'] = "Ścieżka projektu";
+$text['label-path']['sv-se'] = "projekt Path";
+$text['label-path']['uk'] = "шлях проекту";
+$text['label-path']['de-at'] = "шлях проекту";
```
moteus: I think this is wrong

markjcrane:
I think this is right. Translators can come in for each language and improve on them at a later date.

moteus:
I mean only like `$text['label-path']['de-at'] = "шлях проекту"` I belive this is not German :)

markjcrane:
Ah well you are right on that for sure.

JoinFreeMe:
Commit was made to correct this

JoinFreeMe:
And uk is Ukraine I believe as informed by mcrane

moteus:
Uk is united kindom. Ua is ukraine

JoinFreeMe:
In which case I'll commit again when I get back to a machine

reliberate:
Does Google have it wrong then?  See URL...   https://translate.google.com/m/translate#auto/uk/Happy

moteus:
Seems you are right. https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes.

markjcrane:
I see we are in agreement. This is also an interesting link on this topic ISO standard for UA as well as alternative ISO standard for UK. United Kingdom is actually GB.
http://www.loc.gov/standards/iso639-2/php/code_list.php

JoinFreeMe:
Thank you, I've bookmarked this for future reference :)

PR #1668, opened 2016-06-15
Add option to see if an extension is currently registered on the extensions page

blackc2004:
controlled by a permission and is off by default but can be enabled to
see what extensions are registered.


blackc2004:
Just ignore this for now but please don't close it. I will create a clean one.

PR #1683, opened 2016-06-20
Enhance-Fallback for logos

mafoo:
fallback to the theme > logo if menu_brand_image isn't set
this should restore logos for people who haven't set both settings


markjcrane:
You are making an assumption here. menu_style=inline would make it show the logo as they are used to seeing in older versions of fusionpbx.

PR #1687, opened 2016-06-20
Enhance-Rework GUI Install

mafoo:
reworked installer to use a iframe to show progress updates during the install
- [x] Tested on Debian 8.5 VM


markjcrane:
In the process of reworking the install.

PR #1697, opened 2016-06-22
Add. Class described configuration of device vendors

moteus:
screenshot_3


markjcrane:
Not sure I like adding escene class as that then implies we should be consistent and have one for every manufacturer. I'll review more before making a final decision.

moteus:
This is i like it. So it possible just implement class for each vendor. And it easy to add new one with minimal influence to another, May be it possible define common API for such class and do it automatically. I am not sure that my class has best API. I think more about using some think like `device::escene` but not sure how to do this in practice.

moteus:
*NOTE* This require PHP>=5.6.

moteus:
Add support PHP 5.3

markjcrane:
I think I like these better in an array that is picked up from the template directory. So they only exist in the code if the template for that manufacturer is loaded on the system.

moteus:
I am not sure how to describe/define in php search path to such files.

markjcrane:
I added the array via php code in the devices/app_config.php file. The array is used to load the device vendors and device vendor functions tables. This alllows disabling the vendors or functions that you don't want to see in the gui and allowed to assign groups to the vendor functions overall this provided a great deal of flexibility which is why I went with that method instead of hard code in to php classes.

PR #1708, opened 2016-06-26
Fix. generate correct config file.

moteus:
Sorry. Just found this.


markjcrane:
I just accepted your fix.

PR #1715, opened 2016-06-27
Enhance-Warn if posted data to big

mafoo:
check if the posted data exceeds php's configured maximum
- [x] Tested on Debian 8.5 VM

### DIFF
--- a/resources/header.php
+++ b/resources/header.php
@@ -125,6 +125,53 @@ require_once "resources/require.php";
 //start the output buffer
 	ob_start();
 
+//check for posted data to big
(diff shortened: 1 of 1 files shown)


markjcrane:
"The lowest limit of any related setting supersedes a higher setting:"
    post_max_size, upload_max_filesize, and memory_limit
I usually update upload_max_filesize. I'm concerned about the interaction between the multiple settings.

mafoo:
Ahh! would you like me to do some more checking and warn which of any of the three is to low?

markjcrane:
Maybe be good to have the install increase the size automatically in the php.ini file.

mafoo:
Pushed update that now checks the three suggested values
- [x] re-Tested on Debian 8.5 VM

markjcrane:
Since size is added to the php.ini by default I think that is sufficient.

PR #1727, opened 2016-07-01
Update code to iterate over numbers.

moteus:
This code fixes some problems
- range like `009-010` current code produce numbers `009` and `00010`
- range like `200-100` now raise error so it will be easy to debug
- range like `010-20` now generate error.
- `destination` is string so condition `destination == tonumber(sip_from_user)` is always false so change it to `destination == sip_from_user`

PS. simple way to implement `leading_zeros` is just `string.match("^(0*)", str)`


markjcrane:
Please update the pull request to fix the conflict so it can be accepted.

PR #1737, opened 2016-07-06
More efficient and compatible SQL query

daniel-lucio:
Don't do a nested select, it adds unnecessary overload to the db. Instead, use a left join, quicker and more compatible among db types.


markjcrane:
Maybe MySQL should add support for this capability oh but its owned by Oracle so it will likely never get it.

PR #1738, opened 2016-07-06
Enhance Voicemail Emails

mafoo:
* Renamed template variable account to voicemail_id and adjusted all templates
* Added ${voicemail_description}
* Added ${voicemail_name_formatted} (will render the voicemail identifier in accordance with  Default/Domain Settings > Voicemail > display_domain_name)
* Updated templates to be consistent spacing
* Updated templates to include use of ${voicemail_name_formatted}
* Changed ${account} to be voicemail_name_formatted instead of id

### DIFF
--- a/core/default_settings/app_defaults.php
+++ b/core/default_settings/app_defaults.php
@@ -364,6 +364,13 @@ if ($domains_processed == 1) {
 		$array[$x]['default_setting_value'] = '90';
 		$array[$x]['default_setting_enabled'] = 'true';
 		$array[$x]['default_setting_description'] = 'Maximum length of a voicemail greeting (in seconds).';
+		$x++;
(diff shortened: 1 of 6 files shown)


DigiDaz:
I have tested this, it seems to work exactly as expected:

[http://picpaste.com/pics/voicemailtest-hUEbCNoE.1469177216.jpg](http://picpaste.com/pics/voicemailtest-hUEbCNoE.1469177216.jpg)

mafoo:
my only concern in hindsight at this point is I renamed the ${account} to ${voicemail_id} at mark's request, this will mean that custom templates will start to render wrong. Should I either
a) reinstate ${account} as the contents of ${voicemail_name_formatted}
-or-
b) add code to app_defaults that scans all existing templates and reports any using the old template variable?

PR #1753, opened 2016-07-14
Enhance-Add Ping to Registrations Status

mafoo:
Retrieve ping-time and display it

display `&nbsp;` only when value is empty
- [x] Tested on Production machine


blackc2004:
Nice! 👍

mafoo:
for reference ping is the 'SIP endpoint ping response time'
see https://freeswitch.org/jira/browse/FS-8042

PR #1762, opened 2016-07-20
add default settings to dashboard for auto-answer, record, ringback

blackc2004:


markjcrane:
Notice on the dashboard the link is using app/click_to_call therefore the default category would be preferred to be 'click_to_call' then the settings could be re-used for other pages other than just the dashboard when click to call is being used.

blackc2004:
Sure, no problem. I thought about adding a category for click_to_call but since it was on the dashboard I just put it under there. I'll update the PR.

PR #1772, opened 2016-07-22
Fix my PR for the device registered on extensions page.

blackc2004:


DigiDaz:
Seems to work as expected.

http://picpaste.com/pics/extensionregistered-wjnmh3CI.1469516373.jpg

markjcrane:
It is a nice feature request my concern is how well it will scale when there are a lot of registered extensions. It would be great to have a permission to enable or disable it like a permission name 'extension_registered'. This would help protect the system if there were a problem with performance.

Column name would be better as 'Registered' so it doesn't take much horizontal space and should be just before the description.

blackc2004:
@markjcrane, there is a permission called "extension_show_registered" which is disabled for everyone by default. That way it wont' impact larger systems. 

I've changed the name to "registered"

PR #1775, opened 2016-07-25
Add `service` script and make timout optional for EventConsumer ctor

moteus:
Change. Pass pid file first. Because there really no need pass timeout.
Add. `fsc` script to be able shutdown MWI and Call Flow subscribe services.

With `fsc` script timeout is really no need any more because it control how often check pid file on file system. But now we can use FS events to control service and make it more responsive.

screenshot_4


markjcrane:
FSC is not an intuitive name can you think of a more intuitive name.

moteus:
I agree, may be just `services` or `svc`

markjcrane:
I like 'service' as that matches linux before systemd.

PR #1776, opened 2016-07-26
Allow * 22 login/logout of queue without password by adding variable …

DigiDaz:
…auto_authorized to dialplan

Add auto_authorized=true to the agent_status dialplan entry before "app.lua agent_status" is called.

### DIFF
--- a/resources/install/scripts/app/agent_status/index.lua
+++ b/resources/install/scripts/app/agent_status/index.lua
@@ -29,6 +29,7 @@
 			uuid = session:get_uuid();
 			agent_id = session:getVariable("agent_id");
 			agent_password = session:getVariable("agent_password");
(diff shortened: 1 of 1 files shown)


markjcrane:
I will commit this giving you credit for the idea and then will add a commit afterwards that handles it differently. This pull request doesn't take into account where auto_authorized if you set the value to false, no or anything it will assume true. My commit will use agent_authorized and looks specifically for true.

DigiDaz:
mcrane no problem, I'm not interested in credit for anything, I was well aware of it accepting anything other than nil, I just thought I'd make it easy. I'll give it a shot at tidying it though, I know you are busy with bigger fish. Good idea on the agent_authorized :)

markjcrane:
The changes are already done and commited look at the history for that fIle to see what I did.

PR #1778, opened 2016-07-27
Simplify `is_uuid` function.

moteus:
Also because it returns now value itself it possible write
```Lua
my_uuid = is_uuid(value1) or is_uuid(value2)
```


markjcrane:
My version of the function returns boolean.  I had need for this function for languages in xml handler. is_uuid function also matches a function with the same name in the php code.

PR #1781, opened 2016-07-28
Simplify `is_uuid` function.

moteus:


markjcrane:
Can you explain the reason for 'not not'
"return not not string.match(s, pattern)"

moteus:
it convert any value to boolean. e.g. `not not nil` is `false` same as `not not false`
Other is `true` e.g. `not not 'hello'` or `not not 0`.
its short version of 
```Lua
f = not not value
-- same with `if`
if value then f = true else f = false end
```

PR #1783, opened 2016-07-28
Add. Access list of IP an extension can register from.

moteus:
Notes.
1. When use fs_path=true also need create ACL with name `fs_path` and add to it all FS (https://github.com/fusionpbx/fusionpbx/compare/master...moteus:extension_access_acl?expand=1#diff-7ea5ac35f2e54352f73910f11d0c0a24R561)
2. This PR adds one more requiest to `user_data` when used data from cache (https://github.com/fusionpbx/fusionpbx/compare/master...moteus:extension_access_acl?expand=1#diff-7ea5ac35f2e54352f73910f11d0c0a24R551)


moteus:
There already exists Auth ACL wich works similary.
Checkout [ACL](https://wiki.freeswitch.org/wiki/ACL#Users) documentation.

PR #1790, opened 2016-08-02
Change. Use `service::control` event to control services

moteus:
This is more FS way. E.g. sofia sends `sofia::register` event
and add all information to headers. So now `service` script
emit `fusion::service::control` event and each service responsible
for test its own name. This also allows add in future evnets to
e.g. monitor service status like `fusion::service::satus` so it
will be possible write service which will be restart services.


markjcrane:
Would be better if  local service_name = "flow" is called  local service_name = "call_flow" as we don't have a feature called flow but we do have call flow.

PR #1791, opened 2016-08-03
Enhance-Add message stack

mafoo:
Support legacy $_SESSION['message_*']
Use new $_SESSION['messages'] array
New function add_message($message, $mood, $delay) to simplify adding messages
- [x] Tested on CentOS 7 Host firefox client

I think the functional code is complete, all that is left is cosmetic tweaks as required and testing on other platforms

### DIFF
--- a/resources/functions.php
+++ b/resources/functions.php
@@ -1923,4 +1923,10 @@ function number_pad($number,$n) {
 			return false;
 	}
 
+	if (!function_exists('add_message')) {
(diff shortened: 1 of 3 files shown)


markjcrane:
Its the CSS changes that have held up this commit. Will test it when I get a chance.

markjcrane:
Submit this over again without the CSS changes then it could be accepted.

PR #1795, opened 2016-08-04
Fixed scripts to get multiserver memcache working

DigiDaz:
Freeswitch 1.6 does not seem to recognize the MEMCACHE event. This commit changes the event to an event of type CUSTOM and also introduces a new script ha_monitor.lua that can also currently start and stop gateways across all servers.

### DIFF
--- a/resources/classes/cache.php
+++ b/resources/classes/cache.php
@@ -89,8 +89,8 @@ class cache {
 
 		//send a custom event
 			$event = "sendevent CUSTOM\n";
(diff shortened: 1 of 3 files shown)


markjcrane:
There is 3 versions of this file I have to consider each one.

markjcrane:
This script will be renamed and moved out of the main scripts directory. As all scripts in the main scripts directory will be moved to a sub directory to organize the code better.

PR #1802, opened 2016-08-08
Time of day routing convertion fixed.

SJS87:
This is a minor bug which is fixed with a few minor adjustments.

Valid values for 'minute-of-day' in FreeSWITCH are 1-1440, the page is currently converting 'time-of-day' to 'minute-of-day' as 0-1439, the fix will correct the values to 1-1440.

Sources:
https://wiki.freeswitch.org/wiki/Time_of_Day_Routing


markjcrane:
I just backed out these changes as it broke upgrades. I think I have solved this another way by using time-of-day incrementing by one minute increments instead of 5 minutes. Then you can set it to exactly the minute you want. Makes the select list longer but its more flexible this way.

PR #1811, opened 2016-08-12
BugFix-Support spaces in streamfile.lua

mafoo:
use table.concat as a workaround to spaces in filesames
alternative is adjust all destiantions to quote filenames, but would
require app_defaults catch to fix all exisiting written destinations.

### DIFF
--- a/resources/install/scripts/streamfile.lua
+++ b/resources/install/scripts/streamfile.lua
@@ -1,6 +1,7 @@
 --get the argv values
 	script_name = argv[0];
-	file_name = argv[1];
+	file_name = table.concat(argv, " ");
+	freeswitch.consoleLog("notice", "[streamfile] file_name: " .. file_name .. "\n");
 
 --include config.lua
 	require "resources.functions.config";


resources/install/scripts/streamfile.lua
```diff
@@ -1,6 +1,7 @@
 --get the argv values
 	script_name = argv[0];
-	file_name = argv[1];
+	file_name = table.concat(argv, " ");
```
mafoo: note: table.concat seems to be ignoring item 0 so this correctly does what we want

moteus: May be better enquore file name when call this script?

mafoo:
I did consider en quoting values before calling this script but that would be quite a deep modification as every app that creates a streamfile line would need updating and I found during testing it would also mean values would need to be rechecked, the most obvious to this problem was app/ivr which would require heavy modification to detect and correct wrong values.
This solves a edge case where a file has been uploaded with a space in it, where as implementing quoting of all files could cause a lot of knock on bugs as people upgraded.

PR #1815, opened 2016-08-15
Added a key profile edit and add button

minotaur01:
Added a key profile edit and add button for easier navigation, can now hop to profile editing with one click.


markjcrane:
This is a good idea. It would be even better if edit was hidden until a profile has been selected.

minotaur01:
LOL, okay i'll make the change later

PR #1816, opened 2016-08-15
add Diversion Header to dail string

de-jens:
When a call gets forwarded you will see form witch Extension it came from.

https://tools.ietf.org/html/rfc5806#section-6.1

### DIFF
--- a/app/calls/resources/classes/call_forward.php
+++ b/app/calls/resources/classes/call_forward.php
@@ -76,6 +76,7 @@ include "root.php";
 					$dial_string .= ",domain=".$_SESSION['domain_name'];
 					$dial_string .= ",extension_uuid=".$this->extension_uuid;
 					$dial_string .= ",toll_allow='".$this->toll_allow."'";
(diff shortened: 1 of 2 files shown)


markjcrane:
You need to add yourself as a contributor here so that we can accept your pull request.

bigfishvoice:
This seems to have broken the ability to do a *72 from an end point.  Receive the following log file entry when trying to enable forwarding:
2016-08-21 13:43:44.061987 [ERR] mod_lua.cpp:203 /usr/local/freeswitch/scripts/call_forward.lua:250: unexpected symbol near char(226)

If I remove the following line from the call_forward.lua script, everything starts working again.
dial_string = dial_string .. “,sip_h_Diversion=<"..extension.."@"..domain_name..">;reason=unconditional";

bigfishvoice:
DigiDaz figured it out on IRC.  The first double quote is not a standard ASCII quote and breaks the script.  Replace it with a stand double quote and all is well.  Please commit a fix.

markjcrane:
Thanks for reporting the issue the double quote has been fixed.

PR #1828, opened 2016-08-20
Added functionality to show when a devices was last provisioned

minotaur01:
Added functionality to show when a devices was last provisioned

Need addition to database: v_devices table:
device_provisioned_on -> datetime
device_provisioned_by -> char(10)


minotaur01:
Disregard made changes

PR #1838, opened 2016-08-24
Update {$mac}.cfg

Len-PGH:
t20p


blackc2004:
t20p but this is the cp860 file and the cp860 doesn't have expansion.

PR #1850, opened 2016-08-24
Corrects some cases where dialplan conditions were not fully handled

nostikj:
In trying to write a dialplan I ran into some cases where the XML was not correct.  Since a dialplan condition statement could be in one of three states: (1) already output to XML opening a
<condition ...> block, (2) currently saved in var condition, or (3) one or more time conditions saved
in var condition_attribute, the main change is to check/handle all three possible cases at those
places where conditions are finalized.

If a dialplan group only consisted of a condition, then the output XML would become just
a </condition> entry without any opening <condition ...> entry.

If a time-based condition was following by a non time condition, the time condition would be
completely lost.

If a time-based condition, in a new dialplan group follows an action statement, the time condition
would get treated as a non-time condition (i.e. condition_type was being cleared).

If a dialplan was entered (incorrectly) without any condition statement, and only action statements,
the XML <extension> would not get closed at the start of a new dialplan uuid, and thus the erroneous
dialplan uuid and the following uuid would be merged into a single extension.

### DIFF
--- a/resources/install/scripts/app/xml_handler/resources/scripts/dialplan/dialplan.lua
+++ b/resources/install/scripts/app/xml_handler/resources/scripts/dialplan/dialplan.lua
@@ -152,16 +152,21 @@
 
 				--close the tags
 					if (condition_tag_status ~= "closed") then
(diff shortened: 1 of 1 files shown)


markjcrane:
Great improvements well done.

PR #1873, opened 2016-08-28
user dashboard generates errors in httpd error log

nyfon:
Caused by not properly escaped shell_exc. Also, df generates error as well when /home directory is not mounted.


nyfon:
Hmm, I'm not sure why it wants to pull all 4 commits...

mafoo:
I would recommend you perform a rebase against fusionpbx/master branch with squash-all.
If you are using windows it is recommended to also install TortoiseGit as it makes managing some changes easier. If you are on IRC at the same time I can walk you though it

PR #1875, opened 2016-08-29
Make debugging easier

DigiDaz:
Change access denied to more informative message to assist debugging

### DIFF
--- a/app/provision/index.php
+++ b/app/provision/index.php
@@ -229,7 +229,7 @@ The file name is fixed to `Account1_Extern.xml`.
 
 //check if provisioning has been enabled
 	if ($provision["enabled"] != "true") {
-		echo "access denied";
+		echo "provisioning disabled";
 		exit;
 	}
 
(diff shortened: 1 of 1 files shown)


markjcrane:
I just noticed you are not in https://github.com/fusionpbx/open-source/tree/master/contributors please complete as others have done. Once you have I will accept your changes. This was an oversight maybe assumed you had already done it.

moteus:
Does it reduce security?

DigiDaz:
moteus, it just changes the two access denied messages to something a little more informative. Its mainly to help people on IRC who are seeing this message. I don't see any security issue.

moteus:
I mean that attacker then will know that there turn on digest auth.
I think better just return 404 in all cases and not return 200 with text description.

DigiDaz:
Moteus,

I would guess if he is savvy enough to be making such an attack, he would also be having the sense to view the source code and craft the attack accordingly anyway.

moteus:
If server will return 404 attacke will know even if there provision exists.

markjcrane:
Its not descriptive specifically to prevent telling hackers useful information.

PR #1877, opened 2016-08-30
TFTP Service 1.0.7

minotaur01:
Fixed linux and windows run as service issues
Fixed linux systemd service self install issues
Modified windows service self install process


minotaur01:
mcrane, figured it out! :) Nice clean request

PR #1878, opened 2016-08-30
Device edit mac input filter

minotaur01:
Added key press filter for invalid characters
Added automatic striping of invalid character on paste

### DIFF
--- a/app/devices/device_edit.php
+++ b/app/devices/device_edit.php
@@ -668,7 +668,7 @@
 	echo "</td>\n";
 	echo "<td class='vtable' width='70%' align='left'>\n";
 	if (permission_exists('device_mac_address')) {
-		echo "	<input class='formfld' type='text' name='device_mac_address' id='device_mac_address' maxlength='255' value=\"$device_mac_address\"/>\n";
+		echo "	<input class='formfld' type='text' name='device_mac_address' id='device_mac_address' maxlength='12' value=\"$device_mac_address\"/>\n";
 		echo "<br />\n";
 		echo $text['description-device_mac_address']."\n";
 	}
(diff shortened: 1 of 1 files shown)


DigiDaz:
I may be wrong but I believe you should be able to enter the device mac address with colons?

If this is the case, won't:

echo "	<input class='formfld' type='text' name='device_mac_address' id='device_mac_address' maxlength='12' value=\"$device_mac_address\"/>\n";

prevent that?
Should maxlength at least not equal 17?

minotaur01:
That is true, max length 12 would prevent that, but do the mac not get saved without the colons and dashes anyway?

That's why there is a javascript at the bottom that removes all colons, dashes and white space on paste, to normalize the mac, and now that i think about it it should also remove all other non valid characters and  lower case it.

DigiDaz:
I didn't realise it was javascript, I thought it was cleaned up after it was submitted

markjcrane:
I don't like it being cleaned up by javascript should be done server side. 

Also this is not always going to be a mac address so might be renamed to device id as some soft phones do not use a mac address.

markjcrane:
It is stripped on the server side with $device_mac_address = strtolower(preg_replace('#[^a-fA-F0-9./]#', '', $device_mac_address)); I don't like it being done by javascript as it can easily be bypassed. Also because this doesn't have to be a real mac address for soft phones and for hot desking. Closing the pull request.

minotaur01:
K, but it's all the provisioning code based on a mac? And since you are all ready filtering it on the server based on the mac format i didn't think it would be a issue to do it right away on the client side. 
Also the next step to this was to have the client side automatically filter out keys and templates for the specific vendor based on the mac!

PR #1882, opened 2016-08-30
add username/password to the yealink profiles.

blackc2004:


markjcrane:
Looks like you found a better way to do this than I was thinking of. Good Job

PR #1894, opened 2016-09-02
CDR Display functionality change

minotaur01:
Dont make the row clickable if the extended data field (json) is empty


stefanomarty:
Cannot click any cdr row anymore after this commit, looks like:

> (isset($row['json'])

is always false in my system, and:

> (empty($row['json'])) is true
> (property_exists ( $row , json )) is false

minotaur01:
stefanomarty, dose the field have data and what database are you using?

minotaur01:
stefanomarty, the issues is not the code, its that the json field is not being pulled, an oversight on my part. :( so the fix would be to either code the field to be pulled or to undo the change, i will talk to mark and fix it

stefanomarty:
minotaur01, thanks for the update! I do not know Php, just some programming basis. I was just trying to understand why I couldn't see the CDR details anymore after this commit. Not a big problem, as the CDR details are rarely used. :-)

I'll wait for your fix, and then confirm if the problem is solved in my system (I'm using SQLite).
Bye.

PR #1896, opened 2016-09-03
Added Aastra Function Keys And Fixed Update Logic

minotaur01:
Added the rest of the aastra function key options to app_config. And to
my surprise the functions would not save to the database, found the
problem to be the app_defaults logic would only allow adding on a fresh
install. So I fixed the logic to work on a update also.


stefanomarty:
I do have an issue with this commit, I get this error after uppgrading:

`Fatal error: Call to a member function bindValue() on boolean in /usr/share/fusionpbx/app/devices/app_defaults.php on line 128`

It happens whenever I try to run this Advanced menu:
Advanced --> Upgrade --> App Defaults

I tried to follow the logic, but it's way too complicated for me. My system is running on SQLite database, I checked the "v_device_vendor_function_groups" table, here's what I get from phpLiteAdmin:

schermata 2016-09-06 alle 10 40 46

The table is empty. I can send more details if needed.

Best.
Stefano.

minotaur01:
Morning,
Could you open the php  editor and then open file /usr/share/fusionpbx/app/devices/app_defaults.php scroll down to line 128 and copy and paste it in here.  Just want to make sure its the same line on my system. Also what linux distribution are you using? From the file path it looks like FreeBSD?

stefanomarty:
Sorry to be late, looks like the issue was fixed by some later commit, as it's not present anymore. This is what I see (now) in /usr/share/fusionpbx/app/devices/app_defaults.php but I'm afraid it's not the same source I was seeing a few days ago:

```
//add the device vendor function groups
	if (is_array($function['groups'])) {
		$sql = "insert into v_device_vendor_function_groups ";
		$sql .= "(";
		$sql .= "device_vendor_function_group_uuid, ";
		$sql .= "device_vendor_function_uuid, ";
		$sql .= "device_vendor_uuid, ";
		$sql .= "group_name, ";
		$sql .= "group_uuid ";
		$sql .= ") ";
		$sql .= "values ";
		$i = 0;
		foreach ($function['groups'] as $group_name) {
			if ($i == 0) { $sql .= "("; } else { $sql .= ",("; }
			$sql .= "'".uuid()."', ";
			$sql .= "'".$device_vendor_function_uuid."', ";
			$sql .= "'".$device_vendor_uuid."', ";
			$sql .= "'$group_name', ";
			$sql .= "'".$group_uuids[$group_name]."' ";
			$sql .= ")";
			$i++;
		}
		$db->exec($sql);
	}

````

I'm quite sure line 128 was the same in my system as it was here on Github. My system is running under Debian 8 Jessie, FusionPBX was installed by script in /usr/share.

If you need any further information I'll be glad to help.otherwise I consider this issue closed. Thanks for your help, best regards.
Stefano

markjcrane:
Sorry I forgot that I worked on it.

PR #1898, opened 2016-09-05
Optimized the permissions mechanisim

minotaur01:
Changed the permissions array to a named array and stop the system from
looping thru the items.


markjcrane:
Good point the original method was slower than what you have in this pull request. I also switched it to a named array and then looked up the value with the name. If you think your method is faster than the change I just saved we should test it and whichever is faster will be the one we use.

PR #1899, opened 2016-09-05
use extension rather than username

daniel-lucio:
username is set to alias if it exists, otherwise it falls back to the extension value.
sofia_contact does not work with aliases. since we have the extension value already filled, it is safe to use it


daniel-lucio:
Please note this is for 4.0

PR #1900, opened 2016-09-05
Device Templates v2.0.0

minotaur01:
Inital code for databased device templates and provisioning


markjcrane:
Can you clean this up so its only the new changes. Git rebase perhaps.

minotaur01:
Morning Mark,
Rebasing would delete all the change i made. I will delete the pull request and create a new once everything is running right. 

For testing purposes you can just download my branch in a zip.

https://codeload.github.com/minotaur01/fusionpbx/zip/device-templates

mafoo:
rebase should not destroy your changes if done correctly.
I have written a guide how to rebase to squash your commits and preserve changes how fusionpbx likes commits to be. if you are concerned it will mess up just don't perform the last step (git push --force-with-lease) and check over what it do carefully before either pushing(with force) or fetching back your original (git fetch --all; git reset --hard HEAD)

minotaur01:
Thanks mafoo, i will try that!

markjcrane:
I'm worried about backwards compatibility and taking provisioning in the wrong direction.
You have some good idea mixed in with things that I don't agree with. I will need to test your latest version before making a final decision whether to include this or not.

minotaur01:
Yes i understand that thats why i made a bunch of changes after we spoke on the phone. Ill be around this weekend if you have time, i can explain all the changes ive made.

markjcrane:
If you want it considered call me and can use a remote session to show me around the recent changes.

PR #1912, opened 2016-09-12
add http_auth for remote phone book, fix dns issues

blackc2004:


blackc2004:
A couple issues fixed here:
1) set the favorite_setting url to be the same in all files
2) added http_auth support for all remote phone books files. Will work with http auth on or off
3) fixed the network.static_dns_enable which caused some phones to not get DNS on upgrade via provisioning. This change is done per yealink tech support.
4) fixed the linksys spa942 file to not have hard coded DNS and NTP settings.

PR #1920, opened 2016-09-14
Fix load balancing in 4.3

daniel-lucio:
Same rationale as 4.0


markjcrane:
export_vars=domain_uuid,domain_name might be a better way to do this as the value already exists.

moteus:
Set export_vars do not add vars ro export lis but replace it. So it will remove all other exported vars

PR #1921, opened 2016-09-15
Fix e-mail attachment handling / conversion & allow local emails to be valid

hargut:
- Since $fax_name is modified the code didn't work properly, as the filenames didn't change.
The whole code works now fine, when renaming the attachment file to $fax_name.
- local email addresses where not valid, since the valid_email checker required a domain with 2-6 characters.


markjcrane:
Please signup as a contributor so we can accept your pull request.

PR #1929, opened 2016-09-18
Update macro.lua to fix reading voicemail extension number

jebsolutions:
Expected result: The person at extension XXX is unavailable ...
Actual result: The person at is unavailable... 
i.e. it doesn't say the extension number.

bug: trying to access local variable voicemail_greet_id outside it's scope...so it doesn't exist and you get this in the logs:

2016-09-18 12:32:54.615249 [ERR] switch_cpp.cpp:977 Error! invalid args. 

Fix moves local variable creation outside the inner "if" so it's available for the later table.insert


markjcrane:
You need to signup as a contributor using these instructions so that we can accept your pull request.

PR #1933, opened 2016-09-19
[4.2] Fix e-mail attachment handling / conversion & allow local email…

hargut:
…s to be valid (#1921)

* Enable local email addresses to be valid. E.g required for local Fax submission.
* Fix filenames & allow spaces in filenames.
* Fix attachment file names.
* Revert "Fix filenames & allow spaces in filenames."


markjcrane:
So this is same as previous pull request but for the 4.2 release branch correct?

hargut:
Yes, exactly. Just in a single commit.

PR #1941, opened 2016-09-23
t40p yealink template

Len-PGH:
copied from the t41p templates and edited for y000000000036.cfg


konradSC:
The file needs to be called y000000000054.cfg. 

(it may be called that. Just trying learn git )

btw...This is Brian from class   :)

Len-PGH:
hi Brian :+1:

PR #1954, opened 2016-09-26
Look further for domain_name

daniel-lucio:
Same rationale as 4.0


markjcrane:
Looks like a good idea.

PR #1960, opened 2016-09-29
add built in ability for microsoft bing speech to text

blackc2004:
add features to support microsoft bing speech to text for transcription services on voicemail… can be extended easily to add any other transcription service and also the transcription can be used elsewhere (such as adding it to the db)


blackc2004:
FYI, if you're set to record in MP3 we cannot use mod_shout because transcription service requires a high quality recording to get the best results, so we force it to re-encode as lame after transcribe happens.

markjcrane:
All file names are required to be lower case. Right or wrong consistency is a good thing.

blackc2004:
Ah, sorry. I'll fix and update the PR.

markjcrane:
For Lua json we should use this library from Travis Cross.
https://github.com/traviscross/tc-lua-json/blob/master/json.lua

Its likely you know who he is but if not look at the speaker schedule for Travis at this website address.
http://cluecon.com/

blackc2004:
Mark, I updated the file name and moved it to the general functions directory. I emailed Travis last night and I haven't heard back from him. I picked this json.lua class that I used because it's been in active development and has tons of great documentation. I think that this would be best to use as the one from Travis hasn't been updated in 2 years and I couldn't find any documentation on it. Would hate to get into a situation where there's an issue and there's no support/etc for the library.

Let me know if that's ok. Otherwise I will try to contact Travis some more.

markjcrane:
Will have to study the creative Commons to decider whether I consider it compatible with fusionpbx project  before I will accept the pull request.

markjcrane:
Mit and BSD license are preferred because they easiest to comprehend the requirements. Creative Commons has different flavors.

blackc2004:
Mark, I replaced the json.lua with lunajson which is MIT license: https://github.com/grafi-tt/lunajson/blob/master/LICENSE

Hopefully this is better?

markjcrane:
Thanks

PR #1978, opened 2016-10-10
[4.2] allow page-break on fax_message & fax_footer

hargut:
Prevents loss of message text when the message is longer than a single page.
Footer is placed afterwards, and has page break enabled.


hargut:
Just noticed that there is still a little flaw it it. Will update soon.

hargut:
With commit 3ca82f7 single page & multi-page works properly.

hargut:
Can you please cherry-pick it into master as well?

PR #1994, opened 2016-10-14
Add the capability to have extension and DID with same number

daniel-lucio:
Usually, there is a dial plan clash when you have a DID and an extension the same number (10+ digits longer). Adding this condition will force user verification, as it is supposed when you use an outgoing route, the user is not local.

Some internal (999) extension work must be done, I will post that shortly for 4.0. It seems 4.2 doesn't need this as it checks for user existance regarding extension name.

### DIFF
--- a/app/dialplan_outbound/dialplan_outbound_add.php
+++ b/app/dialplan_outbound/dialplan_outbound_add.php
@@ -378,6 +378,13 @@ else {
 						unset($sql);
 
 					$dialplan_detail_tag = 'condition'; //condition, action, antiaction
+					$dialplan_detail_type = '${user_exists}';
+					$dialplan_detail_data = 'false';
+					$dialplan_detail_order = '004';
+					$dialplan_detail_group = '0';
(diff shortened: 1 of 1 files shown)


markjcrane:
You are trying to add a features to 4.0 which is an old release. Its generally considered bad practice to add features to releases. This feature exists on release 4.2. One could easily argue the right course of action if you want this is to upgrade to 4.2.

markjcrane:
After creating the outbound route the condition user_exists can be added manually by a user in the superadmin group. So no one is being deprived in anyway other than a little inconvenience they get for an administrator that will not upgrade customers to the current release.

daniel-lucio:
Though we were trying to make this kind of tricky things as easy as possible for newbies. Not supporting this out of the box damages the image of the project.

markjcrane:
I've told you the policy no features added to already released versions several times for multiple years. It's a dangerous trend to do as adding features can destabilize releases. It also looks unprofessional as almost no project adds features to older versions that were already released.

PR #1995, opened 2016-10-14
Add the capability to have extension and DID with same number

daniel-lucio:
Same rationale as 4.0


markjcrane:
Generally considered bad practice to add features to releases. Release get bug fixes and security fixes.

daniel-lucio:
As I said before, the aim was making life easier for newbies.

markjcrane:
Unprofessional to add features to old releases. I can't think of any software that does that. And this helps encou

PR #1996, opened 2016-10-14
Allow longer extensions

daniel-lucio:
As per my preview pull request, this enables longer extensions.


markjcrane:
Changing behavior of an old release is frowned upon. 4.2 release addresses this and allows extensions of any size by using a condition of user_exists = true rather than checking the length of the destination number.

daniel-lucio:
This is for 4.0, not for 4.2, it was a complementary fix. That's why I didn't push this for 4.2 as I know it does not need it.

PR #2005, opened 2016-10-19
Add German indications

jalr:
Values obtained from [voip-info.org](http://www.voip-info.org/wiki/view/Asterisk+indications+Germany)
Hebrew translation is from Google Translate and therefore possibly wrong


markjcrane:

jalr:
I've signed up via pull-request and sent an email

PR #2011, opened 2016-10-20
faxing smtp_port fix + smtp certificate override

jebsolutions:
smtp_ignore_bad_certificate is added to default settings.  Default is false.

If set to true then will add flags for PHPMailer to ignore certificate checks.  This lets you use self-signed certificates.

Note: you have a similar override as an option for imap on the fax server advanced options.  This patch does something the equivalent on the SMTP side for sending fax and voicemail to email.

This patch lets you use self-signed certs on recent php/phpmailer combinations.  

This patch lets fax-to-email work with non-standard smtp ports by adding similar logic from voicemail-to-email (v_mailto.php).

### DIFF
--- a/core/default_settings/app_defaults.php
+++ b/core/default_settings/app_defaults.php
@@ -170,11 +170,11 @@ if ($domains_processed == 1) {
 		$array[$x]['default_setting_description'] = '';
 		$x++;
 		$array[$x]['default_setting_category'] = 'email';
-		$array[$x]['default_setting_subcategory'] = 'smtp_ignore_bad_certificate';
+		$array[$x]['default_setting_subcategory'] = 'smtp_validate_certificate';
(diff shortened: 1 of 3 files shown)


markjcrane:
I don't like smtp_ignore_bad_certificate as a self signed certificate is a valid certificate.  smtp_validate_certificate or smtp_allow_self_signed maybe a better choice for the name.
Please update this and fix the conflicts or submit a new pull request.

jebsolutions:
@markjcrane,

Thanks for the feedback.  I will change to use smtp_validate_certificate.

Will look at the conflicts.

markjcrane:
branch still conflicting.

jebsolutions:
I am closing this PR since it's obsoleted by my other PRs.

Issue #1987, opened 2016-10-13
Fax Problems


GIELUb:
1. Fax Log (fax_log_view.php) Back-Button to small
2. Fax Sent (fax_files.php) sent-faxes not listed
   This steps will reproduce the problem.
   (1) Send 3 faxes through the web interface.
   (2) Check the Log - the faxes was successfully sent.
   (3) Look in the Sent folder - no fax listed.
3. Fax Logs (fax_logs.php) Date - no Time

Version 4.2.0
Git Information Branch: 4.2
Commit: 65ba2caa57fcb517a9fce07ff85650e16fefa93e
Switch Version 1.6.12 (64bit)
Operating System: Linux fusion-tk 3.16.0-4-amd64 #1 SMP Debian 3.16.36-1+deb8u1 (2016-09-03) x86_64 GNU/Linux

jebsolutions:
part 1: back button fix
https://github.com/jebsolutions/fusionpbx/commit/c9fd79f14ccd516104a9a991eacfc22c5654115b

jebsolutions:
Part 2: sent faxes not listed.  

Cause is lua error in resources/install/scripts/app/fax/resources/scripts/queue/retry.lua
freeswitch logs show:

1972    2016-10-31 17:13:24.940549 [ERR] mod_lua.cpp:203 /usr/share/freeswitch/scripts/fax_retry.lua:289: attempt to index field 'keep_local' (a nil value)
1973    stack traceback:
1974    /usr/share/freeswitch/scripts/fax_retry.lua:289: in main chunk

In my default settings and ["fax"]["keep_local"] fax was not defined.  As soon as I created this boolean the error went away and the files showed up in my sent folder.

I see that this variable should be set by:
/var/www/fusionpbx/app/fax/app_defaults.php 

I'm pretty sure I ran an app default already, and all the other settings here look like they are set correctly.  I'm not sure why this one wasn't set?

Work around: manually create variable.

jebsolutions:
part 3: log only shows date, not time.

v_fax_logs fax_date is defined as date.  it should be "timestamp without time zone" (this is how the other fax date fields are set).

Must alter database to fix:

e.g. Postgres:
alter table v_fax_logs alter column fax_date type timestamp without time zone;

Existing faxes will have 00:00:00 for the timestamp, but new faxes after this will correctly show the time.

jebsolutions:
Fixes for all of this is done.

[...]

PR #2025, opened 2016-11-01
bug: #1987 part 2. Fax log should be timestamp not date re

jebsolutions:
Fix for field being "date" instead of "timestamp" in log file.

Resolves bug.

Note: Advanced->Upgrade->Schema->"data types - Detects and updates incorrect field data types" doesn't change this field type.

work around for existing installs: manually change field from date to timestamp for postgres/mysql.


markjcrane:
This just looks like the previous pull requests that I already committed. Please cleanup your pull request so it shows only current changes. Then resubmit the changes.

Issue #1821, opened 2016-08-17
When copying extensions does not work limit on the number of extensions (the default limit or limit in the domain)


ash144:
Hi there. 
Sorry for my English..

FusionPBX 4.1.0
Default settings:
 Сategory = limit
 Subcategory = extensions
 Type = numeric
 Value = 10
 Enabled = True 

Domain test.pbx.org settings:
 Сategory = limit
 Subcategory = extensions
 Type = numeric
 Value = 3
 Enabled = True 

In the domain test.pbx.org created 3 extensions: 1000, 1001, 1002.
I edit 1002 and click "Copy" button. This creates another еxtention, and the limit is exceeded.
Or, open https://my.pbx.org/app/extensions/extension_copy.php. This creates a "blank" extension, edit it, and limit is exceeded.

Another problem: when you copy, you can add the existing name of extensions. As a result, you will create two identical extensions. I do not think it is correct.
Thank you.

jebsolutions:
@ash144 Thanks for reporting this.  I can duplicate that cloning can create duplicate extensions.

Pull Request #2033  
Pull request #2034 

This only fixes the copy creating duplicate extensions.

I think the limit part is a misunderstanding of what the limit does.  I admit the first time I saw this screen it wasn't clear to me either.

As I understand it, Limit basically says "When you click save create X number of extensions".  E.g. if you set limit to 3 and extension was 1000 it would create extension 1000, 1001, 1002 (i.e. 3 extensions) automatically.  

Also, clone always clones only one extension.  It has nothing to do with the limit.

@ash144 if you meant something else and there's another bug please clarify.

jebsolutions:
test case:
Go to Accounts->Extensions
If you don't already have an extension, then create one (e.g. 1001)

Edit extension 1001
Click Copy button at the top
at the Enter the New Extension prompt type in "1001" and click ok

expected result: rejects duplicate extension 1001
actual results: you now have two extension 1001's

PR #2033, opened 2016-11-03
prevent cloning duplicate extensions (master #1821)

jebsolutions:
app/extensions/app_languages.php
+ add duplicate extension error prompt

app/extensions/extension_copy.php
+ add check to see if duplicate extension exists

app/extensions/resources/classes/extension.php
+ add function to check if extension exists

### DIFF
--- a/app/extensions/app_languages.php
+++ b/app/extensions/app_languages.php
@@ -1598,17 +1598,4 @@ $text['button-call_routing']['ro'] = "rutare de apel";
 $text['button-call_routing']['ar-eg'] = "توجيه الدعوة";
 $text['button-call_routing']['he'] = "ניתוב שיחות";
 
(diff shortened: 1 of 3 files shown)


markjcrane:
In master I have added method exists to the extension class.
https://github.com/fusionpbx/fusionpbx/blob/master/app/extensions/resources/classes/extension.php

In 4.2 branch there is a function called extension_exists that is located in switch.php and accessible by default. It can be used to keep changes to a minimal.

As far as the message we have used this for the duplicate-message.
https://github.com/fusionpbx/fusionpbx/search?utf8=%E2%9C%93&q=duplicate+detected

Please make these changes and then it will be more consistent and can be committed.

jebsolutions:
@markjcrane Thank you for the feedback.

I will make the changes you suggested.

PR #2034, opened 2016-11-03
fixes for #1821 4.2 branch

jebsolutions:
app/extensions/app_languages.php
+ add duplicate extension error prompt

app/extensions/extension_copy.php
+ add check to see if duplicate extension exists

app/extensions/resources/classes/extension.php
+ add function to check if extension exists
+ all cherry picks "AND AND" fix from master


markjcrane:
See comment on https://github.com/fusionpbx/fusionpbx/pull/2033

PR #2041, opened 2016-11-07
Wizard App

konradSC:
This app will enable adding users, extensions, devices, contacts and voicemails all at one time and and from one place. Users will be linked to devices, extensions and groups. Contacts are linked to users. Devices are linked to extensions. Also included are "Wizard Templates". These templates allow a super admin to a common set of attributes and apply these during the process of adding a user. For instance, emergency caller-id can be set in a template and this never needs to be entered again when adding a user  through the Wizard. I plan on expanding the cabilities of Templates by adding exta fields for things like "Limit Max". 

Another major feature is the ability to add multiples through a CSV import. Use the CSV headers specified in the App (Username, Extension, FirstName, LastName, Email, MAC, WizardTemplate, DeviceTemplate, DeviceProfile, Password, PIN). This will allow superadmins bulk import large amounts of users into the system. The import also utilizes Wizard Templates as well for common settings. 

Things to add in the future:
-Make the app fully multi-lingual. The language file is lacking in a few fields.
-Add a "copy" function to the Templates.
-Expand the options that can be set in Templates
-Possibly add a way to make sure the following fusionpbx apps are installed. (Extensions, Voicemail, Devices, Users, & Contacts)

Feedback is welcome! This is my first ever attempt at anything PHP, so be kind :-) Thanks!!


markjcrane:
Will review as time permits to see if it matches coding standards. For example you won't see words run together in fusionpbx unless it a native php function. Underscore is used between words and lower case. I beleive in being consistency is important to me. Makes it easier to code when it's consistent.

konradSC:
My logic behind running words together in the CSV header was simply that it is sometimes difficult to read underscores in spreadsheet editors like Excel. Example: "_FirstName_" instead of "_First_Name_".  It's fine to adjust though. I think that is the most glaring place that I strayed from the "fusion" standards.

markjcrane:
In the interface shown to the user underscores are replaced by spaces.

konradSC:
I updated the CSV headers to fall inline the with fusion naming conventions. It looks some like this when you look at it in the web interface:

**Wizard Import**
_Import extensions from a CSV file. Use the following headers:

username (Username),
extension (Extension),
first_name (First Name),
last_name (Last Name),
email (Email Address),
mac_address (MAC Address),
wizard_template (Wizard Template),
device_template (Device Template),
device_profile (Device Profile),
user_password (User Password),
vm_password (Voicemail Password)_

PR #2043, opened 2016-11-07
Update sip.trust_ctrl for yealink firmware version 80

blackc2004:
In version 80 of the yealink firmware the trust_ctl changed from
account.X.sip_trust_ctrl=1  to sip.trust_ctrl=1. Making this change in
the provisioning files.


blackc2004:
http://support.yealink.com/faq/faqInfo?id=181

PR #1629, opened 2016-06-06
Add. attr_xfer analog based on conference.

moteus:
Control DTMF sequence
 `*0` transfer `self` to `enter number` state and `peer` leg to conference room
 `##` transfer `self` to `enter number` state and hangup `peer` leg
 `*#` transfer `self` to conference room and hangup `peer` leg

PR #2077, opened 2016-11-14
Fix. Add missing file for #1629

moteus:


markjcrane:
**File location**

No scripts should really be placed in the root of the scripts directory. All of the current ones will be eventually moved into a directory to organize the files better. For a reference look at the directories in /var/www/fusionpbx you will see directory structure is more organized for PHP than LUA. 

If its a general function it should go into resources/install/scripts/resources/functions/transfer2.lua. 

**File name**
Maybe a better name for it is channel_transfer.lua as the 2 isn't an intuitive name .

moteus:
I agree about orginization of Lua directory. But I think that some script can be put in root if they provide some general functionality.
This script in fact allows to transfer bridged channels to 2 separate destinations and in fact have no 
relation to any application in Fusion at all. Also it is not `resource`. It just simple util to be used in dialplan.
And because of that some one may be used it in its own dialplan and it will be convinient do not require enter long path to script.

Also I agree that name is not ideal. But not sure that `channel_transfer.lua` is better.
May be `channels_transfer.lua`?

markjcrane:
General functions that we use as a library should go in scripts/resources/functions.

As far as the filename goes are you transferring one call at a time or multiple calls at the same time? My guess is you are transferring one call which make me like channel_transfer.lua.

moteus:
> General functions that we use as a library should go in scripts/resources/functions.

This is not function in terms of Lua. But it more like application in terms of FS XML dialplan.
So not sure `scripts/resources/functions` is right place for it

> As far as the filename goes are you transferring one call at a time or multiple calls at the same time? My guess is you are transferring one call which make me like channel_transfer.lua.

It works to both A-leg and B-leg to 2 different destination. 
```
conf-xfer,*0,api:lua,transfer2.lua ${uuid} conf_enter_number::XML::conf-xfer@${domain_name} conf_enter_to::XML::conf-xfer@${domain_name}
```
In this example we send one channel to `conf_enter_number` and other to `conf_enter_to`

PR #2102, opened 2016-11-20
Fix for empty dial plan on upgrade

blackc2004:


markjcrane:
I'm just going to remove this line as I don't think it really needed.
p.dialplan_xml as text = ''

PR #2105, opened 2016-11-20
[master] include stderr in git pull results and check for error 

jebsolutions:
Old result (seems correct but actually failed):
Updating d190f6a..d958bea

Corrected result (includes failure message):
Updating d190f6a..d958bea
error: Your local changes to the following files would be overwritten by merge:
app/fax/fax_log_view.php
Please, commit your changes or stash them before you can merge.

### DIFF
--- a/core/upgrade/index.php
+++ b/core/upgrade/index.php
@@ -65,7 +65,7 @@
 						$update_failed = false;
 					}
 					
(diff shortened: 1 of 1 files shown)


markjcrane:
Good job this is a great improvement.

PR #2114, opened 2016-11-21
Add. Use params in fax_retry.lua

moteus:


moteus:
In this case we have parameter only if value not `nil`. So result parameters table will have them in same way.
But e.g. `fax_date` calculate always. I plan handle named parameters by hand in Database class and convert them to positional. (Named params support not all databases/drivers). Now it ignores any parameters which not present in query.
Do you have any arguments to do such test?

PR #2120, opened 2016-11-21
Add. Use params in send_mail.lua

moteus:


jebsolutions:
p.s. you are awesome for fixing these sql vulnerabilities so quickly.  Thank you!

PR #2126, opened 2016-11-21
Add export to Pin Numbers application

andycol:
add the ability to export to csv


markjcrane:
Why I haven't committed this change was primarily this.
 app/call_flows/call_flows.php Removing line 122 breaks the HTML.

I will accept the change and fix it for you.

markjcrane:
I've fixed it. The committed line number is different.

Issue #1973, opened 2016-10-07
Feature Request: Add Device Settings to Device Profile


konradSC:
Under Device Profiles we are able to add common Keys for devices. It would be very helpful to be able to add common Settings in the Profiles as well like we can enter under Devices. (Name,Value,Enabled,Description)
[settings.pdf](https://github.com/fusionpbx/fusionpbx/files/516775/settings.pdf)

markjcrane:
This feature is the to do list.

jlcrews:
I would vote for this as well.

konradSC:
Added

PR #2165, opened 2016-11-23
Device profile settings

konradSC:
Adds device settings to device profiles. This is in response to [Issue#1973](https://github.com/fusionpbx/fusionpbx/issues/1973). 

The precedence for settings are system, domain, profile, device.


markjcrane:
Thanks for completing something that was on the to do list. Looking at the code it looks like it was done well.

konradSC:
Thanks. I pretty much stole all the code from other parts of Fusion for this one :-)

markjcrane:
Totally fine with me the end result is a good feature that will be very
useful.

PR #2166, opened 2016-11-23
user_edit field preservation + eliminate variable not found warnings

jebsolutions:
This fixes some of the "I clicked save and my changes were lost" bug reports.
Also fixes many of Undefined index/Undefined variable warnings that are hidden due to php error_settings blocking warn/notify.

steps to replicate initial problem:
click + to add a new user.  
Add SOME but not all of the fields (e.g. leave out group).  
Click Save.

expected result: it warns you the field is missing, but other fields are preserved

actual result: fields are reset

this fix: fields are preserved across save attempt.

### DIFF
--- a/core/users/user_edit.php
+++ b/core/users/user_edit.php
@@ -36,13 +36,13 @@
 
 //get user uuid
 	if (
-		(is_uuid($_REQUEST["id"]) && permission_exists('user_edit')) ||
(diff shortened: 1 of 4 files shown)


markjcrane:
I've considered the changes you are recommending here and they definitely are good. I've been hoping to find an even more elegant more automatic way to accomplish the same thing. However if that is not found then may accept your changes.

markjcrane:
Still will look for a more elegant way as time permits.

PR #2201, opened 2016-11-27
add iNum Dial pattern to dialplan drop-down

jebsolutions:
+011-883-5100-xxxx-xxxx

Most voip providers will give you a free iNum number.  Calls between iNum numbers are typically free.


markjcrane:
This should probably say 'North America iNum' since the +011 it will only work in North America.

PR #2202, opened 2016-11-28
Lua escape fixes + adds transcription checks + voicemail del

jebsolutions:
@blackc2004 also avoids errors when transcription has sound, but no speech.

### DIFF
--- a/app/dialplan_outbound/app_languages.php
+++ b/app/dialplan_outbound/app_languages.php
@@ -258,6 +258,13 @@ $text['label-800']['fr-fr'] = "Sans Charge";
 $text['label-800']['pl'] = "Połaczenia specjalne (toll-free)";
 $text['label-800']['uk'] = "Безкоштовний";
 
+$text['label-inum']['en-us'] = "iNum 0118335100xxxxxxxx";
+$text['label-inum']['es-cl'] = "iNum 0118335100xxxxxxxx";
+$text['label-inum']['pt-pt'] = "iNum 0118335100xxxxxxxx";
+$text['label-inum']['fr-fr'] = "iNum 0118335100xxxxxxxx";
+$text['label-inum']['pl'] = "iNum 0118335100xxxxxxxx";
+$text['label-inum']['uk'] = "iNum 0118335100xxxxxxxx";
+
 $text['label-711']['en-us'] = "711 TTY";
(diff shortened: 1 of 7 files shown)


moteus:
1. You do not need add spaces as last symbol in SQL qyer. e,g. 
```Lua
local sql = [[SELECT * FROM v_domains WHERE domain_name = :domain_name]];
....
dbh:query(sql, params, function(row)
```
There no need space after `:domain_name`.

2. This is yuor notes from #2203 

>Note ... the number of variables in the SQL must match the number of variables in the parameter array.

This is true but it is not related to Fusion PBX `Database` class.

>Note ... if you need to access the same variable twice within a sql statement, you need to use two different variable names. You can't use the same parameter name twice.

Not sure where you find this. None of the supported by FS database drivers (SQLite, ODBC, PgSQL, MySQL) does not support named parameters at all. 
So we should write some code to convert named parameters to positional. e.g.
```Lua
local message_count_by_uuid_sql = [[SELECT
  ( SELECT count(*)
    FROM v_voicemail_messages
    WHERE voicemail_uuid = :voicemail_uuid
    AND (message_status is null or message_status = '')
  ) as new_messages,
  
  ( SELECT count(*)
    FROM v_voicemail_messages
    WHERE voicemail_uuid = :voicemail_uuid
    AND message_status = 'saved'
  ) as saved_messages
  ]]
  local params = {voicemail_uuid = voicemail_uuid};
```
will be converted to
```SQL
SELECT
  ( SELECT count(*)
    FROM v_voicemail_messages
    WHERE voicemail_uuid = ?
    AND (message_status is null or message_status = '')
  ) as new_messages,
  
  ( SELECT count(*)
    FROM v_voicemail_messages
    WHERE voicemail_uuid = ?
    AND message_status = 'saved'
  ) as saved_messages
```
```Lua
params = {voicemail_uuid, voicemail_uuid}
```
But now Database class does not support real params and just do string escape + string replace.
So there also no need unique names as well.

So I suggust do not accept this PR

moteus:
I am not maintainer. But I think it better make PR wich fix one specific problem.
And combine in single PR totally unrelated fixes is not good idea.

jebsolutions:
I agree about putting it into separate pr's.

markjcrane:
I agree on separate pull requests as well.

PR #2210, opened 2016-11-29
Debian pkgs

kb3vgw:
updated with new template


markjcrane:
The package would need to be added to its own repo. Not likely to put the debian packaging files inside the web directories. Seems like it would be messy when supporting multiple operating systems. Also seems like files that don't belong in the website directory. I would like it a lot more outside of the website directory for the packaging files.

PR #2218, opened 2016-12-01
Fix. Improve parameters patterns.

moteus:
E.g. now it handle queries like
```Lua
local sql = "select :J::text as value"
print(dbh:first_value(sql, {J='hello'}))
```

### DIFF
--- a/resources/install/scripts/resources/functions/database.lua
+++ b/resources/install/scripts/resources/functions/database.lua
@@ -23,7 +23,7 @@ local unpack = unpack or table.unpack
 
 local NULL, DEFAULT = {}, {}
 
-local param_pattern = "[:]([^%d%s][%a%d_]+)"
+local param_pattern = "%f[%a%d:][:]([%a][%a%d_]*)"
 
 --
 -- Substitude named parameters to query


moteus:
Current parameter pattern
 * does not support single letter parameters name (like `:A`)
 * treats strings like `HH:MM:SS` as parameters `:MM` and `:SS`.
 * does not allows use pgsql type cast. In my example it is pgsql typecast parameter `:J` to `text` type.
  But now it convert to `:J:$1` with parameter `:text` which is totally wrong.

All this cases does not exists in current codebase. So it safe to replace.

markjcrane:
When freeswitch adds real parameterized queries then I don't see the need for this extra sytnax will.

> local sql = "select :J::text as value"
> print(dbh:first_value(sql, {J='hello'}))

Please explain why its needed as they are making movement to supporting real parameterized queries.

moteus:
FS will add native parameters support. That means it will supports only positional parameters.
So we still need convert named parameters to parameter placeholder in queries.
E.g. we have query like `select * from v_users where user_name=:name`.
When we emulate parameters we convert it to `select * from v_users where user_name='some_user_name'`
When we use real parameters we convert it to `select * from v_users where user_name=?` and pass `some_user_name` as parameter.
I use named parameters because it easy to use when query has many parameters. When you change query you do not need bother to change order in parameter list.
And it also easier to find parameters in query text to be able emulate parameterized queries.

markjcrane:
I understand the purpose and parameterized queries and this is what we have right now.
select * from v_users where user_name=:name

The real question was to do with the pull request where you added something like this.
select * from v_users where user_name=:name::text

name::text and I'm questioning whether that is really needed when using real parameterized queries.

PR #2225, opened 2016-12-02
Increase Max Device lines

andycol:
increase lines from 12 to 255 to accomodate multiple BLF's


markjcrane:
We do allow 255 keys. However this change is referring to the device key line number its unlikely there are 255 registrations which is why the number is currently set to 12 as in 12 registrations per phone.

PR #2234, opened 2016-12-06
Support for E-164 format \ Allow sendmail() to be called more than once \ Bug fixes

SJS87:
The preg_replace functions are updated to allow '+' to be used as the prefix (first character only) for phone numbers, other non-digit characters are removed as before.


SJS87:
Declare the mime_parser_class if it doesn't exist

Calling sendmail() more than once causes "PHP Fatal error:  Cannot redeclare class mime_parser_class".

This resolves the problem by only declaring the class if it doesn't exist.

SJS87:
Declare the rfc822_addresses_class if it doesn't exist

Calling sendmail() more than once causes "PHP Fatal error:  Cannot redeclare class rfc822_addresses_class".

This resolves the problem by only declaring the class if it doesn't exist.

PR #2235, opened 2016-12-06
Remove unused database connection.

moteus:


daniel-lucio:
Can you backport this to 4.2, this will eventually hit DB performance by keeping connections open until it expires, which it affects availability

moteus:
>this will eventually hit DB performance by keeping connections open until it expires, which it affects availability

not sure what you mean. I read your post about that FS does not free database connections. But as I can see FS correctly destroy all objects when Lua VM closed.  And as I can see FS create new Lua VM for each 
Lua request.  So all connections should be free correctly in any case.

markjcrane:
daniel-lucio 4.2 doesn't currently use the code that replaces that connection so you can't remove the database connnection without a replacement. In master branch there was a replacement.

moteus:
markjcrane. This obhect does not use by script at all. It just cteats and thats it.

Also I think i found memory/resource leak in mod_lua. It happens only for xml handler when
script raises error. Can test/fix tomorrow. If you want to test can try make **global** dbh object 
in xml handler script and raise error.

PR #2249, opened 2016-12-13
Add. Speed dial respects contacts user.

moteus:
One user can not use speed dial numbers from contacts
associated with another user


markjcrane:
Is the purpose of this change to restrict the caller so they can only use speed dial for contacts assigned to them by group or user? Explain your change better to get it accepted.

moteus:
Yes. If user assign to contact then only extension assigned to this user can use it.
But if contact has no users assigned to it then any extension can use it.

PR #2256, opened 2016-12-19
Fix IVR when using custom recordings

daniel-lucio:
if the filename is not an absolute path, freeswitch will look the file in the sounddir directory. This makes sure the file gets the absolute path.

### DIFF
--- a/app/ivr_menus/ivr_menu_edit.php
+++ b/app/ivr_menus/ivr_menu_edit.php
@@ -476,7 +476,7 @@
 			echo "<optgroup label='Recordings'>\n";
 			foreach ($recordings as &$row) {
 				$recording_name = $row["recording_name"];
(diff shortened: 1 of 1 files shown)


markjcrane:
Relative paths were used by design and are working.

daniel-lucio:
Thats my big concern, the lua script was not able to find the recording using relative path. This means there is a condition that it is failing (i bet not always), what file or piece of code does that decision? I will take a look and find if there is a wrong condition there.

markjcrane:
Well you failed to mention details like version of fusionpbx and xml handler or ivr_menu.lua.

daniel-lucio:
4.2.1, it was an installation not done by me. After looking, it seems it was using xml handler, I didnt find any ivr_menu.lua reference.

PR #2257, opened 2016-12-19
Fix IVR when using custom recordings

daniel-lucio:
Same rationale as 4.2

### DIFF
--- a/app/ivr_menus/ivr_menu_edit.php
+++ b/app/ivr_menus/ivr_menu_edit.php
@@ -555,7 +555,7 @@
 			echo "<optgroup label='Recordings'>\n";
 			foreach ($recordings as &$row) {
 				$recording_name = $row["recording_name"];
-				$recording_filename = $row["recording_filename"];
+				$recording_filename = $_SESSION['switch']['recordings']['dir']."/".$_SESSION['domain_name']."/".$row["recording_filename"];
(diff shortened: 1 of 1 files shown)


markjcrane:
It uses relative path directories on purpose this would hard code the paths.

daniel-lucio:
Lets keep conversation in this post.

I know it is a relative path, but there are 2 issues that are not meeting here.
if relative path, sound is looked in the sounds_dir directory, aka /usr/share/freeswitch/sounds/. Custom recordings are not there.

custom recordings (the ones that are available at app->recording) path should be absolute as usually it is  recording_dir/domain/something, for example /var/lib/freeswitch/recordings/mydomain/file.wav

The patch only uses absolute paths when using custom recordings, not in another place.

markjcrane:
We look in sounds and custom recordings for the relative path.

daniel-lucio:
Where is that part the code that looks when relative path, I couldn't find not in the PHP not in the LUA. This is the reason of this patch.

markjcrane:
4.2 uses xml handler by default but also the option for to use ivr_menu.lua. Both of the files have relative path searching to fix the set the full path. You should probably explain state which one you are using.

markjcrane:
Relative path is more flexible allow the base path to be changed without having update the path on all of the IVR Menus.

PR #2260, opened 2016-12-20
Fix. Use CID from extension for call forward if set `call_forward_fix`.

moteus:
This is same behavior as in Web UI.

### DIFF
--- a/resources/install/scripts/call_forward.lua
+++ b/resources/install/scripts/call_forward.lua
@@ -82,6 +82,8 @@
 --connect to the database
 	local dbh = Database.new('system');
 
+	local settings = Settings.new(dbh, domain_name, domain_uuid);
+
 --request id is true
 	if (request_id == "true") then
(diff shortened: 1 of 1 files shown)


markjcrane:
The variable call_forward_fix is a not a accurate name it implies caller id is broken by default. By default it uses the caller's original caller id which makes more logical sense so that the person being called knows who is calling them.

The call_forward_fix overrides or replaces the original caller id with the extension's outbound caller id and you no longer know who is calling you. This is required for some carriers and maybe governments that  didn't think about the fact that there is sometimes legitimate reasons like this to continue to use the same caller id from the original caller.

It is okay to have the setting to change the behavior of the caller id but it must have a better name for the variable.

Ideas for a variable name:
call_forward_caller_id
caller_id_source

Value could be true or false... or it could be outbound, original or caller.
outbound (extension outbound caller id), original (or caller)
Useful if this were extended for follow me.

PR #2275, opened 2016-12-27
show shell exec errors.  Nicer hide/show.

jebsolutions:
test #1:
enter an invalid command for the shell.  e.g. "asdf" then click execute.

expected result:
You get some type of result about invalid command.

Actual result: nothing shows up.  You have no idea if your command worked or not.

fixed result: When doing an exec include errors in the response: 2>&1

Test #2:
Enter something invalid in the switch command.  e.g. "asdf".  Click Execute.  Change to SQL.  Enter something valid e.g. "select * from v_users;"  Click execute.

Expected result:
You see the SQL output right below your "select * from v_users;" input text box.

Actual result:
Because the output from the previous switch/php/shell is not hidden the SQL output is pushed down off the screen.  Unless the user scrolls down they will not see the result and think their SQL command didn't actually work.

Fixed results:
When you do switch/phps/shell show the Response section.
When you do sql hide the response section.

This way only the relevant response/sql_response is shown to avoid confusion/screen clutter.

### DIFF
--- a/app/exec/exec.php
+++ b/app/exec/exec.php
@@ -128,6 +128,7 @@
 							$('#iframe').prop('src','');
 							$('#sql_response').hide();
 						<?php } ?>
+						$('#response').show();
 						break;
 				<?php } ?>
 				<?php if (permission_exists('exec_php')) { ?>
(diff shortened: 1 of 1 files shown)


markjcrane:
Looks great except this might break win does support.

jebsolutions:
@markjcrane thank you for the feedback,

Which part?  If you mean the " 2>&1" part it also works under windows.  

If I am missing something please let me know and I will re-test?

markjcrane:
Yes that was the part I thought was questionable.

Issue #2270, opened 2016-12-22
Fax reports email 


volga629:
Emails status from sent fax 
 is should show up under status ---> Emails ?
 Retained in v_emails as email_uuid = b080deff-a48c-4761-bfcf-23794a7f74c7
 email tables empty 
 look like a insert failing 
 here are failing insert log 
 
```
Dec 22 17:28:32 capgp01 pgpool[11258]: [12058-1] 2016-12-22 17:28:32: pid 11258:    # printf-style string to output at beginning of each log line.LOG:  pool_send_and_wait: Error or notice message from backend: : DB node id: 0 backend pid: 20422 statement: "insert into v_emails ( email_uuid, domain_uuid, sent_date, type, status, email ) values ( 'b080deff-a48c-4761-bfcf-23794a7f74c7', '', now(),'', 'failed', 'MIME-Version: 1.0

```

jebsolutions:
@volga629 if possible, can you post the full error log.  i.e. what comes after 'MIME-Version: 1.0

Based on line 359 I think I know what the problem is:
https://github.com/fusionpbx/fusionpbx/blob/130147990ec6bb8d8a665439ba5a310751ec6830/secure/v_mailto.php

If I can see the full SQL statement it tried to run I will be able to confirm.

volga629:
Here are what I found.


```
Dec 22 17:28:32 capgp01 pgpool[11258]: [12058-1] 2016-12-22 17:28:32: pid 11258:    # printf-style string to output at beginning of each log line.LOG:  pool_send_and_wait: Error or notice message from backend: : DB node id: 0 backend pid: 20422 statement: "insert into v_emails ( email_uuid, domain_uuid, sent_date, type, status, email ) values ( 'b080deff-a48c-4761-bfcf-23794a7f74c7', '', now(),'', 'failed', 'MIME-Version: 1.0
Dec 22 17:28:32 capgp01 pgpool[11258]: [12058-2] #011Content-Type: multipart/mixed; boundary="XXXX_boundary_XXXX"
Dec 22 17:28:32 capgp01 pgpool[11258]: [12058-3] #011To: info@myorg.org
Dec 22 17:28:32 capgp01 pgpool[11258]: [12058-4] #011From: info@myorg.support
Dec 22 17:28:32 capgp01 pgpool[11258]: [12058-5] #011Subject: Fax to: 5064516130 FAILED
[...]

PR #2276, opened 2016-12-28
fixes bug #2270 reported by volga629

jebsolutions:
Please do further testing/review and give comments and suggest cleanup, etc.

This bug is only tripped if all of the following are true:
a) fax send fails
b) email notification of fax failure also fails
c) Other?

how to test
- have a 100% working fax to email setup.
- Change Advanced->defaults->email and set the port to be something invalid.
- e.g. smtp port 9999 is invalid
- send a test fax to something without a fax
- e.g. I just faxed my own DID which does NOT have fax answering

- each time the phone rings answer...wait 5 seconds, and then hang up
- note: it will retry sending the fax 5 times so repeat the above.

- on the last try it will give up, and try to send the failure email
- the email will fail (bad port) and then it will insert the failed email into v_emails
- the insert will fail and nothing in v_emails

Error was caused by trying to insert '' into the domain_uuid field.
'' is invalid in a uuid field, but null is okay.

The null is caused by no x-headers being passed.  So these two fields end up null:
$headers["X-FusionPBX-Domain-UUID"]
$headers["X-FusionPBX-Email-Type"]

Fix: feed the X-Header in the calling lua code.

### DIFF
--- a/resources/install/scripts/fax_retry.lua
+++ b/resources/install/scripts/fax_retry.lua
@@ -416,6 +416,11 @@
 	freeswitch.consoleLog("INFO","[FAX] mailfrom_address: ".. from_address .."\n");
 	freeswitch.consoleLog("INFO","[FAX] mailto_address: ".. email_address .."\n");
 	freeswitch.consoleLog("INFO","[FAX] hangup_cause_q850: '" .. hangup_cause_q850 .. "'\n");
+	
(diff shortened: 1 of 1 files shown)


jebsolutions:
note: for testing when setting smtp_port in defaults also make sure to make this setting enabled.

PR #2297, opened 2017-01-13
Fix. Display Lan IP for Grandstream

moteus:
Grandstream send IP like `BJC.BGI.B.CAB` so we
should convert it to numeric form.


ethanmullins:
Awesome. We've had this issue for quite awhile

bigfishvoice:
Anyway we can get this applied as a bug fix to 4.2.1 as well?

PR #2310, opened 2017-01-17
Add contact column to registration table

sergey-mz:
Command "sofia xmlstatus profile internal reg" returns contact in this format:
"Contact Name" <sip:236@10.0.15.108:5060;>
Only Aastra phones send Contact name with REGISTER message. 
I tested Grandstream and Polycom phones, but contact name is missing. 
Adding just "	<td class='".$row_style[$c]."'>".$row['contact']."</td>\n"; will produce same visible result because <sip:....> portion will be hidden.


markjcrane:
I finally got a chance to test this on my system it looks good. Accepting your change.

f1lint:
This add leads to a fail with 500 code...

sergey-mz:
Did you apply this change too:
https://github.com/fusionpbx/fusionpbx/pull/2311

PR #2325, opened 2017-01-24
Fix a bug that breaks load balancing

daniel-lucio:
Some Freeswitch deployments will create an empty core.db file (no tables inside) even when the core-db-dsn and core-recovery-db-dsn options are set. 

There is no reason to try looking for the core.db file when using balancing. The existence of an empty core.db file breaks the balancing.

This patch is totally safe, as line 141 only executes this code if balancing is on.

### DIFF
--- a/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
+++ b/resources/install/scripts/app/xml_handler/resources/scripts/directory/directory.lua
@@ -157,13 +157,8 @@
 								require "resources.functions.file_exists";
 
 							--connect to the switch database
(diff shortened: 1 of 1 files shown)


daniel-lucio:
A side note, I see you fixed this already in 4.2 and 4.3.

markjcrane:
Committing this but anyone using 4.0 still should reconsider as there is a major security risk in the lua code in 4.0. Fixed in 4.2.1 requires a considerable time to back port the changes so they will not be back ported. To fix you need to upgrade to 4.2.1.

PR #2330, opened 2017-01-28
Add variable to disable autoprovisioning password

emaktech:
Adds a variable http_auth_disable where when set true disables password authentication. If disabled or any other value is set, password authentication remains enabled.

Disabling password check currently necessary to set new autoprovisioning password on devices via autoprovisioning.

Better alternative may be to enable two passwords at the same time temporarily until all devices have been provisioned with the new password.

### DIFF
--- a/app/provision/index.php
+++ b/app/provision/index.php
@@ -265,7 +265,7 @@ openlog("fusion-provisioning", LOG_PID | LOG_PERROR, LOG_LOCAL0);
 
 //http authentication - digest
 	if (strlen($provision["http_auth_username"]) > 0 && strlen($provision["http_auth_type"]) == 0) { $provision["http_auth_type"] = "digest"; }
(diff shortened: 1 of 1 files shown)


markjcrane:
"Better alternative may be to enable two passwords at the same time temporarily until all devices have been provisioned with the new password."

I agree that a second http auth username and password would be better.

markjcrane:
This is needed so that http authentication can be enabled on phones that don't have it http authentication enabled yet to prevent a chicken before the egg problem.

TAP-Tech:
Two passwords is good, but if you have not set up an http auth username/password on devices, and have not cleared out their default values, you may be in trouble.  I think the option to provision the username/password yet not enforce it is a good one.

markjcrane:
Good point. I'm accepting the improvement.

markjcrane:
Should have been made for master branch first.

PR #2331, opened 2017-01-28
Push t41p DND fix from master to 4.2

emaktech:
The T41p provisioning template in 4.2 breaks the DND key which is resolved in template in master. Fixes should be pushed to 4.2 branch.


markjcrane:
Its not longer needed to disable DND on Yealink because we can use the feature code option to point it to the server this is the preferred way to fix this.

PR #2346, opened 2017-01-31
use local mkdir if $_SESSION['event_socket_ip_address'] pints to loca…

daniel-lucio:
…lhost

if $_SESSION['event_socket_ip_address'] points to the local host, it is better to use local PHP mkdir rather than waste resources connecting to the event socket and running a lua script.

### DIFF
--- a/resources/functions.php
+++ b/resources/functions.php
@@ -1897,6 +1897,16 @@ function number_pad($number,$n) {
 	}
 
 	function event_socket_mkdir($dir) {
+		
+		switch($_SESSION['event_socket_ip_address'])){
+			case 'localhost':
+			case '127.0.0.1':
+			case '::1':
+			case '':
+				mkdir ($dir, 02770, true);
+				return true;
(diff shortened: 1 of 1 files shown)


markjcrane:
The idea was to enable freeswitch to see the directories as its own because it would create it. If the web server and freeswitch are running under different users then this is a problem. However on my current installs I changed how the permissions are handled and don't have to worry about this. But this is event socket mkdir so I don't plan to undermine the name unless the function is renamed.

daniel-lucio:
I see your point, but when you are saying " If the web server and freeswitch are running under different users then this is a problem." -- this can be easily fixed by having a group in common, if you see in my patch groups where created with 02770, which fixes this by only having a group in common.

Then, are you suggesting to have another function, like create_paths(), and inside of it deciding if it goes for local mkdir or socket ?

markjcrane:
I have done the group in common.

PR #2364, opened 2017-02-07
Yealink Provisioning - Add Trusted Cert Variable

emaktech:
Allows trusted certificate signing enforcement to be enabled or disabled via variable $trusted_cert_only.

If set to true, enforcement is enabled.
If set to false, enforcement is disabled.
If not set or any other value is entered, provisioning templates set nothing for this field.


markjcrane:
Please signup as a contributor using your github username.
https://github.com/fusionpbx/open-source/tree/master/contributors

PR #2379, opened 2017-02-16
Added Italian translation

drenda:


drenda:
Hi markjcrane, could you tell me where is the list of languages that the user can choose in his profile? I don't see Italian language in the menu yet and I would like to continue the translation checking what I did so far.
Thanks

PR #2392, opened 2017-02-23
Update user_dashboard.php

Ad5t4r:
Calling 'which' via php-fpm on centos returns null.
CentOS has df in path for nginx/php-fpm scripts.


markjcrane:
Please signup as a contributor so we can accept your changes.

PR #2400, opened 2017-02-25
Add the missing billing support for CDR

daniel-lucio:
when you move the sql query from a * to a specific fields one per one, the 3 billing fields were taken out. this patch allows the pull and it will show the correct pricing


markjcrane:
I double checked the code this is really duplicating functionality so closing the issue.

PR #2420, opened 2017-03-13
Fix CID alteration in Ringroups

daniel-lucio:
I really don't know if this is an IP phone issue (tested on grandstream) or a new behavior on FreeSWITCH, but in order to add the prefix to the caller id (name or number) correctly, you need to export instead of set.


daniel-lucio:
Please see my pending pullrequests

PR #2431, opened 2017-03-17
Update app_languages.php

konradSC:
Users app crashes. Missing single quote in file.


The-Operator:
Should be fixed in the latest version awaiting approval.

The-Operator:
Is also fixed in the update.

PR #2439, opened 2017-03-20
Voicemail Change - record_message.lua

shaunjstokes:
Return to main menu while leaving a voicemail should be disabled as it opens backdoor for hackers, also there is no voice prompt for this option.

### DIFF
--- a/resources/install/scripts/app/voicemail/resources/functions/record_message.lua
+++ b/resources/install/scripts/app/voicemail/resources/functions/record_message.lua
@@ -132,10 +132,12 @@
 						freeswitch.consoleLog("notice", "[voicemail] dtmf_digits: " .. string.sub(dtmf_digits, 0, 1) .. "\n");
 						if (dtmf_digits == "*") then
 							--check the voicemail password
-								check_password(voicemail_id, password_tries);
+							--	check_password(voicemail_id, password_tries);
 							--send to the main menu
-								timeouts = 0;
-								main_menu();
+							--	timeouts = 0;
+							--	main_menu();
+							--this should be disabled as it opens backdoor for hackers, also there is no voice prompt for this option
+							session:hangup();
 						elseif (string.sub(dtmf_digits, 0, 1) == "*") then
 							--do not allow dialing numbers prefixed with *
 							session:hangup();


markjcrane:
What you are calling a vulnerability is a standard voicemail feature.

SJS87:
This is a known vulnerability used by fraudsters to exploit FusionPBX and commit financial fraud.

Anyone calling in from a PSTN number (i.e. via SIP trunk) and leaving a voicemail message on FusionPBX can simply press * then be prompted for the password if they succeed they press 2 to listen to saved messages then 5 to redial the number. This puts the user in a position of trust to choose a secure voicemail password at a significant financial risk to the business.

Unless there are other mechanisms in place such as the following then the cons out way the pros:
- Enable\Disable option for access to voicemail main menu while leaving a message.
- Option to enforce a secure password policy for voicemail accounts.

markjcrane:
Good voicemail password have always been important a good password is provided by default. However there are sadly people that replace random good pin number with one that is much less.

You don't know how many people will complain when the feature is taken away. The bullet points are definitely better suggestions than just disabling it for everyone. Some people want this standard voicemail feature just how many people is hard to tell but will be grounds for more complaints.

markjcrane:
What your pull request does is take a shortcut. Which is the reason it hasn't been accepted. As the real solution is not to take the shortcut. But to add optional pin number strength enforcement and   ability to enable or disable the feature.

konradSC:
I'm interested in PIN strength requirements and account lockout after a specified amount of failures. I believe that would be a useful feature for Fusion.

markjcrane:
Sure it would do you want to make it?

SJS87:
In our environment we have removed the re-dial through voicemail and login access to voicemail while leaving a message, we don't plan to re-enable these features any time in the near future. PIN strength requirement and automatic lockout is on my todo list but currently low priority as we've worked around the problem, I'm not opposed to someone else working on these features should they desire them.

markjcrane:
I have added default settings -> voicemail -> remote_access and set it to false by default. Already had the first complaint but choice secure by default over convenience by default. 

Requiring more secure voicemail passwords also will be a good thing that will be completed in the future.

PR #2440, opened 2017-03-22
Let the apps be more self-contained regarding LUA scripts

daniel-lucio:
At this moment, all LUA scripts must be at resources/install/script, this patch allows applications to be more self-contained by having app/*/resources/install/scripts directory, which it will be copied to the scripts_dir directory as well.

I would like to discuss the addition to this into 4.2.3


daniel-lucio:
are you okay by porting back this to 4.2?

PR #2442, opened 2017-03-23
[master] Add help link to menu for admin and superadmin

mafoo:
Fix for $mod_li and $mod_a_li not being blanked/reset foreach menu
parent item rendered, causing the first menu with sub items to force sub
item rendering for all remaining menu items

### DIFF
--- a/app/help/app_config.php
+++ b/app/help/app_config.php
@@ -1,0 +1,12 @@
+<?php
+
+	$apps[$x]['name'] = "Help";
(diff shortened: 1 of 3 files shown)


konradSC:
I prefer not to put a link to docs.fusionpbx.com in source. If a system is branded or using a small subset of fusion, then they do not need to see all the help resources or may have a branded help site of their own. Better to build your own menu and link.

markjcrane:
Please move this to https://github.com/fusionpbx/fusionpbx-apps

PR #2443, opened 2017-03-23
[4.2] Add help link to menu for admin and superadmin

mafoo:
Fix for $mod_li and $mod_a_li not being blanked/reset foreach menu
parent item rendered, causing the first menu with sub items to force sub
item rendering for all remaining menu items


markjcrane:
This should be an optional feature. Please add it to https://github.com/fusionpbx/fusionpbx-apps

PR #2445, opened 2017-03-23
device export

andycol:
gives the ability to export your devices


markjcrane:
This exports device data only from the devices table.
It does not not include data from

- device_lines
- device_keys
- device_settings

andycol:
yes that is correct i did this so you can pull a csv with extensions and mac addresses.
many customers requested thus

PR #2451, opened 2017-03-25
Voicemail tutorial

konradSC:
This change creates a voicemail tutorial for a voicemail user on their first login. Once the tutorial is completed the user will not be prompted again. An admin has the ability to turn the tutorial back on for a voicemail box if necessary. 

The tutorial asks 4 questions. The user has the ability to skip any of these. 
#1 - Do you want to setup your voicemail?
#2 - Do you want to record your name?
#3 - Do you want to change your password?
#4 - Do you want to record a greeting?

If #1 is skipped, the tutorial will finish and route them into the main menu.
If #2, 3 or 4 is skipped, then the tutorial will simple go to the next step. 

After the tutorial the user is routed to main menu. 

Existing "Callie" recordings were used.

### DIFF
--- a/app/extensions/extension_edit.php
+++ b/app/extensions/extension_edit.php
@@ -385,6 +385,7 @@
 								//if voicemail_uuid does not exist then get a new uuid
 									if (!isset($voicemail_uuid)) {
 										$voicemail_uuid = uuid();
+										$voicemail_tutorial = 'true';
 									}
 
 								//add the voicemail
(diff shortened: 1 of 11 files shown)


konradSC:
I had a conflict in voicemail_edit.php. 

This will need to be added under "default values". 
`if (strlen($voicemail_tutorial) == 0) { $voicemail_tutorial = "false"; }`

PR #2459, opened 2017-03-29
BugFix [master] Added missing execute rights

mafoo:


daniel-lucio:
Please explain why in the world you would need files with 755 if they are run by a HTTP/PHP server, not CLI.

Only few PHP would need to have the +x

moteus:
Why php files need `+x` ? You always can run like `php file.php` whithout this rights. And Fusion uses this way to run php.

mafoo:
something in the install process is adding +x to these files (from fusionpbx-install.sh) so when you come to perform a upgrade it will moan that there were modified files so will block the upgrade. This patch was so going forward it wouldn't block the upgrade, but as to why they need +x I agree it shouldn't do but because something was doing it we needed to compensate

PR #2462, opened 2017-03-30
Grandstream templates for 2130/35/40/60/70 and modern firmware

kusznir:


markjcrane:
Something is wrong with this pull request. Need to make a new one.

PR #2463, opened 2017-03-30
Added some Grandstream templates

kusznir:


kusznir:
This is the long-awaited set of Grandstream templates for the 21xx series of devices (that actually work), specifically GXP2130, 2135 (new phone), 2140, 2160, 2170 (new phone).  Use Line keys for the 2135/2170 for all buttons.  All other phones, line keys map only to the 2/3/4/6 line keys that phone has. 
 2140/2160 has "memory buttons" for the multi-purpose keys.

markjcrane:
Thanks have applied these new templates to 4.2 branch and will have Len add them to Master.

PR #2465, opened 2017-04-01
Added variables to disable call waiting and t.38 faxing

kusznir:
This adds some variables and comments to the ATA config to allow disabling of call waiting and t.38 faxing (needed in the case of integration with a physical PBX or fax machine, and t.38 isn't supported by all carriers)


markjcrane:
disable_cw please change to disable_call_waiting
disable_t38 please change to disable_fax_t38

Might be useful to prefix both with grandstream_ so they would naturally be groups together in the default settings.

PR #2475, opened 2017-04-05
BugFix for VM Transcription

konradSC:
VM Transcription uses WAV files even if 'vm_message_ext = mp3'. This causes issues with playback when dialing into voicemail to listen to messages.

### DIFF
--- a/resources/install/scripts/app/voicemail/index.lua
+++ b/resources/install/scripts/app/voicemail/index.lua
@@ -157,6 +157,13 @@
 						storage_path = storage_path:gsub("${voicemail_dir}", voicemail_dir);
 					end
 				end
+				if (settings['voicemail']['transcribe_enabled'] ~= nil) then
+					if (settings['voicemail']['transcribe_enabled']['boolean'] ~= nil) then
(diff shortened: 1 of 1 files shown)


blackc2004:
BING needs WAV to transcribe, but it should then convert to MP3 and save it as MP3, was it not doing this? It was working in my testing.

konradSC:
On my two lab boxes it is storing them as WAV files. 
vm_message_ext is set to MP3.

blackc2004:
Interesting. I'm not sure this PR is the right way to fix the issue. I think this needs to be fixed in record_message.lua around about line 229.

konradSC:
Agreed. Let me have a look as well.

konradSC:
1. 

This works on line 229 in record_message.lua:
`if (shout_exists == "true" and transcribe_enabled == "false") or (shout_exists == "true" and transcribe_enabled == "true" and voicemail_transcription_enabled == "false") then`

2. 

Also, my boxes didn't have LAME installed so they were not converting to MP3. Once I installed LAME it started working fine. 

I think this PR should be closed and a new one created to fix record_message.lua.

markjcrane:
Closing issue as requested.

PR #2476, opened 2017-04-05
LDAP Plugin Improvements

MelonSmasher:
I've updated the LDAP plugin with a few improvements. 

Base DN's do not need to be supplied to bind a user. Now it expects that the username is the `sAMAccountName` attribute. The previous method tried to join the username value in a DN and attempt to bind that over an array of base OUs. In large environments with may nested OU's it would be nearly impossible to supply all of the possible DN's that a user might be a member of and it would require many subsequent requests to the LDAP server when a user attempts to authenticate. The new method attempts to bind using the short domain name and sAMAccountName like so: `ldap_bind($connection, $short_domain . '\\' . $bind_user, $bind_password);`

The `server_host` option is now an array so multiple servers can be supplied. This allows for fallback servers to be supplied. A loop is performed that tries to connect and bind to each ldap server, and a connection is set based on the first successes bind.

Support for `LDAPS` by providing the `secure` boolean in the LDAP settings.

Attached is a screenshot with an example of working settings.

screen shot 2017-04-05 at 3 59 05 pm

### DIFF
--- a/core/authentication/resources/classes/plugins/ldap.php
+++ b/core/authentication/resources/classes/plugins/ldap.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * plugin_ldap 
+ * plugin_ldap
  *
  * @method ldap checks a local or remote ldap database to authenticate the user
  */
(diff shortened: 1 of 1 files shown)


markjcrane:
It was working with Microsoft LDAP and Open LDAP. Which LDAP have you tested this against? To get accepted we need to make sure this works against the two that are currently supported. Need to attempt to be backwards compatible as well.

MelonSmasher:
This was tested against two Microsoft Activate Directory servers. I was unable to get the original method working with Active Directory.

I can set up an Open LDAP environment today to test. 

An off the cuff idea I just had for backwards compatibility would be to make a Boolean that signifies that you want to use this method to connect to LDAP. If that Boolean does not exist it would use the previous method.

markjcrane:
Needs to be tested with Open LDAP to make sure its still works.

daniel-lucio:
This patch has a big error. It is a good approach but it won't work on environments where 1 AD host more than 1 domain. Example:
ou=Support,dc=domain,dc=local -> support.domain.local tenant
ou=Customer1,dc=domain,dc=local -> customer.domain.local tenant

using $short_domain . '\\' . $bind_user is a bad idea, because it will authenticate and it wont allow fusion to detect the branch-tenant relationship.

It is responsibility of the sysadmin to have a good management of the LDAP. I am working on something that will help.

mschaefers:
I'd love to see improvements on ldap authentication as well, but I'd suggest a different approach:

1. Don't support multiple LDAP servers, because configuration would be a hassle (you would have to configure everyhing below for each server to maintain real flexibility)
2. define a possible list of identifying attributes (email, uid, sAMAccountName)
3. Use an explicitly defined BIND_DN and BIND_PW to find the user's DN only by his identifying attributes
4. allow for a filter expression when finding users, e.g. objectClass=posixAccount
5. have the search base, search depth, configurable as well
6. FInally, use the found users DN (if there is exactly one) for authentication

This would support large environments with nested OUs, as well as OpenLDAP and Microsoft LDAP.

I would be willing to work on a different pull request for that, since this is just our company's requirement.

PR #2484, opened 2017-04-07
Fix call limit

daniel-lucio:
Make easier the call block per domain

### DIFF
--- a/app/dialplan/resources/switch/conf/dialplan/025_call_limit.xml
+++ b/app/dialplan/resources/switch/conf/dialplan/025_call_limit.xml
@@ -1,7 +1,9 @@
 <context name="{v_context}">
-	<extension name="call-limit" number="" continue="true" app_uuid="4670c44c-45dd-4bae-97ba-b0dfe0aca639" enabled="false">
-		<condition field="${call_direction}" expression="^(inbound|outbound)$">
-			<action application="limit" data="hash inbound ${domain_uuid} ${max_calls} !USER_BUSY"/>
-		</condition>
-	</extension>
+        <extension name="call-limit" number="" continue="true" app_uuid="4670c44c-45dd-4bae-97ba-b0dfe0aca639" enabled="false">
+                <condition field="${user_exists}" expression="^true$">
+                        <anti-action application="set" data="domain_name={v_context}" inline="true"/>
+                        <anti-action application="set" data="max_calls=3" inline="true"/>
+                        <anti-action application="limit" data="hash inbound ${domain_name} ${max_calls} !USER_BUSY"/>
+                </condition>
+        </extension>
 </context>


markjcrane:
There are multiple things I think are wrong about these changes.

- Anti-action instead of using action user-exists false.
- Hard coded max_calls there are better places to set this variable.
- Setting domain_name equal to user context its usually the domain name but not always the domain name.

markjcrane:
If you want to get the changes approved please state your reasons for the changes. Sorry 'Fix' is a lousy justification if one doesn't think its broken.

daniel-lucio:
ok, fair enough, I put my reasons here:

- Anti-action is used, because of user_exist variable, the only value I need to know is if the user_exist=true, any other value, included nil/blank means that user doesn't exist.
- max_calls is hard coded because there is no reference to that variable in any other place if there is let me know where. The thinking of this dial plan is to limit calls per tenant, as limits per extension are already placed. 3 is a random number, dial plan is disabled it won't affect
- the domain_name = context is the logic Fusion has been using for a while, since 3.4? (if I am not wrong). You had domain_uuid there, but that variable is not declared anywhere and the domain_uuid is a value that not everyone knows how to extract, domain_name is pretty obvious. Therefore, using domain_name={context}, when creating the new tenant it will be replaced by a hard-coded value that equals to the current domain name.

And the most important of all, I think
- the current dial plan doesn't work out of the box, you turn it on and it won't work.With my patch, you just turn it on and it will work. The only adjust you may need to do is the max_call which I think it is pretty obvious.

markjcrane:
- Anti-action why not use action: user_exists=false 
- max_calls hard coded is more limiting could be set in multiple other places advanced -> variables the first location in this dialplans is another option and other dialplans.
- domain_name = context is currently the default but not the only option for context.

daniel-lucio:
- Anti-action is used because user_action could be other value rather than "false", for example " (blank). And since FreeSWITCH is not able to do negative comparisons (!= or !~), the anti-action is used. If you have a technical argument for this please state. I have stated why I compare against DIFFERENT of "true", rather than equal to "false".

- so instead of putting a default value of max_calls per domain (that's why it is in a dialplan) you prefer to put it in vars.xml which it is global and it will fix all to a static number. no thanks. Putting the variable in the dialplan is the best place as this is not a global value, it is per domain. One could have 1 simultaneous call in domain1, and 6 simultaneous calls in the domain2.

- you have been using the logic of domain_name since 3.6, and now suddenly you change your argument. Furthermore, I think you are misreading the file, it says domain_name={v_context}, which it will hard code the default domain when you create a new tenant, it is {v_context}, NOT ${context}. If I'd put ${context} I would agree with you on this point.

This fix has been designed to work out of the box, turn it on and it will start working. A seasoned VoIP admin will be able to do the proper modifications but a newbie won't.

PR #2485, opened 2017-04-07
Update 025_call_limit.xml

daniel-lucio:
Makes the limit per domain/tenant much easier


daniel-lucio:
Same as 4.3

markjcrane:
There are multiple things I don't like about these changes.

- Anti-action instead of using action user_exists false.
- Hard coded max_calls there are better places to set this variable.
- Setting domain_name equal to user context its usually the domain name but not always the domain name.

PR #2494, opened 2017-04-11
More support to let you add your own provisioning devices

daniel-lucio:
it just flush the v_device_vendor* tables before adding the big array

### DIFF
--- a/app/devices/app_defaults.php
+++ b/app/devices/app_defaults.php
@@ -23,7 +23,21 @@
 	Mark J Crane <markjcrane@fusionpbx.com>
 */
 
-if ($domains_processed == 1) {
+if ($domains_processed > 0) {
+	        
+	//flush everything
+                $sql = "delete from v_device_vendors";
+                $db->exec(check_sql($sql));
+                unset($sql);
+
+                $sql = "delete from v_device_vendor_functions";
+                $db->exec(check_sql($sql));
+                unset($sql);
+
+                $sql = "delete from v_device_vendor_function_groups";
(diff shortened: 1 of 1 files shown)


markjcrane:
This is not a good solution it would be okay as a restore default settings but doing this anytime someone runs Upgrade -> App Defaults is not the best way to do it. These settings are there for a reason and this pull request is a short cut and ruins some of the benefits of them existing in the database.

Under its current form I will not accept the pull request.

daniel-lucio:
Explain more, what benefits?    I just followed your advice about watching how the defaul_settings/app_default.php default does that at https://github.com/fusionpbx/fusionpbx/blob/master/core/default_settings/app_defaults.php

The idea of this is to allow anyone in an already-production-working fusionpbx to add an extra phone support. So, dropping those device tables and adding it again from the array will do it.

The only tables that are touched are the ones related to the soft-buttons. Which it leads to the benefit of not forcing an upgrade to 4.3 if I am in 4.2 and I only need the new support of a new brand added in 4.3

markjcrane:
Default settings are never deleted with running upgrade -> app defaults that would be the equivalent to what this commit does.

> The idea of this is to allow anyone in an already-production-working fusionpbx to add an extra phone support. So, dropping those device tables and adding it again from the array will do it.

Yes it will, but it will also dump any changes so its not the right thing to do. When someone runs upgrade -> app defaults to we dump all the custom changes in default settings? No.

If you want to convert your code into an action that occurs when you press 'Restore Default' on the vendors page then that would be a good improvement.

PR #2495, opened 2017-04-11
Same rationale than 4.3

daniel-lucio:


markjcrane:
This is not a good solution it would be okay as a restore default settings but doing this anytime someone runs Upgrade -> App Defaults is not the best way to do it. These settings are there for a reason and this pull request is a short cut and ruins some of the benefits of them existing in the database.

Under its current form I will not accept the pull request.

daniel-lucio:
Agreed, a button will be. I will send you the new pull request soon. Dont close this yet, I need it as a reference to remember what I did.

markjcrane:
I see that you sent the commit with restore default so I closed this issue.

PR #2501, opened 2017-04-13
Fix. Redirect back to correct profile after delete setting.

moteus:
Fix. Add setting (seems need also set sip_profile_uuid)
Change. Order settings by name when output.


daniel-lucio:
is 4.2 affected?

PR #2510, opened 2017-04-14
Update app_languages.php

vkrt-11:
Russian is added ru-ru


vkrt-11:
Russian is added ru-ru

markjcrane:
Just a note about your other message the correct place to report issues is not the **unofficial  3rd party forum** you will not get official responses there.

Issues should be reported to https://github.com/fusionpbx/fusionpbx/issues

PR #2512, opened 2017-04-16
Fix renaming domains

ssiegel:


markjcrane:
Good job and thanks for the help.

PR #2514, opened 2017-04-16
Make presence for conferences work out of the box

ssiegel:
Use '@' instead of '-' for separating conference name and domain, which
is what FreeSWITCH mod_conference uses.


markjcrane:
Please sign up as a contributor so we can accept your pull requests.
thttps://github.com/fusionpbx/open-source

PR #2515, opened 2017-04-17
Update app_config.php

konradSC:
Add the necessary permissions in order to use the database save function (fusion 4.3) when coding for voicemail option adds/deletes/updates.

I realize these voicemail permissions overall will probably get cleaned up even more once the whole app is updated to use the database function, but this is a stop gap measure. I am working on adding voicemail options to the Bulk Account Settings app and I'm stuck without these permissions. The function is kicking a out 403 Forbidden.


markjcrane:
I don't consider this temporary you found missing permissions that need to be added. Good job.

PR #2516, opened 2017-04-17
Add. Display origin DID number for incoming calls.

moteus:
image


markjcrane:
There is another way to achieve this that are you aren't aware of please contact me on a private chat to discuss it easier call the FusionPBX support number 208-906-8227.

moteus:
Ping

moteus:
Resolved conflicts

moteus:
Ping

markjcrane:
This request proposes origin_destination_number to hold the destination that the caller called.
You are getting the information from the caller_profile->destination_number
We also have information from the caller in the following two variables.
caller_id_name
caller_id_number

Following this logic I added
caller_destination

However caller_profile->destination_number doesn't work well to get this information. Reason is that the destination number gets updated multiple times with follow me and ring groups and the destination still doesn't end with the expected value.

markjcrane:
It should work now by adding an global inbound route and one dialplan with it too. Considering changing caller_destination to just Destination on the CDR list. Please test it and let me know what you think.

moteus:
My varian works with call forward and with call transfer correctly
image
First call (with dest 104) is call to 104 which has call forward to some external number.
Second call is answer 104 and then transfer to 105.
Like I told you `caller_profile` is not element in root xml but is part of element in `call_flow` array.
So when FS do transfer it add element to this array with new `caller_profile` element.
FS add such elements to begin of array. So last element in array has initial value.
I also think just add new custom variable but decide do not do this because its possible do without it.
But I test your variant as whell.

moteus:
Your variant is also works.
image
But I have few notes.
1. `Caller Destination` is too long caption for field. All my numbers 11 digits long so it waste too much space for me.
2. Why you need do `export` for those variables? I can only think about loopback channel but not sure is correct bihavior in such case.
3. Why you need export `caller_id_XXX` vars?

markjcrane:
1. Caller Destination too long. In fact I kind of like it as just Destination. Which leaves renaming the other destination field or replacing it.
2. I changed export to set.
3. export also changed to set for caller_id_xxx but I'm still setting the variables so that you get the variable with the value of caller id before it gets modified.

PR #2519, opened 2017-04-19
Change. Make sip settings more consistent with other pages.

moteus:
Fix. Remove domain from sip profile.

image

### DIFF
--- a/app/sip_profiles/sip_profile_domain_delete.php
+++ b/app/sip_profiles/sip_profile_domain_delete.php
@@ -42,35 +42,24 @@
 	$language = new text;
 	$text = $language->get();
 
+	$sip_profile_uuid=@$_GET['sip_profile_uuid'];
+	$sip_profile_domain_uuid=@$_GET['id'];
+
 //delete the data
-	if (is_uuid($_GET["id"])) {
-
-		//get the id
-			$id = $_GET["id"];
-
-		//get the details of the sip profile
-			$sql = "select * from v_sip_profile_domains ";
-			$sql .= "where sip_profile_domain_uuid = '$id' ";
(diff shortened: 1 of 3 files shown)


markjcrane:
I was intending to do this same thing so it would look similar to the dialplan edit page.

However why I have not committed it yet is because of the functions used to write out the HTML. I'm not certain I want to move in that direction.

**Functions**
- echo_input_cell 
- echo_select_cell

**Concerns**

- Plan is to move to a template based system unless a better way is proposed. Your new approach would require the template to use functions but my ideal for a template system is a minimal thin layer that would not include functions in the template. Therefore I prefer to stick with IF statements and an array that will be passed to the template in the future.
- Once a code design decision like this is made then design changes need to be worked on across the entire code base. Otherwise you end up with many many different styles in the code base. I've seen this multiple times and it makes the code harder to maintain when there are many different styles.

moteus:
I am not sure it related to tmeplates.
If you whant use template I belive you will have to rewrite all pages by hand.
I try just make copy/paste but to make just 3 fields for domain I spend more time than rewrite all with this 2 simple functions. And first variant has many errors.

markjcrane:
- On one hand the changes make the interface more consistent and on the other your changes make the code less consistent as only 2 pages would be using this new approach.
- The use of the functions I understand why you used them. I just think that it will make converting code to a template system that doesn't have functions in the future more difficult. The template system change is a big one and to be done consistently would need to be done across the entire project.
- Why have I been considering templates for the 'view' because it would clearly separate the model and controller. In other words to implement MVC or HMVC.

markjcrane:
I have made the front end more consistent with Device edit page and similar to the Dialplan Edit the code follows the Dialplan -> Edit closely.

PR #2521, opened 2017-04-19
Enhance add message stack

mafoo:
Support legacy $_SESSION['message_*']
Use new messages class
New function messages::add($message, $mood, $delay) to simplify adding messages
- [x] Tested on debian 8, Host firefox client

simplification of giving messages, allows multiple messages to be displayed with individual moods and delays

### DIFF
--- a/core/users/user_edit.php
+++ b/core/users/user_edit.php
@@ -63,8 +63,7 @@
 		}
 		unset($prep_statement, $row);
 		if ($total_users >= $_SESSION['limit']['users']['numeric']) {
-			$_SESSION['message_mood'] = 'negative';
-			$_SESSION['message'] = $text['message-maximum_users'].' '.$_SESSION['limit']['users']['numeric'];
+			messages::add(text['message-maximum_users'].' '.$_SESSION['limit']['users']['numeric'], 'negative');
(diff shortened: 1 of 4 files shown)


resources/functions.php
```diff
@@ -1917,4 +1917,10 @@ function event_socket_mkdir($dir) {
 			return false;
 	}
 
+	if (!function_exists('add_message')) {
+		function add_message($message, $mood, $delay) {
```
moteus: I think need provide default values in parameter list to php 7

markjcrane:
I wondering a bit before effort is used to place across fusionpbx whether it would be better to use an object with add method instead of function add_message. Thoughts on this?

resources/functions.php
markjcrane: PHP 7.1 will error if one of the function params are not set with a default value and not provided to the function when its called.

markjcrane:
A class is more flexible then functions alone.

mafoo:
Class has been implemented as requested

markjcrane:
That looks great! Separates the array data from the display much better this way similar concept to how the menu is handled. +1

PR #2524, opened 2017-04-20
N-Way conference

konradSC:
Allows for network based ad hoc conferencing on devices that support this feature. This will get around the 3-way calling limit that most phones have all allow for unlimited number of people  (N-Way) in ad hoc conferences started by the user on a phone. 

The feature needs to be "turned on" to work. Admin will need to enable this in Default Settings and enable the Dialplan rule. 

If this PR is accepted i'll update the device templates to support this as well.


app/provision/app_config.php
```diff
@@ -401,13 +401,22 @@
...
-$y++;
+		$y++;
 		$apps[$x]['default_settings'][$y]['default_setting_uuid'] = 'bf083e2c-ad73-48ba-a981-9e1c92b2554e';
 		$apps[$x]['default_settings'][$y]['default_setting_category'] = 'provision';
 		$apps[$x]['default_settings'][$y]['default_setting_subcategory'] = 'grandstream_call_waiting';
 		$apps[$x]['default_settings'][$y]['default_setting_name'] = 'text';
 		$apps[$x]['default_settings'][$y]['default_setting_value'] = '0';
 		$apps[$x]['default_settings'][$y]['default_setting_enabled'] = 'true';
 		$apps[$x]['default_settings'][$y]['default_setting_description'] = 'Call Waiting 0=enabled 1=disable';
+		$apps[$x]['default_settings'][$y]['default_setting_uuid'] = '203711b9-1d2a-4b52-9d8c-d6f23b451d8f';
```
moteus: I think you miss this line position :)

konradSC: Good call. Thanks!

markjcrane:
Maybe 'device_nway_conference' as 'nway_conference'? If you like it more as device_ prefix explain some reasoning to consider it.

konradSC:
I like nway_conference. It keeps it consistent. Changes are made.

PR #2547, opened 2017-04-27
BugFix-[master]-mos

mafoo:
mean opinion score is calaculated 1-5 (not 0-5) so values need to be
offset before converting into percentage


markjcrane:
The problem with this change is it doesn't use the translation framework so it hard codes English only.

mafoo:
I was going to make the change to use $text[insername] for those values, but wanted to show you the idea before doing the whole change hence the hard code initially

daniel-lucio:
So, this change proposes to use labels instead of the numeric value. If my understanding is good, these only things on MOS as an integer, late versions of Freeswitch shows it as a float with at least 2 digit precision.

I think it would be better to put colors instead labels., let us red, orange, yellow, light green with yellow, green and keep the number.

mafoo:
For anyone whom had not followed the original IRC conversation about these changes the summary is as follows
my reference is - https://en.wikipedia.org/wiki/Mean_opinion_score

At the moment when you hover over MOS on the cdr's it reports a percentage, the calculation of which is wrong as it basis it on the scale 0-5 rather than 1-5 which is what the stand is. I originally proposed a small update to use the correct calculation however this brought to mark's attention that it would rarely be possible to hit the mythical 5 and accountant type peeps would likely nit pick over not being able to get 'perfect'
At the point of this comment I am proposing to instead display the score but the tool tip to instead display the rating name based on the rounded value (this allowing 4.5 calls to show as Excellent but 4.49 display as Good) however to use out $text system to display the localized wording of that rating.
I like the idea of also colouring the value based on the same index (maybe Green, light green, yellow then Red)

Issue #2559, opened 2017-05-04
active conferences not working on latest commit


andycol:
Below images show 2 calls in the coference but nothing when you click it live

![image](https://cloud.githubusercontent.com/assets/13945173/25699831/2bff6cd4-30c5-11e7-8faa-e48f15dee5fd.png)

![image](https://cloud.githubusercontent.com/assets/13945173/25699840/3831b502-30c5-11e7-9e64-a32da0b98baf.png)

andycol:
fix is to add  $conference_name = substr($name, 0, strlen($name) - strlen('-'.$_SESSION['domain_name']));
in conferences_active_inc.php and 
$conference_name = $conference_uuid.'-'.$_SESSION['domain_name']; in 
in conference_interactive_inc.php

change the @ to a -

andycol:
fix in pull request https://github.com/fusionpbx/fusionpbx/pull/2562

markjcrane:
This fix is based on ignorance. You don't know why we changed from - to @.  And you so called fix is changing it back to what it was.

markjcrane:
I'm testing now to see if this is actually a problem without the fix.

andycol:
A fix is a fix but if you find a better way then thats fine. Maybe you can explain why it was changed to a @?

markjcrane:
I believe all you needed to do to fix the problem that you originally reported is re-save the conference which would update the dialplan.

markjcrane:
The fix was a retreat back to the older way of doing it because the developer that made the change had no idea why it was changed.

andycol:
Ps. Not based on ignorance just trying to help others with the same issue i had.

markjcrane:
The real fix would be making it use the @ successfully.

markjcrane:
But I don't think it was broken. More likely you didn't know that you needed to re-save your conference to get an updated dialplan.

markjcrane:
If bumping into this problem with Apps -> Conference re-save your conference room and it should update the dialplan and work.

PR #2562, opened 2017-05-04
fix the issue where active conferences do not show as per issue #2559

andycol:


markjcrane:
You don't know why we changed from - to @. This changed by a developers pull request and they suggest that it would help with BLF. All your fix is doing it changing it back to what it was originally so I'm not accepting the change.

If bumping into this problem with Apps -> Conference re-save your conference room and it should work.

PR #2570, opened 2017-05-09
Fix MySQL syntax

powerpbx:
Tested on CentOS v7 using MariaDB v5.5.52 and PHP v5.4.16

When reinstalling using the same superuser, install will fail due to incorrect MySQL syntax.  `SET` is used twice instead of once and separating the two commands with a `,`

After that correction the install is able to proceed to the next step which fails because `row['count']` does not exist and is actually `row['count(*)']`.  It will always match `==0` and incorrectly try execute the insertion of the duplicate `v_contact`.


markjcrane:
I'm accpeting your pull request and then changing it in the following way. I will use a SQL Alias adding 'as count' which means that count will exist.

Change this:
sql = "select count(*) from v_contacts ";

To this:
sql = "select count(*) as count from v_contacts ";

PR #2572, opened 2017-05-10
BugFix [master] Languages add missing text placeholders

mafoo:
add placeholders for missing language text
reorganise file so languages are in a consistent order with en-us always first, and the remaining stored alphabetically
space out tag names to make them line up for readability


markjcrane:
Why isn't the organizer function in the in a PHP class?

mafoo:
added the organizer into text class

PR #2573, opened 2017-05-10
Enhance [master] provide $app_path during upgrade

mafoo:
if a application is installed via a symlink PHP will resolve the symlink
for `__DIR__` and `__FILE__` making it impossible for an application to work
out it's relative path withing the fusion system.
By providing $app_path during upgrade routines an application will
beable to provide the correct link for menu items (e.g.
/opt/languages/index.php) instead of having to hard code it where it
could change depending on how the application was installed

### DIFF
--- a/resources/classes/domains.php
+++ b/resources/classes/domains.php
@@ -239,6 +239,8 @@ if (!class_exists('domains')) {
 				$db = $this->db;
 				$x=0;
 				foreach ($config_list as &$config_path) {
+					$app_path = dirname($config_path);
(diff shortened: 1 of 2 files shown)


markjcrane:
Can you explain the regex further for the app_path.

mafoo:
The regex match is being used to get the last two directories off the path
so where the file was found in /usr/src/fusionpbx/opt/my_app/app_config.php it would shorten that to /opt/my_app
I have loaded it into regex101 that can hopefully enlighten it's exact work - https://regex101.com/r/Krj7xI/1

PR #2586, opened 2017-05-19
Use the correct CID

daniel-lucio:
If one extension is forwarding to the PSTN and another extension calls, the PSTN gets the extension number instead the outbound caller id. For example, my cellphone gets 101 as caller instead 5551011234


markjcrane:

daniel-lucio:
This is not broken, 4.0 doesn't have the code implemented in 4.1+. However, I added some missing effective_caller_id_* variables

markjcrane:
No one should use 4.0 anyone still using it should upgrade and get the security fixes included in 4.2.2.

PR #2587, opened 2017-05-19
Fixes CID

daniel-lucio:
Fix the CID
Although you already had somehow fixed the caller id here. you need to put origination and effective. For some reason, not all carriers get the CID from the same field in the SIP payload. origination_* and effective_* affects different parts of the SIP invite.

I also fix the dialstring when using follow_me_fix, by adding a dummy variable at the start of the square brace [


markjcrane:

daniel-lucio:
Fixed, please check

PR #2588, opened 2017-05-19
Fix the CID

daniel-lucio:
Although you already had somehow fixed the caller id here. you need to put origination and effective. For some reason, not all carriers get the CID from the same field in the SIP payload. origination_* and effective_* affects different parts of the SIP invite.

### DIFF
--- a/app/calls/resources/classes/follow_me.php
+++ b/app/calls/resources/classes/follow_me.php
@@ -378,7 +378,7 @@ include "root.php";
 					}
 					else {
 						$presence_id = extension_presence_id($this->extension, $this->number_alias);
-						$dial_string .= "[";
+						$dial_string .= "[anyvar=";
 
 						//set the caller id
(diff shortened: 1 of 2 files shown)


markjcrane:

markjcrane:

daniel-lucio:
I just fix it. Please check.

app/calls/resources/classes/follow_me.php
```diff
@@ -378,26 +378,35 @@ public function set() {
 					}
 					else {
 						$presence_id = extension_presence_id($this->extension, $this->number_alias);
-						$dial_string .= "[";
+						$dial_string .= "[anyvar=";
```
moteus: What is this for?

markjcrane: It was a place holder really so that he could use a comma. I moved presence_id into this position and dumped the place holder.

PR #2589, opened 2017-05-19
Fix. Remove using `call_prompt` field.

moteus:
It marked as deprecated and do not used anywhere.


daniel-lucio:
Just wondering, you don't use that at all right?

moteus:
call_flow.lua do not use this value. 
I make search in repo and could not find any place where  it used.
Also upgrade scheme does not create this field (i suppose because it market as depricated)  in database so this sql is broken.

PR #2590, opened 2017-05-19
BugFix - Convert all languages to 2-2 digit

mafoo:
Some applications were missed due to permissions issue on the server performing the update, this update corrects the missing changes
ordering in xml_cdr was done manually in a previous PR, this corrects the ordering and incorrect description-pdd and description-mos for nl-nl


markjcrane:
Broke the contacts app_languages.php again.

Parse error: syntax error, unexpected 'pipe' (T_STRING) in /var/www/fusionpbx/app/contacts/app_languages.php on line 3145

line 3145
$text['description-import_data']['fr-fr'] = "Copier et coller les données séparées par une virgule ou un "pipe" dans la zone de texte pour commencer l'import.";

I'm changing the " to " to single quotes.

markjcrane:
I just fixed it.

mafoo:
darn it i used php's own addslashes routine to try and avoid this, it would seem that it is not up to the job of what it should do, i will setup some parsing tests and create a solution that can prevent this happening.

PR #2594, opened 2017-05-22
BugFix [master] database class

mafoo:
Fix for if you send an array containing both updates and new data, it
would end up placing all the new data ontop of the last item containing
the uuid of the last update (if updates are earlier in the array than
new items)

### DIFF
--- a/resources/classes/database.php
+++ b/resources/classes/database.php
@@ -595,8 +595,7 @@ include "root.php";
 										//determine if the parent_key_exists
 										$parent_key_exists = false;
 										if (isset($array[$parent_key_name])) {
-											$this->uuid = $array[$parent_key_name];
-											$parent_key_value = $this->uuid;
+											$parent_key_value = $array[$parent_key_name];
 											$parent_key_exists = true;
 										}
 										else {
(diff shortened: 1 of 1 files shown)


resources/classes/database.php
```diff
@@ -595,8 +595,7 @@ public function delete($array) {
 										//determine if the parent_key_exists
 										$parent_key_exists = false;
 										if (isset($array[$parent_key_name])) {
-											$this->uuid = $array[$parent_key_name];
-											$parent_key_value = $this->uuid;
+											$parent_key_value = $array[$parent_key_name];
```
mafoo: if the next record does not contain a uuid, copying the uuid back to $this->uuid causes the next iteration of the array to automatically assume the uuid of the previous record

resources/classes/database.php
```diff
@@ -622,7 +621,7 @@ public function delete($array) {
 									//get the data before the delete
 										if ($parent_key_exists) {
 											$sql = "SELECT * FROM ".$table_name." ";
-											$sql .= "WHERE ".$parent_key_name." = '".$this->uuid."' ";
+											$sql .= "WHERE ".$parent_key_name." = '".$parent_key_value."' ";
```
mafoo: use the $parent_key_value instead as this would get pre-populated correctly on each iteration

PR #2597, opened 2017-05-23
Enhance [master] messages::add convert core/upgrade

mafoo:
convert message handling to use messages class
unify response display to reduce boilerplates
fix for when performing source upgrade with any other operation going wrong (required preservation of cwd)
add missing $text for default permissions
removed surpless $text provided by reources/app_languages.php
correct all br tags
formatting updates for messages class to make multiple messages render
with the same indent in source


markjcrane:
Have to test this from command line to see if it still works.

PR #2600, opened 2017-05-24
New FAKE Call Block Option

daniel-lucio:
I had a trunk that was working very odd. If the PBX sends a SIP answer different than 200 (4xx, 5xx, 6xx), the carrier pushes its voicemail instead of a busy/rejected tone.

This patch allows adding a fake answer. The switch will answer but the caller will listen to the ring tone sequence.
Times and tone sequence is totally configurable through default settings.

I will text you to discuss backporting

### DIFF
--- a/app/call_block/app_languages.php
+++ b/app/call_block/app_languages.php
@@ -353,6 +353,20 @@ $text['label-hold']['ru-ru'] = "Удержание";
 $text['label-hold']['sv-se'] = "";
 $text['label-hold']['uk-ua'] = "";
 
+$text['label-fake']['en-us'] = "Fake";
+$text['label-fake']['es-cl'] = "Falso";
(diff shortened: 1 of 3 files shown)


markjcrane:
Why call it Fake when what you are really do is sending it Tones.

daniel-lucio:
Because it is faking it, user gets the ring (preanswer), but call is answered. Your comment makes me thing you didn't read all. you can change the tone sequence with default settings. Not attached to any specific tone sequense

markjcrane:
I did read it but I think calling it a tone or ring tone is better than labelled as fake. But I do understand why you call it fake but I doubt it's intuitive for most people.

daniel-lucio:
If it is just a label name that stops this pull request, i will rename the label shortly

PR #2613, opened 2017-05-30
Assign a default Domain name instead IP Address 4.2

ruslanguns:
An option that allow you use a Domain name as default, in case if you use the IP instead Default Domain then redirect you to the Default Domain.

Thank you@!
Rusguns!


markjcrane:
Right now I'm assuming this was some kind of accident.

ruslanguns:
yes .... sorry about that. Tried to post a issue post but  for some reason i miss up... so sorry.

markjcrane:
Master branch is the development branch and 4.2 is the current stable branch.

PR #2615, opened 2017-05-30
Fixes variable overwriting

daniel-lucio:
Especially when you are in cluster mode but in other use cases as well, adding a domain resets all the variables.


markjcrane:
Not sure if I want this code I'll spend some time thinking about it. But the so called variable overwrite problem has already solved and everyone who has taken my advanced class is aware of it.

markjcrane:
Hint: look at the bottom of config.lua.

PR #2616, opened 2017-05-30
Same rationale as 4.3

daniel-lucio:


markjcrane:
This is an attempt to add a feature to the release branch which is supposed to be reserved for bug and security fixes.

PR #2624, opened 2017-06-01
voicemail lights up on missed calls

jamesorose:
this behavior is annoying, and creates confusion among users. please add this as the default.


markjcrane:
Looks good. It needs to be applied to all Yealink phones.

markjcrane:
Keeping this open right now as a reminder to add across all Yealink phones.

Len-PGH:
got this applied to the master branch.

PR #2627, opened 2017-06-02
BugFix [master] - Convert all menus to 2-2 format

mafoo:
Have the menu restore create both the legacy 2 digit and 2-2 digit formats of known legacy languages
convert all app_menu.php files to have all placeholders and languages and use 2-2 format for languages


markjcrane:
Create both seems like the wrong way to solve if we standardize on 2-2 then why not use app defaults to fix the settings that are just 2.

mafoo:
Okies i will look at the needed steps to do it within app_defaults

core/default_settings/app_defaults.php
```diff
@@ -1,519 +1,528 @@
...
+					$sql .= "('".uuid()."', 'Yemen', 'YE', 'YEM', 887, '967'), ";
+					$sql .= "('".uuid()."', 'Zambia', 'ZM', 'ZMB', 894, '260'), ";
+					$sql .= "('".uuid()."', 'Zimbabwe', 'ZW', 'ZWE', 716, '263'), ";
+					$sql .= "('".uuid()."', 'Aland Islands', 'AX', 'ALA', 248, '358') ";
+					$db->exec(check_sql($sql));
+					unset($sql);
+				}
+				unset($prep_statement, $row);
+			}
+		//update any defaults set to legacy languages
```
mafoo: This should catch the default_settings that have been set to legacy values

core/domain_settings/app_defaults.php
```diff
@@ -30,6 +30,15 @@
 			$sql = "update v_domains set domain_enabled = 'true' where domain_enabled = '' or domain_enabled is null";
 			$db->exec(check_sql($sql));
 			unset($sql);
+		//update any domains set to legacy languages
```
mafoo: This should catch the domain_settings that have been set to legacy values

core/user_settings/app_defaults.php
```diff
@@ -0,0 +1,40 @@
...
+	Portions created by the Initial Developer are Copyright (C) 2008-2017
+	the Initial Developer. All Rights Reserved.
+
+	Contributor(s):
+	Mark J Crane <markjcrane@fusionpbx.com>
+*/
+
+//proccess this only one time
+	if ($domains_processed == 1) {
+		//update any users set to legacy languages
```
mafoo: This should catch the user_settings that have been set to legacy values

PR #2629, opened 2017-06-03
Enhance [master] - Display application descriptions in native

mafoo:
Where application description is availible in the session language,
display it, if not fall back to en-us then &nbsp;


markjcrane:
- This commit is missing a ; on one of the lines
- It's also not consistent in that in FusionPBX brackets are always used to be consistent. I know PHP allows it without that but it saves time if something needs to be added to IF statement if they are already there.
- Also setting the value of the variable should be before the presentation in this case the echo statements. Notice I have been moving the SQL statements before the presentation as well. This trend needs to continue until we have full separation.

markjcrane:
Accepting it but will be immediately updating it to fix the issues I mentioned.

PR #2654, opened 2017-06-08
Enhance vmail to support direct acces to a VM

The-Operator:
Added the possibility to dial into a voicemail box directly by specifying the destination VM in the dialstring, e.g. *981234 to access VM 1234. Usefull to access shared VM via function keys.


app/dialplan/resources/switch/conf/dialplan/320_vmain.xml
```diff
@@ -9,5 +9,13 @@
 			<action application="lua" data="app.lua voicemail"/>
 			<!--<action application="voicemail" data="check default ${domain_name}"/>-->
 		</condition>
+		<condition field="destination_number" expression="^(vmain$|^\*4000$|^\*98)(\d{2,7})$">
```
moteus: This can bring some problems.
E.g. if some one set `voicemail_authorized=true` in e.g. `Variables` extension
this rules allows access to any vm in the systme.
So may be at least add `voicemail_authorized=false`

The-Operator: Thanks for reviewing the code. Added voicemail_authorized=false as suggested.

PR #2664, opened 2017-06-09
Add. BLF indication for call forward.

moteus:
To display general status of CallForward use blf like `forward+<extension>`
To display status for specific phone number use blf like `forward+<extension>/<number>`

I think good idea is dislplay number specific key like `orange` when CF set but for different number.
But I have no such phone and not sure how to do it.


markjcrane:
In these changes you added find_call_forward to call_flow_subscribe.lua. The call flow subscribe should have nothing to do with call forward I don't see why that was done. If it was to monitor blf for then call forward its seems very out of place in call flow subscribe. I believe it needs its own file like call_forward_subscribe.lua.

moteus:
Using multiple files like that require some code duplication.
I think may be rename `call_flow_subscribe.lua` to e.g. `blf_monitor.lua`
and allow run it as single service. I think it may be convinient.
But also I think it may be need allow run it as set of separate services
because it allows handle more subscribes in parallel.
So it can be run like `blf_monitor.lua [flow] [dnd] [forward]`

moteus:
I add script called `blf_subscribe` which can be used to run services to handle several BLF protocols.
Now it supports Call Flow, Call Forward and DND. I think to add also Follow me as whell.

PR #2668, opened 2017-06-09
Test yealink sample user/group phonebook

dmonty2:
Mark as per our phone conversation, I've attempted to get yealink contacts sync working in conjunction with the grandstream contacts.   The idea is to display the address book on the phone with groups that were created in FusionPBX.  e.g. ITdept, Finance, Accounting

Instead of merging the code, it was easier to include 3 new smarty provisioning templates that support department-groups on the phone itself.
I do not have a yealink phone so the changes need to be tested.  According to Yealink manual the phones can provision both:
1) A Local Contact "contact.xml" which the phone uses for Caller ID lookup.
2) A Remote XML Phone Book"Menu.xml" and "Department.xml"

Instructions are in the README file.


markjcrane:
Add 'Vtech' and 'Add gxp16xx' would be great if both of these were there own separate commits.

PR #2677, opened 2017-06-11
Enhance [master] - database class

mafoo:
If an application uses database class to do it's save operation for app_defaults, when it is executed via the commandline there is no $_SESSION['user_uuid'] so this causes the transaction to fail.
This update is to use the correct uuid-null string when there is no user_uuid
add a SYSTEM_USER with the null-uuid so database transaction will render
this was highlighted by the new number_translation app.

### DIFF
--- a/app/database_transactions/app_defaults.php
+++ b/app/database_transactions/app_defaults.php
@@ -1,0 +1,58 @@
+<?php
+/*
+	FusionPBX
+	Version: MPL 1.1
+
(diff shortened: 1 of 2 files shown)


markjcrane:
Not sure I like this fix. Show me the code that caused this issue.

markjcrane:
I used null user_uuid instead of a fake user account.

PR #2678, opened 2017-06-12
Enhance [master] - database transactions

mafoo:
Change the select statement to a left outer join so when either domain_uuid or user_uuid is null it will still be able to display the data.
This will keep the total number of records the same as the count done above


mafoo:
I have split the update in two, first commit is the actual change second commit is the dos2unix, this should make the changes a lot easier to read

PR #2684, opened 2017-06-12
Enhance [master] - permissions

mafoo:
expand permissions system to make permission checks a bit more streamlined
currently contains demo app to show functionality
 - [ ] Remove demo app


markjcrane:
I'm not going to accept the permissions change right now this will put us further behind on the upcoming release. Also prefer different method names.

PR #2690, opened 2017-06-15
Hide pin codes when a user pushes the redial button

andycol:
This is a security risk as other users can push redial and see the pin code


markjcrane:
I like this change my only hesitance on is that it is not done across all yealink phones.

markjcrane:
Also because it can be security related it seems just setting it to 0 as Disabled might be better than a variable.

markjcrane:
I've paid Len to add this to all yealink phones.

PR #2692, opened 2017-06-15
Enhance [master] - Report when software version is to low

mafoo:
Report when software version is below minimum_version for an application


daniel-lucio:
I like this idea. Can you elaborate a little more. I see you put a miniumum_version element in the app_config which it points to software_version.

1 things.

- There is a numeric_version() function that converts 3.4.2 to 30402, which it can be compared with mathematical functions

I will suggest something related to this pull request if it is accepted.

mafoo:
the main system apps are always always going to need to be the same as the system itself, hence them just using software_version()
an example of a 3rd party app saying it wants a minimum version is my languages one - https://github.com/fusionpbx/fusionpbx-apps/blob/master/languages/app_config.php#L12
which is how we would expect other applications to specify it, for the actual version comparison i am using php's inbuilt version_compare see here - https://github.com/fusionpbx/fusionpbx-apps/blob/master/languages/app_config.php#L12

markjcrane:
I'm really not excited about calling a function from inside app_config.php.

mafoo:
Not a problem. I have changed it all to use the value of 'current' where it is an app included in fusion itself

mafoo:
Somehow my previous commits were not up to date to include the changes I previously commented about. I have fixed and pushed new commits

markjcrane:
minimum current_version not excited about this because it will always changed and there is not guarantee that the future current version will be compatible.

PR #2694, opened 2017-06-16
Fix the alternate destinations when mixing conditions

daniel-lucio:
alternate destination used to have 999
pre sets 100
and custom conditions 5000

when using the 3 in the same time condition. alternate destination will hit and custom conditions will never apply

### DIFF
--- a/app/time_conditions/time_condition_edit.php
+++ b/app/time_conditions/time_condition_edit.php
@@ -461,7 +461,7 @@
 		if (strlen($dialplan_anti_action_app) > 0) {
 
 			//increment group number, reset order number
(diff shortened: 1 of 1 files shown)


markjcrane:
All dialplan orders are 3 digits. In this changing you are trying to set that higher.

markjcrane:
Custom Conditions order number starts at 500.

daniel-lucio:
Maybe, I have seen you have hard coded the 500 in the javascript. and you have +5 increments in the group_id variable. However, sometimes instead of doing 500, it assigns 5000. Some very odd conditions,  doy ou remember having before 5000 at some point?

PR #2695, opened 2017-06-16
Same as 4.2

daniel-lucio:
Same issue


markjcrane:
All dialplan groups are 3 digits or lower for other dialplans.

Seems to me be the better solution would be to fix the entries that have 5000.

PR #2701, opened 2017-06-20
Same as 4.3

daniel-lucio:


markjcrane:
I tested the change and it doesn't work because of 'direction needs to be just direction without the single quote.

After fixing that then there is another error.
SQLSTATE[22012]: Division by zero: 7 ERROR: division by zero

Fixed that and then there was another error. So not applying the change.

daniel-lucio:
Explain me more to fix it. I was reviewing the changes, I don't find 'direction (with a quote). I don't see the problem doing direction='some_value' in the SQL statements.

about the division by zero, I think I know what it is, I will fix it. 

Put me more, I tried here under a live production system and it worked.

Issue #2333, opened 2017-01-30
Voicemail - Return Call


Tatermen:
We have most of our tenants setup with dialling 9 for an outside line. This has caused some fun. For example, on our phone of choice (Grandstream) they would store the call history as the incoming number without the 9, we got around this by editing the inbound dialplan and adding a set action of ```effective_caller_id_number=9${caller_id_number}```. This sends the caller ID to the handset with the 9 in front of the caller ID number, and the call history records it as such - it's inconvenient, but it works.

Now we've hit another issue - a tenant wants to use the option in voicemail to return the call. When they press this, the call hangs up. It appears that in ```scripts/app/voicemail/index.lua``` at around line 419, it stores the ```caller_id_number``` variable in the database.

As the ```caller_id_number``` variable cannot be modified via a dialplan, is there any way around this?

PR #2728, opened 2017-07-10
Use effective_caller_id_number instead of caller_id_number when storing voicemail messages

Tatermen:
This resolves issue #2333. On systems where 9 is used to make an external call we can use effective_caller_id_number in the inbound plan to modify the incoming CLI - for example to insert digit 9. This modified number is then passed to the extensions and the users handset's will store this modified CLI for redial and call history.

This pull request modifies the voicemail recording to use the effective_caller_id_number variable when available when saving voicemail messages, meaning that option 5 to return the call will work in the same manner. Without it, on a system that is using 9 for an outside line, the call back option in voicemail simply hangs up as it has stored the unmodified CLI for which there is no matching outbound dialplan.


Tatermen:
Is there a problem with this?

PR #2729, opened 2017-07-10
Update click_to_call.php

netpro25:
Use of transfer instead of Bridge did not allow setting of outbound caller ID. E.G. When using click to dial to call two external numbers and setting the "Destination Caller ID Number" this setting would be ignored and the source caller ID info would be passed.


markjcrane:
This change will fix one thing and break several other things. If I were to accept the change someone in the future looking at the problems it causes will switch it back to transfer. Either need a switch that can allow people to choose which one they want or need to find a fix for Caller ID with a transfer.

PR #2730, opened 2017-07-10
Add IVR Menu Language Option

emaktech:
Added a field on the ivr_menu_edit page which allows the default_language variable to be set. This is required to fix a bug where when phrases are set to a language other than the default_langauge for the domain, Freeswitch is unable to load the phrase and the call gets disconnected.

Also made changes to ivr_menu in order to make sure dialplan is generated correctly with the new variable. Would apply on next user-resave of the IVR.

Defaults to en unless the user overrides this. Maybe eventually pull the current domain default language, but shouldn't be a big deal as everything is running with en by default anyways.

**IMPORTANT:** As is, the code requires the addition of the _ivr_menu_language_ field in the _v_ivr_menus_ table.

**IMPORTANT:** I use the 4.2 branch so this pull request is for 4.2. I don't think there should be any problems merging with Master but I have not tested on that branch.

### DIFF
--- a/app/ivr_menus/app_languages.php
+++ b/app/ivr_menus/app_languages.php
@@ -54,6 +54,15 @@ $text['message-maximum_ivr_menus']['pl'] = "Maksymalna ilość zapowiedzi głoso
 $text['message-maximum_ivr_menus']['sv-se'] = "Max Antal IVR Menyer: ";
 $text['message-maximum_ivr_menus']['de-at'] = "Maximale Anzahl an Automatischen Vermittlungen:";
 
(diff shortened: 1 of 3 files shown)


markjcrane:
I don't add new features to the stable branch because that would make it more likely to be unstable.

markjcrane:
The language can be set in the dialplan before the IVR Menu. So claiming this is a bug is false. What you are recommending is another way to set the language. But this also reduces functionality as well. Contact me on the phone www.fusionpbx.com support number if you want to know how.

PR #2734, opened 2017-07-12
Enhance [master] - Display moh context

mafoo:
display the context of the moh otherwise if a customer has a name that matches global it is impractical to spot which is which in the drop down


mafoo:
Readjusted to work with current master

markjcrane:
Category default should not be used for a domain category it wouldn't be a default and does add confusion.

PR #2739, opened 2017-07-15
Add CIDR notation check

powerpbx:
Administrators not familiar with FusionPBX will sometimes add IP instead of the full CIDR notation.  No errors are generated and then they cannot figure out why their SIP trunks still don't work.

This is a simple regex check for CIDR notation resulting in generic CIDR notification if not in proper format.

A potential further enhancement would be to check for IP with no mask and automatically assign /32.


markjcrane:
Good suggestion if we only supported IPv4. Notes on cidr for IPv6
https://www.mediawiki.org/wiki/Help:Range_blocks/IPv6

PR #2740, opened 2017-07-17
Enhance [master] - domain counts for superadmin

mafoo:
add default setting for dashboard > system_counts
use dashboard > counts for displaying the domain counts instead


markjcrane:
This commit attempts to change count information code from one non-ideal way to another. The best way to get the count information is with a SQL query direct from the database rather than iterating through the arrays.

PR #2744, opened 2017-07-17
Add IVR language option (master)

emaktech:
Added a field on the ivr_menu_edit page which allows the default_language variable to be set in the XML dialplan for the IVR. This is required to fix a bug where phrases set to a language other than the default_langauge for the domain, Freeswitch is unable to load the phrase and the call gets disconnected.

The error that shows is similar to the following and appears even when the phrase UUID is correct and exists:

`switch_ivr_play_say.c:145 Can't find macro 28852b23-447d-496c-b754-714a19ee0577`

Future improvements may be to allow the creation of a single IVR and then have a dropdown to edit the single IVR in multiple languages.

**IMPORTANT:** As is, the code requires the addition of the ivr_menu_language field in the v_ivr_menus table.

**IMPORTANT:** This is the same functionality for the master branch as was submitted in pull request #2730


markjcrane:
My hesitance on this change that you can set the language before you get to the IVR Menu. Which is why I don't consider it a bug but it still might be convenient as an option.

markjcrane:
This pull request is incomplete as it asks for the language and then does nothing with it.

markjcrane:
Please calls 208-906-8227 to discuss this commit.

konradSC:
This PR does not add 'ivr_menu_language' column to 'v_ivr_menus' table.

PR #2753, opened 2017-07-21
Update provision.php

dmonty2:
Some extensions are accessible to multiple users.  Sort the list to select best directory name match for extension.


markjcrane:
Please explain why the 'like' is needed.

PR #2755, opened 2017-07-24
Fix. Cache class.

moteus:
* `send_event` raise error so `Cache.del` did not remove key or send any event
 * use `memcache` method by default even if `cache` table does not defined in config
 * `Cache.get` did not return any data when use `memcache` method
 * `Cache.get` did not close file. (Its should not be a big problem because GC should do it by self).
 * `Cache.get` can returns some undefined global value. (if method is `file` and file not exists then method returns global `result` value)
 * `Cache.get` does not need check for file existence
 * Value escaping does not needed for `file` method
 * Needed different key escaping for `memcache` and `file` methods
 * Update self test

### DIFF
--- a/resources/install/scripts/resources/functions/cache.lua
+++ b/resources/install/scripts/resources/functions/cache.lua
@@ -33,11 +33,16 @@ if not api then
   end
 end
 
-local function send_event(action, key)
-  local event = freeswitch.Event("CUSTOM", "fusion::" .. cache_method)
-  event:addHeader("API-Command", cache_method)
-  event:addHeader("API-Command-Argument", action .. " " .. key)
-  event:fire()
+local send_event
+if freeswitch then
+  send_event = function (action, key)
+    local event = freeswitch.Event("CUSTOM", "fusion::" .. cache_method)
+    event:addHeader("API-Command", cache_method)
+    event:addHeader("API-Command-Argument", action .. " " .. key)
+    event:fire()
+  end
+else
+  send_event = function() end
 end
 
 local Cache = {}
@@ -63,6 +68,14 @@ local function key2file(key)
     ['\\'] = '_',
     ['/']  = '_',
   })
+end
+
+-- generate random file name
+local function tmp_file(key)
+  -- @todo may be it worth fallback to `os.tmpname`
+  local uuid = check_error(api:execute("create_uuid", ""))
+  if uuid then key = key .. '.' .. uuid end
+  return key .. '.tmp'
 end
 
 -- convert cache key to memcache key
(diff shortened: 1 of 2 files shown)


markjcrane:
* `send_event` raise error so `Cache.del` did not remove key or send any event

Removed fine on multiple servers when I tried it.

 * use `memcache` method by default even if `cache` table does not defined in config

Sounds fine but the setting is a default so it should be added on every system that upgrades.

 * `Cache.get` did not return any data when use `memcache` method

Didn't notice that but didn't look specifically for this so taking your word for it.

 * `Cache.get` did not close file. (Its should not be a big problem because GC should do it by self).

When I closed the file on my server it emptied the value  was trying to send to FreeSWITCH.

 * `Cache.get` can returns some undefined global value. (if method is `file` and file not exists then method returns global `result` value)

At least in one case I returned back the same reply as memcache would return.

 * `Cache.get` does not need check for file existence

It done this way as to kind of mimic a locked file which you you recommended in a previous conversation.

 * Value escaping does not needed for `file` method
True that could be skipped for the file cache.

 * Needed different key escaping for `memcache` and `file` methods
True.

 * Update self test

moteus:
I remove some unrelated code from code examples.

> send_event raise error so Cache.del did not remove key or send any event
```Lua
local function send_event(action, key)
  if (cache.method == "memcache") then
    local event = freeswitch.Event();
  end
  if (cache.method == "file") then
    local event = freeswitch.Event();
  end
  -- here `event` is global object which should be `nil` so it raise Lua error.
  event:addHeader();
end

function Cache.del(key)
  send_event('delete', key)
  -- because `send_event` raise error Lua does not get here
end
```
> Cache.get did not return any data when use memcache method
```Lua
function Cache.get(key)
  local key = key:gsub(":", ".")
  if (cache.method == "memcache") then
    local result, err = check_error(api:execute('memcache', 'get ' .. key))
  end
  -- Because `result` is local inside block this value does not avaliable outside this block
end
```
>>Cache.get did not close file. (Its should not be a big problem because GC should do it by self).

>When I closed the file on my server it emptied the variable value I was using to send to FreeSWITCH.
```Lua
function Cache.get(key)
  if (cache.method == "file") then
    if (file_exists(cache.location .. "/" .. key)) then
      --freeswitch.consoleLog("notice", "[cache] location: " .. cache.location .. "/" .. key .."\n");
      local file, err = io.open(cache.location .. "/" .. key,  "rb")
      result = file:read("*all")
    else
      err = 'NOT FOUND';
    end
  end
  -- here `file` is not avaliable because it is local inside block
  -- so `file` need close there
  --file:close()
end
```

>Cache.get can returns some undefined global value. (if method is file and file not exists then method returns global result value)
```Lua
function Cache.get(key)
  if (cache.method == "memcache") then
    -- here you assign to local `result` which will not avaliable outside this `if` block
    local result, err = check_error(api:execute('memcache', 'get ' .. key))
  end
  if (cache.method == "file") then
    -- here you assign to global `result` (bad idea)
    result = file:read("*all")
  end

  -- here you return some global value `result` which is defined when use `file` method
  -- but when use `memcache` is not defined
  return result
end

result = 'hello'
-- in case of `memcache` method next call returns `hello` for any key
result = Cache.get('any_key')
```
>> `Cache.get` does not need check for file existence

>It done this way as to kind of mimic a locked file which you you recommended in a previous conversation.

Unfortunately Lua same as C has no idea about any file lock. So when you do `fopen` is tottaly undefined whan share mode C runtime will use. As I know it do not use any locks and same file can be opened more then once to write. Also there no any guarantee that it not possible that more than 2 parallel processes/threads pass such condition at same time. So this conditions is quite useless.
In Lua there exists [LuaFileSystem](https://keplerproject.github.io/luafilesystem/manual.html) library which implement file locks on posix/windows systems.

PR #2763, opened 2017-07-25
Detect DTMF when call originates from PSTN

dedayoa:
After setting up the Conference Centre/Conferece Room, I discovered it was not detecting the PINs entered when the call originated from the PSTN via my ITSP. It worked OK, if the call came from a local extension.
This fixes it. Thanks to lesson learnt from earlier in the day - and thanks to MafooUK on IRC

### DIFF
--- a/resources/install/scripts/app/conference_center/index.lua
+++ b/resources/install/scripts/app/conference_center/index.lua
@@ -373,7 +373,9 @@
 
 --make sure the session is ready
 	if (session:ready()) then
-
+		-- start expecting dtmf
+			session:execute("start_dtmf","")
+	
 		--answer the call
 			session:preAnswer();
 


markjcrane:
I appreciate the enthusiasm for this fix. Unfortunately I believe this will cause a problem for many people start_dtmf is used for dtmf type inband. So you should add start_dtmf before getting to the conference center. You could add it as a global inbound route that applies to all incoming calls if all inbound calls are from the same carrier.

dedayoa:
mmhn...I see. Now, I'm learning something again.

 From my tests though, setting start_dtmf did not have any adverse 
effect if using RFC2833/INFO type DTMF (tested with X-lite)

Perhaps start_dtmf is now "smarter"

PR #2780, opened 2017-08-02
Enhance [master] - number translation

mafoo:
couple of new profiles based on feedback


markjcrane:
These would need to be loaded as disabled hopefully because there are some cases where you may need letters.

mafoo:
the profiles don't come active or do anything unless you specifically use them e.g.  <action application="set" data="effective_caller_id_number=${translate ${destination_number} profile_name}" />

PR #2782, opened 2017-08-02
Hide in the phone directory

konradSC:
Hide the extension in the phone directory if Extension -> Directory_Visible = false.


markjcrane:
Nice change.

PR #2785, opened 2017-08-03
Enhance [master] - switch.php

mafoo:
Report a error when vars.xml cannot be opened for writing
add <include> tags arounf the vars.xml to stop code highlighting issues
add a note that teh file is auto generated
make all descriptions render, not just the first one in a category
use indenting to make the file easier to follow


markjcrane:
Split this into just tidy as in your indentation. Seperate from the changes so its possible to see what actually changed.

PR #2792, opened 2017-08-10
Windows ODBC driver and DSN are not required

sergey-mz:
You are using PDO to access database. It doesn't need ODBC driver unless you want to use php_pdo_odbc extension.


moteus:
FS natively supports only SQLite on Windows.
So to be able access to Fusion database from Lua scripts
need ODBC DSN. In other case FS will not works.

sergey-mz:
Where do you set ODBC DSN name for lua scripts to use?

moteus:
dsn name should be same as database name.
and `freeswitch` for FS database if you want to use it for FS core database.

markjcrane:
Advanced -> Databases and set type to ODBC. Advanced -> Upgrade -> Defaults to rewrite the config.sh

PR #2798, opened 2017-08-16
Check Windows ODBC DSN during install

sergey-mz:
Show message only if "fusionpbx" DSN doesn't exist


markjcrane:
Not really sure about the purpose of these changes.
When I setup ODBC on windows never done it with the c:/Windows/ODBC.INI.

Please describe this with more detail as to the purpose of the changes.

sergey-mz:
Currently, setup page on Windows doesn't check if ODBC connection exists and always displays a message about ODBC. It is true that you create ODBC connection with odbcad32.exe, but when you do so c:/Windows/ODBC.INI is updated automatically. The easiest way to check if ODBC connection exists is to check ODBC.INI file. Another option is to check registry. You can replace c:/Windows/ODBC.INI with %windir%/ODBC.ini. If the connection exists then the message is not necessary.

PR #2805, opened 2017-08-26
use UUID native function if extension available

daniel-lucio:
If the module extension is installed, use it. It's faster, it uses system libs. Otherwise, fallback to your random uuid way.

### DIFF
--- a/resources/functions.php
+++ b/resources/functions.php
@@ -117,26 +117,31 @@
 
 	if (!function_exists('uuid')) {
 		function uuid() {
-			//uuid version 4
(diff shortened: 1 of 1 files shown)


markjcrane:
If uuid_create was really a php native function you could find it with php.net/uuid_create which doesn't exist. So please explain the alternative method to install this uuid_create function that must be an add on.

markjcrane:
Finally found it but I wouldn't call PECL a native php function if it was a native php function or even C based PHP extension. This is license LGPL and not an official PHP project which I would say makes it a 3rd party add on. If PHP does add a native function we will adjust the code to use it.
http://pecl.php.net/package/uuid

PR #2806, opened 2017-08-26
Lets reboot all extension with one click

daniel-lucio:
This is a merge of your cmd.php with another file. It allows rebooting all endpoints with one click (all endpoints of the server or all endpoints of the server, depending the show parameter value). Very useful if you have hundreds of endpoints on a server.


markjcrane:
I'm not accepting this pull request because I believe the better way is to add check boxes to the registrations page so that 2 or more devices can be rebooted based on what is selected. I have made changes to the registrations page to make this possible to add. Reboot all phones at once could essentially DOS attack the web server.

markjcrane:
Changes have been completed. If you want the commit accepted don't take the short cut.

PR #2844, opened 2017-09-18
Italian translation added

chrisbware:
A full translation of FusionPBX in italian.


mafoo:
This seems to be reverting a lot of changes and removing a lot of existing translations.
Did you remember to start from master rather than 4.2 as it looks heavily like the 4.2 version?
Try and catch me on IRC and i can see if there is a way to scavenge out just the translation

markjcrane:
As MafooUK has mentioned this must have been based on 4.2 rather than the Master branch. Please work with MafooUK on IRC as he offered to find a way to salvage your changes to get them added to master branch correctly.

PR #2851, opened 2017-09-20
File cache sync via curl

konradSC:
This extends the memcache sync over curl to include file based cache as well. Even with multi-master file replication there is a delay. My people are about to shoot me with the delay. 

**resources/classes/cache.php** -> Sends an event to FS

**scripts/app/servers/resources/file_cache.lua** -> Listens for an event. Sends command via curl to other FS servers. 

**scripts/app/servers/resources/clear_cache.lua** -> Invoked on remote FS server. Clears the cache.


Len-PGH:
just a heads up with syncthing, can adjust the delay on each folder that is sync. some folders make sense to set lower like to 5 or ten seconds rather than 60 the default

konradSC:
Len, I agree however on servers with 10,000 files in the cache directory and syncing it every 10 seconds doesn't seem to be the way to go.

markjcrane:
I'm sure the method works I'm just not certain if this is really the best way to do it. So I'll think about it before accepting the pull request.

PR #2855, opened 2017-09-26
Use BasicEventService to implement MWI service.

moteus:


markjcrane:
Add more explanation if you want it included.

moteus:
This PR allows remove all code related to create/remove/monitor to PID file and handle control events to base class. So service itself does not need it.

PR #2862, opened 2017-09-29
Fix Fusion to work with Homer out of the box

daniel-lucio:
Adding this will make fusion work out of the box by only adding the capture_server_uri variable in vars.xml (or variable menu). This will prevent each time a domain is added (or upgrade script is run) will take out the changes


markjcrane:
Looks like another shortcut.

PR #2863, opened 2017-09-29
Same as 4.2

daniel-lucio:
Same rationale


markjcrane:
Looks like another shortcut.

Issue #2877, opened 2017-10-16
certain extensions showing multiple registrations under accounts


andycol:
certain extensions are showing multiple registrations under accounts/extensions, but are showing once under status registrations
this is on the latest commit on master branch

![image](https://user-images.githubusercontent.com/13945173/31598373-892abc92-b24d-11e7-9755-28b718271db1.png)

PR #2878, opened 2017-10-16
Fix registrations showing on extensions page

andycol:
This will fix the extensions showing as registered multiple times on extensions.php
as per issue
https://github.com/fusionpbx/fusionpbx/issues/2877


markjcrane:
In chat I mentioned to andycol that this fix would cause other problems. However I'm keeping this open as a reminder to take a look at the issue.

markjcrane:
I'm not sure how to reproduce this problem. Does multiple registrations need to be enabled?

markjcrane:
Found a way to duplicate the issue and fixed it.

PR #2882, opened 2017-10-17
Fix MWI for Extensions that with 0

konradSC:
MWI was not turning off after messages were deleted for extensions that start with 0. This requires that real voicemail_id to be pulled from the database and not just taken from the destination number. To get around this we try to send mwi to the original vociemail_id (destination number) and voicemail_id_copy (from database). Index.lua is already doing something similar.

### DIFF
--- a/resources/install/scripts/app/voicemail/resources/functions/menu_messages.lua
+++ b/resources/install/scripts/app/voicemail/resources/functions/menu_messages.lua
@@ -44,10 +44,10 @@
 					--get the voicemail_id
 					--fix for extensions that start with 0 (Ex: 0712)
 							sql = [[SELECT voicemail_id FROM v_voicemails WHERE voicemail_uuid = :voicemail_uuid]];
+							local params = {voicemail_uuid = voicemail_uuid};
 							if (debug["sql"]) then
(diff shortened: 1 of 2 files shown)


resources/install/scripts/app/voicemail/resources/functions/menu_messages.lua
```diff
@@ -41,6 +41,17 @@
 		--message_status new,saved
 			if (session:ready()) then
 				if (voicemail_id ~= nil) then
+					--get the voicemail_id
+					--fix for extensions that start with 0 (Ex: 0712)
+							sql = [[SELECT voicemail_id FROM v_voicemails WHERE voicemail_uuid = :voicemail_uuid]];
+							if (debug["sql"]) then
+								freeswitch.consoleLog("notice", "[voicemail] SQL: " .. sql .. "; params:" .. json.encode(params) .. "\n");
+							end
+							local params = {voicemail_uuid = voicemail_uuid};
```
moteus: It make sence create params before try log them

konradSC: Good catch!

PR #2890, opened 2017-10-23
Sort by CIDName if populated, otherwise CIDNumber

TAP-Tech:
I believe this will help users to sort through the caller ID numbers on the dashboard.


markjcrane:
Thanks!

PR #2891, opened 2017-10-24
Azure TTS

daniel-lucio:
This is the first of a set of pull requests that will give FusionPBX TTS capabilities in many places. This is the main class that will interact with the other modifications.

Please suggest if this is the best place to put this file or if a /app/azure/resources/class is best. I am not sure about this.

This class is sponsored by my friend since he paid some Indian guys for a development and I discovered they were offering/selling that private development to other companies. He took the call to make this capability free.


markjcrane:
I suggest creating app/azure as this will be more modular.

daniel-lucio:
Agreed, I will fix this.

daniel-lucio:
Fixed, in new location.

PR #2892, opened 2017-10-25
Small optimization with intercept.

moteus:
In cluster installation first node search channel uuid
and send it inside special SIP header. So other node
can just use it and do not call Lua script again.


markjcrane:
Have you added the chang to the liability script to add the custom sip header?

markjcrane:
Your change looks intersting but it seems that only part of it is there? As in what sets the sip x header?

moteus:
Code was there all the time.
But it handaled by [intercept.lua](https://github.com/fusionpbx/fusionpbx/blob/master/resources/install/scripts/intercept.lua#L109-L115) Lua script now.
In fact this changes I made when implement intercept at first place.
But just did not add them to PR.
Now I update my PBX and found that.

markjcrane:
Okay thanks

PR #2893, opened 2017-10-26
Update to provisioning engine

hcaldicott:
Adding this fix enables both remote phone book and provisioning to work correctly on the newly released android based Yealink devices.

It appears that during the provisioning process, the T5X series supply a colon formatted mac address in their user agent.

However, when requesting a remote phone book, the presented user agent has no colons. This allows both formats to work as expected.


markjcrane:
Please signup as a contributor at this link so that we can accept this change.

markjcrane:
Thanks

PR #2894, opened 2017-10-26
Fix. Continue show old recording files on CDR page.

moteus:
After upgrade there no info about recordings in database
so to be able display thous files need continue check it.

### DIFF
--- a/app/xml_cdr/resources/classes/xml_cdr.php
+++ b/app/xml_cdr/resources/classes/xml_cdr.php
@@ -51,6 +51,41 @@ if (!class_exists('xml_cdr')) {
 		public $include_internal;
 		public $extensions;
 
+		/** Return full path to record file or false.
+		 * This function does not guarantee that file actually exists.
+		 */
+		static public function recording_file($row) {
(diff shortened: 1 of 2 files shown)


markjcrane:
Maybe you guys should talk to me rather than wasting time on work that didn't need to get done.

markjcrane:
This has already been addressed you just aren't aware of how to do it.

markjcrane:
We can get the name to put in the database.

markjcrane:
So that it doesn't need a work around for old recordings.

moteus:
I understand that it possible just do `update set record_name=uuid||'.wav' ....`
But I have only few ext which need this. Also I remove old recordings after some time.
So there choice either write more comlicated SQL, or write external script which scan FS and
updata database or change PHP part. I choose last one.
Also my modifications put this logic in one place so it will be easy to remove it after some time.

markjcrane:
But this hides the fact that it wasn't upgraded completed. I also changed the recording download and increases its level of security and now the download and playback are unlikely to work with your fix.

The additional .ogg support you put in your pull request is completely fine.

PR #2896, opened 2017-10-27
xml cdr recordings

moteus:
1. This makes easy to upgrade.
2. It is also checks either file exists even field in database not empty.
 I need this because I do not like when CDR page displays play button but when
 you press it it does nothing.
3. Because all logic inside single function it will be easy to change it.
e.g. if you think it is bad idea or need new settings it quite easy to implement.

### DIFF
--- a/app/xml_cdr/resources/classes/xml_cdr.php
+++ b/app/xml_cdr/resources/classes/xml_cdr.php
@@ -51,6 +51,41 @@ if (!class_exists('xml_cdr')) {
 		public $include_internal;
 		public $extensions;
 
+		/** Return full path to record file or false.
+		 * This function does not guarantee that file actually exists.
+		 */
+		static public function recording_file($row) {
+			// new in 4.3 save file metadata in database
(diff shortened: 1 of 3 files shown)


markjcrane:
You need to talk to me directly not via github and then we can likely come to an agreement. But right now you see some of what is going on this subject but not all of it.

markjcrane:
Sad you decided not to talk to me about this. However here is an upgrade script that will fix the problem. Its only required to be used one time for .wav and one for .mp3. We know the path and the file name which is the uuid of the call and wav or mp3. And so this can be handled consistently without changing the code to support the old way.  You can take section of the php code and run it on windows on the advanced -> command page

Instructions for Linux
cd /usr/src
wget https://raw.githubusercontent.com/fusionpbx/fusionpbx-scripts/master/upgrade/record_path.php
php record_path.php

markjcrane:
I have updated the upgrade instructions in the documentation.
http://fusionpbx-docs.readthedocs.io/en/latest/advanced/version_upgrade.html

PR #2907, opened 2017-11-14
Add. Use `route_to_bridge` module to build routes for ring groups.

moteus:
This commit has several improvements
1. Select only needed fields. (do not select quite big XML text strings)
2. Filter routes also by context name
3. Filter dialplans also by hostname
4. Handle conditions based not only `destination_number`
5. Handle `break` and `continue` attributes for extensions
6. Escape vars inside dial-string
7. Add log messages similar as FS dialplan do
8. Handle inline set/export actions
Also its fix bug when script may set 
```Lua
dial_string = "regex m:~"..destination_number.."~"..r.dialplan_detail_data
```

### DIFF
--- a/resources/install/scripts/app/ring_groups/index.lua
+++ b/resources/install/scripts/app/ring_groups/index.lua
@@ -30,6 +30,9 @@
 --include the log
 	local log = require "resources.functions.log".ring_group
 
+-- include libs
+	local route_to_bridge = require "resources.functions.route_to_bridge"
+
 --connect to the database
 	local Database = require "resources.functions.database";
 	dbh = Database.new('system');
@@ -118,6 +121,7 @@
 		uuid = session:getVariable("uuid");
 		context = session:getVariable("context");
 		call_direction = session:getVariable("call_direction");
+		accountcode = session:getVariable("accountcode");
 	end
 
 --default to local if nil
@@ -170,6 +174,9 @@
 	--	error();
 	--end
 
+--get current switchname
+	hostname = trim(api:execute("switchname", ""))
+
 --get the ring group
 	ring_group_forward_enabled = "";
 	ring_group_forward_destination = "";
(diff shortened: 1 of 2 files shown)


resources/install/scripts/resources/functions/route_to_bridge.lua
```diff
@@ -253,51 +272,118 @@ local function self_test()
...
+
+local function preload_dialplan(dbh, domain_uuid, fields)
+	local hostname = fields and fields.hostname
+	if not hostname  then
+		require "resources.functions.trim";
+		hostname = trim(api:execute("switchname", ""))
+	end
+
+	local dialplan = {}
+	local current_dialplan_uuid, extension
```
moteus: Remove unused vars

resources/install/scripts/resources/functions/route_to_bridge.lua
```diff
@@ -253,51 +272,118 @@ local function self_test()
...
+	dbh:query(select_outbound_dialplan_sql, {domain_uuid=domain_uuid, hostname=hostname}, function(route)
+		if context and context ~= route.dialplan_context then
```
moteus: Define `context`  var

resources/install/scripts/resources/functions/route_to_bridge.lua
moteus: Remove unused vars

resources/install/scripts/app/ring_groups/index.lua
```diff
@@ -318,7 +325,7 @@
 					AND r.domain_uuid = :domain_uuid
 					AND r.ring_group_enabled = 'true'
 			]];
-
+
```
moteus: Remove whitespace

markjcrane:
What is the purpose of setmetatable?

moteus:
>What is the purpose of setmetatable?

In this case I set `metatable` with `__index` metamethod. Lua will call this metamethod when
there access to unexisted key. And this metamethod try lookup var inside current session object.
So this allows pass all session variables and override some of them if we need it.
```Lua
local params = setmetatable({
	destination_number  = destination_number,
}, session_mt)
print(params.destination_number) -- get value from `params` table
print(params.context) -- there no `context` inside `params` so Lua try use `__index` metamethod
                      -- as result will be called function `session:getVariable('context')`
````

PR #2908, opened 2017-11-14
Bug Fix - Match correct database column names

konradSC:
call_recording_name & call_recording_path are the correct values.


markjcrane:
This is debateable and would break the code in multiple places.

konradSC:
&#128077;

PR #2911, opened 2017-11-17
Fix Typo

konradSC:


markjcrane:
I fixed all the other SQL queries that gave an error during the install or on the login.

PR #2925, opened 2017-11-23
Removed all yealink {$mac}.boot files

TekMason:
When available to the phones other .cfg files aren't requested.  More testing required.
I'll take run at figuring out the 404 their non-existence causes later.


TekMason:
More info on auto-inserts and missing templates here:
https://gitlab.com/tekmason/PubDevNotes/blob/master/YealinkScripts.xlsx

PR #2930, opened 2017-11-26
dss key length

TekMason:
created new variable yealink_dsskey_length
modified t54s to include it


TekMason:
yealink_dsskey_length-0
yealink_dsskey_length-1
yealink_dsskey_length-2

PR #2936, opened 2017-11-28
Same as 4.2

daniel-lucio:


daniel-lucio:
I will send you the rediff soon. I think this is a important patch

PR #2938, opened 2017-11-29
Fix. Do not request number when do call forward toggle.

moteus:
It is consistent with enable/disable actions.


markjcrane:
I will accept your pull request but I will go back and add <action application="set" data="request_id=false"/> as removing this hides that fact that it could be requested and false should accomplish the same thing but it would be known that you could enable it if desired.

PR #2942, opened 2017-12-11
Update xml_cdr.php

konradSC:
Add check for duplicate call records. Duplicate records will cause the entire db insert to fail. I have tested this with the /app/xml_cdr/xml_cdr_import.php method for cdr insertion.


konradSC:
This PR looks a little scarier than it actually is. A check for duplicates was added. The rest is just formatting changes due to the additions.

markjcrane:
I just opened the changes and compared with Beyond Compare. It does a far better job showing comparisons than github or git. For example changes show in red and indentation shows in purple.

konradSC:
Thanks. I just didn't want you to reject it b/c it looked ugly. It was beautiful on the inside and that's what counts  : )

PR #2948, opened 2018-01-01
Rendering error of .cfg file in XML format

anshkatriya:
Check whether loaded file is xml or not, if not xml than set Content-Type to text/plain


markjcrane:
Please signup as a contributor here.

markjcrane:
Thanks... Accepting your change.

PR #2958, opened 2018-01-12
email_templates Russian 

Pro-pra:
Add Russian translation strings


markjcrane:
Thanks Pro-pra accepting your translations.

PR #2962, opened 2018-01-25
BugFix - lua Send_email

mafoo:
when sending email via freeswitch make function identical to smtp
version for parameters
use freeswitch_from* for email from address where availible
use correct tab indent in function
remove the html escaping from the missed_call template, it is text/plain
have failure_handler and hangup handlers missed function use send_email
instead


mafoo:
Tested on production system at commit https://github.com/fusionpbx/fusionpbx/commit/1af20b8bbd2e2f7dbb8603a828682f183c3635b3

resources/install/scripts/resources/functions/send_mail.lua
```diff
@@ -78,27 +78,46 @@ if not freeswitch then
...
-	if file then
-		freeswitch.email(address, address, mail_headers, body, file)
-	else
-		freeswitch.email(address, address, mail_headers, body)
-	end
+	function send_mail(headers, address, message, file)
+        local domain_uuid = headers["X-FusionPBX-Domain-UUID"]
+        local domain_name = headers["X-FusionPBX-Domain-Name"]
+        local email_type = headers["X-FusionPBX-Email-Type"] or 'info'
+        local call_uuid = headers["X-FusionPBX-Email-Type"]
```
moteus: Wrong parameter

mafoo: Thank you, addressed this and where I copied it from

markjcrane:
I don't see a reason to use freeswitch_from and freeswitch_from_name.

PR #2968, opened 2018-01-29
Add customization for email

ewyand:
Give the ability to customize the smtp from domain and name for voicemail and fax separately


markjcrane:
It looks like a good change. Missing an app_defaults.php that changes email to voicemail for the default settings. Otherwise anyone that upgrades will be broken unless they know about this change.

markjcrane:
So that needs to be completed.

ewyand:
Where would the app_default.php go? Sorry for the dumb question. I do not code. I can semi-read it and figure out what is going on.
I was able to add the smtp_from and smtp_from_name under the respective category ie Fax and Voicemail and it worked for me. I was also able to copy it to the domain and the domain setting took priority.

ewyand:
Will these changes work?

PR #2971, opened 2018-01-30
Changes to 4.2

ewyand:
Add customization for send_from and send_from_name for Voicemail and Fax email


markjcrane:
Change was accepted in master branch but I think it is too much of a change for current stable. See commits I added to fill in missing default settings.

PR #2972, opened 2018-02-02
Sendmail from address to read smtp_from variable.

ConnorStrandt:
Fixed the email from address to prevent sending from and to the same email address and getting blocked by spam filters.


ConnorStrandt:
Adjusting requires. and reverting formatting to create separate commit.

PR #2975, opened 2018-02-02
BugFix - [master] - LUA send_mail()

mafoo:
take call_uuid from the correct header
use smtp_from_name as part of the from address

### DIFF
--- a/resources/install/scripts/resources/functions/send_mail.lua
+++ b/resources/install/scripts/resources/functions/send_mail.lua
@@ -82,7 +82,7 @@ if freeswitch then
 		local domain_uuid = headers["X-FusionPBX-Domain-UUID"]
 		local domain_name = headers["X-FusionPBX-Domain-Name"]
 		local email_type = headers["X-FusionPBX-Email-Type"] or 'info'
-		local call_uuid = headers["X-FusionPBX-Email-Type"]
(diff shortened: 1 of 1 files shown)


resources/install/scripts/resources/functions/send_mail.lua
```diff
@@ -92,19 +92,21 @@ if freeswitch then
 		xheaders = xheaders:sub(1,-2) .. '}'
 
 		local from = settins:get('email', 'smtp_from', 'var')
+		local from_name = settins:get('email', 'smtp_from_name', 'var')
+		local from_address = from_name .. "<" .. from .. ">"
```
ConnorStrandt: This didn't seem to work on my install, it does however see to have the from header already set correctly.

markjcrane: I think its because settins:get is spelled wrong and should be settings:get

markjcrane:
It likely doesn't work because it is spelled wrong settins:get should actually be settings:get. I will fix it.

resources/install/scripts/resources/functions/send_mail.lua
ConnorStrandt: I tried it prior to submitting my pull the other day. Then again maybe I had a typo in mine too (my laptop doesn't like certain keys sometimes).

PR #2978, opened 2018-02-06
Enhance [master] - send_mail.lua

mafoo:
use the to address when email:smtp_from:var is empty
don't use email:smtp_from_name:var when empty

### DIFF
--- a/resources/install/scripts/resources/functions/send_mail.lua
+++ b/resources/install/scripts/resources/functions/send_mail.lua
@@ -1,4 +1,3 @@
-
 local Settings = require "resources.functions.lazy_settings"
 local Database = require "resources.functions.database"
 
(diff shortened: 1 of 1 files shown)


resources/install/scripts/resources/functions/send_mail.lua
```diff
@@ -91,6 +90,12 @@ if freeswitch then
 		xheaders = xheaders:sub(1,-2) .. '}'
 
 		local from = settings:get('email', 'smtp_from', 'var')
+		local from_name = settings:get('email', 'smtp_from_name', 'var')
+		if from == nil or from == "" then
+			from = address
```
mafoo: Maybe we should issue a warning to somewhere if email:smtp_from:var is found to be empty as it is a incorrectly configured system?

resources/install/scripts/resources/functions/send_mail.lua
```diff
@@ -91,6 +90,12 @@ if freeswitch then
...
+		elseif from_name ~= nil and from_name ~= "" then
+			from = from_name .. "<" .. from .. ">"
```
mafoo: We could also issue a warning if email:smtp_from_name:var is blank

markjcrane:
Yes this is much better far more robust.

PR #2980, opened 2018-02-06
Yealink T52s/T54s: Add Variables from T4x Templates + New Ones

emaktech:
Merged in all variables from the Yealink T4x templates (specifically based off T46s) to ensure the T52s and T54s phones provision with similar settings to the T4x phones.

Notable changes include admin user login password is now correctly set, timezone settings are fixed and many other changes. No behaviors have been hardcoded/changed that don't already exist on T4x phones so should not break any existing configurations.

In addition, there are a ton of new variables that can be set in the Fusion GUI. All these new changes should probably also be backported to all the other Yealink templates as well, time permitting.


markjcrane:
{$yealink_bt_enable}
{$yealink_bt_name}

will be changed 
{$yealink_bluetooth_enable}
{$yealink_bluetooth_name}

PR #2981, opened 2018-02-07
Update signup.php

ednt:
A new created special admin-xbound with additional rights was not able to add users.


markjcrane:
Master branch this file has been replaced by user_edit.php which handles add and edit.

PR #2983, opened 2018-02-08
Enhance [master] - send mail lua get from address

mafoo:
**Enhance send_mail.lua**

**Commit 1**
* comment pass

**Commit 2**
* formatting pass

**Commit 3**
create a function to resolve the from address
* start with the default smtp_from
* use the title if availible
* observe a type map for email types that might come in that actually need to use a standard type address (e.g. email2fax notifications)
* if a specific type is defined (e.g. voicemail:smtp_from) use that instead
* if a specific address is defined always use the title
* if a specific title is defined use that (even if the address is default)
* if a address cannot be found use the from address and no title

**Commit 4**
* default settings update


markjcrane:
These changes were to fix the code which I fixed so closing the issue.

PR #2985, opened 2018-02-08
BugFix [master] - default-domain settings var to text

mafoo:
when updating varibles from var to text check if they already exist and
adopt the description
delete the new setting if it is empty

### DIFF
--- a/app/voicemails/app_defaults.php
+++ b/app/voicemails/app_defaults.php
@@ -27,20 +27,6 @@
 //if the number of rows is 0 then read the sip profile xml into the database
 	if ($domains_processed == 1) {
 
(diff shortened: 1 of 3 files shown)


markjcrane:
Every default setting that gets added when upgrading or from the install in FusionpPBX as of 4.2.1 or 4.2.2 gets a specific UUID. This was done to prevent future duplicates but is dependent on the unique ID begin used for that setting. So the duplicates you have indicate that the settings aren't using the unique ID. Because of this new installs will not get duplicates unless a human added it with the web interface. Older installs could get it until the default settings are using the unique uuids that are now preset in the code as of around 10 to 11 months ago.

PR #2988, opened 2018-02-11
Enhance  Call Blocking [master]

The-Operator:
- Added support for regular expressions in numbers to be blocked. Usefull for blocking number ranges used by call centers, etc.
- Added default sort order to GUI display

### DIFF
--- a/resources/install/scripts/app/call_block/index.lua
+++ b/resources/install/scripts/app/call_block/index.lua
@@ -37,7 +37,7 @@ This method causes the script to get its manadatory arguments directly from the 
 	12 Jun, 2013: update the database connection, change table name from v_callblock to v_call_block
 	14 Jun, 2013: Change Voicemail option to use Transfer, avoids mod_voicemail dependency
 	27 Sep, 2013: Changed the name of the fields to conform with the table name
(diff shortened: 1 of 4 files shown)


resources/install/scripts/app/call_block/index.lua
```diff
@@ -104,7 +105,7 @@ This method causes the script to get its manadatory arguments directly from the
 		--check if the the call block is blocked
 			sql = "SELECT * FROM v_call_block as c "
 			sql = sql .. "JOIN v_domains as d ON c.domain_uuid=d.domain_uuid "
-			sql = sql .. "WHERE c.call_block_number = :cid_num AND d.domain_name = :domain_name "
+			sql = sql .. "WHERE :cid_num ~ c.call_block_number AND d.domain_name = :domain_name "
```
moteus: Not sure this support at least SQLite.
Better wrap this code and check database type 
to use this condition only for pgsql.
And even then this leads to different behavior
with different database backend

markjcrane:
Regex comparison would be a slower so be optional. Need a setting to turn it on or off.

The-Operator:
Added option to select regex expression matching and added code to check for pgsql and mysql.

PR #2991, opened 2018-02-13
BugFix [master] - upgrade schema

mafoo:
Actually capture the error if the query fails
display consistently between text and html for style


markjcrane:
I think its likely better to fix the problem rather than painting a bigger sign to highlight it.

PR #2992, opened 2018-02-14
Update switch.php

ednt:
outbound_route_to_bridge() failed for us.
Originally it does not respect the groups and if the order numbers are not continous it fails.
Also was not checked if a condition has a bridge action inside his group.
It happens that a condition of group 0 is used, but a bridge action from an other group.
This should be fixed now.
Also the data records are minimized since now only condition/bridge pairs are gathered.

### DIFF
--- a/resources/switch.php
+++ b/resources/switch.php
@@ -438,87 +438,98 @@ function save_var_xml() {
 }
 
 function outbound_route_to_bridge ($domain_uuid, $destination_number) {
-	//get the database connection
-	require_once "resources/classes/database.php";
-	$database = new database;
-	$database->connect();
-	$db = $database->db;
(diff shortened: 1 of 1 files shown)


moteus:
Couple months ago I wrote example how to use Lua version of route to bridge
function from PHP.
https://gist.github.com/moteus/74c410a0a15509bf4aae84620721f9be

Lua version can handle several conditions in same the extension
execute inline set/export action and eve calls some FS api inside them.

markjcrane:
Changes are required to be added to Master branch so they don't get lost.

markjcrane:
Because github change comparison is so poor I will need to compare this with another tool to see what was actually changed.

PR #2993, opened 2018-02-14
Update fax_emails.php

ednt:
Sending a fax via 'new' in the web GUI worked, but failed via sending per mail.
The account code was not set when sending via mail.
This is now fixed.


markjcrane:
Changes do not go to 4.2 stable first if you don't provide a pull request for master then I'm not very tempted to accept the changes to 4.2.

markjcrane:
If I accept the change on stable branch then it needs to be added on master or the change will be lost with the new release. This puts more work on me and makes me less likely to accept the changes that you want to add. So please add changes to master.

PR #2994, opened 2018-02-15
Update fax_retry.lua

ednt:
If the fax transmission fails the first time, the ident and header was not transmitted during the retries.


ednt:
Also for 4.2 please. I forgot to select the 4.2 branch for my changes.

PR #2995, opened 2018-02-15
Update call_forward.lua

ednt:
max_digit was too low. It was not possible to forward to a german mobile with 12 digits.
According to E.164 https://en.wikipedia.org/wiki/E.164 the maximum allowed length is 15 to be safe, the limit is now 17.
Also increased the timeout to 5 seconds.

### DIFF
--- a/resources/install/scripts/call_forward.lua
+++ b/resources/install/scripts/call_forward.lua
@@ -24,9 +24,9 @@
 
 --set default variables
 	min_digits = "1";
-	max_digits = "11";
+	max_digits = "17";
 	max_tries = "3";
-	digit_timeout = "3000";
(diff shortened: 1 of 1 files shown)


markjcrane:
This was done on purpose to prevent forwarding to international.

markjcrane:
Best approach on these limits would be to move them to dialplan variables. This approach would make it customizable.

markjcrane:
Read this again and I hesitate to allow call forward for 17 digits as this would allow forwarding to international. However the way we do call forward now will work with limitations like toll allow and would work with a pin number. So I think it can be accepted now safely. Older versions of 4.2 before the call forward redesign this would have been a risk.

PR #2998, opened 2018-02-18
4.2 Cloning Menu

canadamcpherson:
Hello,

I think it is a good idea to make menu cloning work using default settings for both 4.2 and master branch.

Example, creating a new menu, and pressing the "Restore Default" button to have a new cloned menu. This way, we don't have to spend a day to have a copy of default menu before making modifications.


markjcrane:
You need to add changes to a clean clone of git repo so it doesn't include things you didn't change.

PR #2999, opened 2018-02-20
Small fixes for RU translation strings

Pro-pra:


markjcrane:
If you look at the changes on this commit some of them are exchanging Russian for English. Please correct the problems with the translation.

markjcrane:
This commit improves some Russian and then changes some from Russian to English.

Pro-pra:
yes, I think that it will be more understandable than Russian translation

markjcrane:
Can you come up with an actual Russian Translation that is intuitive? Just seems odd to replace Russian with English for the translation.

markjcrane:
Rejecting this because its replacing Russian with English for Russian translation just doesn't seem right. I will gladly accept Russian translations.

Pro-pra:
Many Russians do not like translated programs, because the translator translates incorrectly, sometimes the English name is more understandable.

markjcrane:
But with that logic they might as well use the full English translation.

Pro-pra:
Yes, its use full english translation. precisely because the translation is done poorly and it would be better to leave the part in English

markjcrane:
Just seems goofy to create a translation that is a mix.

Pro-pra:
Normal way. see many other translation in other programm. 
IP telephony is a rather complex topic for translation. Terminology is based on English initially.

PR #3000, opened 2018-02-20
Update index.lua

DigiDaz:
Change delimiter for rollover - Stable


markjcrane:
Finally got a chance to test call waiting and it worked fine without changes.
In a bridge statement a , (comma) is used to mean simultaneous, and | (pipe) means sequence.
Rollover is sequence and skipping phones that return a busy.
Comma would be switching it to simultaneous and would actually break the feature.

PR #3003, opened 2018-02-23
Check added if the mac address is already in an other domain available

ednt:
We searched for an hour why a device was wrong provisioned.
Reason: it was still available in our test domain.
To avoid this an additional check was necesarry which also shows in which domain the device already exists.

### DIFF
--- a/app/devices/device_edit.php
+++ b/app/devices/device_edit.php
@@ -298,15 +298,40 @@
 
 				//save the device
 					if ($save) {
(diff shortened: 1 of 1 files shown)


benwiesner1:
can we re visit this?

markjcrane:
This has been fixed by a FusionPBX employee sometime ago just tested on duplicates on the same domain and across domains both notified about the duplicate.

PR #3004, opened 2018-02-23
Check added if a mac address is already available in an other domain

ednt:
We searched for an hour why a device was wrong provisioned.
Reason: it was still available in our test domain.
To avoid this an additional check was necesarry which also shows in which domain the device already exists.


markjcrane:
This would remove the PHP devices class so closing the issue. Code was likely different at the time of writing this change.

PR #3008, opened 2018-03-02
Update xml_cdr_conf_xml() -> Line 1329

futureFry:
adding the domain name into the "xml_cdr.conf.xml" instead of 127.0.0.1.
it failed to post CDRs, if fusionPBX is not default server page.

### DIFF
--- a/resources/switch.php
+++ b/resources/switch.php
@@ -1330,7 +1330,8 @@ if (!function_exists('xml_cdr_conf_xml')) {
 
 		//get the global variables
 			global $domain_uuid;
-
+			global $domain_name;
+		
 		//get the database connection
 			require_once "resources/classes/database.php";
 			$database = new database;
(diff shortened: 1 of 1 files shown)


markjcrane:
Not sure if domain_name is as safe a bet as 127.0.0.1.

futureFry:
I understand. I would say safer, but I guess it depends on the perspective.
Eventually it should be the correct setting when/if DNS is setup correctly.

Should maybe be a option (checkmark: use domain as ... ) in the install process. 
Didn't work for me because I didn't setup the machine as PBX only. So FusionPBX was
setup as site (multiple domain hosting type).

futureFry:
Alternative suggestion:
When the install was started from a domain name not an ip, use that domain name for posting to the CDRs and if not use 127.0.0.1.

mafoo:
maybe a default setting cdr:text:server that is default 127.0.0.1 that can then be leveraged to indicate a different address for non default installs?
If that is the case we may want a new type 'ip_or_dns' which means we can apply some regex testing (i have some formulas that would work well!

futureFry:
let's do this different

PR #3010, opened 2018-03-02
removed double quote from $v_mailer_app path

futureFry:
if this is left in freeswitch (1.6.20) will not start up. 
Freeswitch: Cannot Initialize [[error near line 4105]: missing >]
there might still be a problem on line 1299 -> please correct me if I'm wrong.


markjcrane:
FreeSWITCH is starting with 1.6.20 without this change. What operating system are you using?

PR #3011, opened 2018-03-02
mailer-app & mailer-app-args placeholder missing

futureFry:
prior to this, the mailer-app/args weren't set. Also requires a change in resources/switch.php

### DIFF
--- a/app/call_centers/app_menu.php
+++ b/app/call_centers/app_menu.php
@@ -1,7 +1,7 @@
 <?php
 
 	$y=0;
-	$apps[$x]['menu'][$y]['title']['en-us'] = "Call Center";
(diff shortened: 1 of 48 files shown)


markjcrane:
I removed this because it wasn't working in the past. To put it in will have to do a lot of tests to makee sure it works on all operating systems. Which Operating System did you test this on?

futureFry:
macOS 10.13.3 it worked fine; I also tested in a freeBSD install worked as well. There is another pull request #3011 that kinda of belongs to this one.
But you are right this needs to be tested in other OSes.

markjcrane:
Closing pull requests with merge issue. If you want to add change please update the changes to the new code base and submit a new pull request.

PR #3017, opened 2018-03-15
Add Yealink GUI Language Variable

emaktech:
Adds yealink_language_gui variable to all Yealink templates.

Note: Some templates were forcing "English" through conditional statement or hard-coded.

Now, if no value is set, it will not set any value in the template. This should not hurt anything as default language on Yealink phones is "English"


markjcrane:
Thanks, Better language support is a good thing.

PR #3018, opened 2018-03-16
Update {$mac}.xml

krstork:
Added Screensaver, SIP Transport, and SRTP Mode.
Fixed DNS Mode and NAT Traversal naming.

### DIFF
--- a/resources/templates/provision/grandstream/gxp2170/{$mac}.xml
+++ b/resources/templates/provision/grandstream/gxp2170/{$mac}.xml
@@ -3,7 +3,7 @@
 <gs_provision version="1">
 <mac>{$mac|replace:'-':''}</mac>
 <config version="1">
-<!-- ########################################################################################## -->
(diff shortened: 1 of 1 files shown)


markjcrane:
-{if $sip_transport_1 == 'udp'}{$tp=0}{/if}
-{if $sip_transport_1 == 'tcp'}{$tp=1}{/if}
-{if $sip_transport_1 == 'tls'}{$tp=2}{/if}
-{if $sip_transport_1 == 'dns srv'}{$tp=1}{/if}

This value comes from device lines and this would be line 1 and there can be line 1, 2, 3, and more until the limit of that phone. Your change dumps that and only allows one value for all registrations.

krstork:
I didn’t see where sip_transport was coming from. In PR 3019 I made grandstream_sip_transport. Would it be better to do grandstream_sip_transport_1, grandstream_sip_transport_2, etc for each account?

PR #3020, opened 2018-03-26
Fix dashboard disk space calculation

jbrnd:
This pull request switches the disk space calculation from shelling out to using standard PHP functions. This fixes the calculation, which would previously appear as 0% on the dashboard.

Note that the percentage may be slightly different from what's reported by df, due to the presence of reserved blocks for root.

### DIFF
--- a/core/user_settings/user_dashboard.php
+++ b/core/user_settings/user_dashboard.php
@@ -941,18 +941,8 @@
 
 			//disk usage
 			if (PHP_OS == 'FreeBSD' || PHP_OS == 'Linux') {
-				$df = shell_exec("/usr/bin/which df");
-				if($df){
-					$tmp = shell_exec($df." /home 2>&1");
-				} else {
-					$tmp = shell_exec("df /home 2>&1");
-				}
-				$tmp = explode("\n", $tmp);
-				$tmp = preg_replace('!\s+!', ' ', $tmp[1]); // multiple > single space
-				$tmp = explode(' ', $tmp);
-				foreach ($tmp as $stat) {
-					if (substr_count($stat, '%') > 0) { $percent_disk_usage = rtrim($stat,'%'); break; }
-				}
+				$fractional_disk_usage = 1 - disk_free_space('/home') / disk_total_space('/home');
+				$percent_disk_usage = number_format(100*$fractional_disk_usage,0);
 
 				if ($percent_disk_usage != '') {
 					$hud[$n]['html'] .= "<span class='hud_stat' onclick=\"$('#hud_'+".$n."+'_details').slideToggle('fast');\">".$percent_disk_usage."</span>";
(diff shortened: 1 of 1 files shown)


markjcrane:
Your changes look good. Please sign up here as a contributor and I will accept your changes.

markjcrane:
ping

markjcrane:
If you have questions you can give me a call at +1 208 906 8227

markjcrane:
Please follow instructions or the pull request will be closed.
p.s. Not sure why you care about the home directory.

jbrnd:
I had a look at the CLA, and unfortunately I don't feel comfortable signing that. I'm closing this request.

markjcrane:
It's the Apache contributor agreement not something we invented. Your change is very small the majority of the work was done by php functions. We will not use your code but will likely switch to php functions.

Contributor agreement is there to protect the project from multiple risks and reserve the right to switch license to MIT or BSD license which is less restrivive.

jbrnd:
Concerning the change in this pull request, I am releasing this code to the public domain - feel free to use the code if you like.

But it is indeed relicensing that's my main concern - as a personal preference, I'd prefer that my contributions stay under their original license forever (or at least, that they're guaranteed to stay under an open-source license, a la the FSF contributor agreement).

markjcrane:
You assume the worst of me but its ironic because I've spent the last 10 years of my life working on FusionPBX and sharing it with the world and I'm considering moving it to a more permissive license.

The FSF articles discourages the term "open source" they hate it... as they strongly prefer a form of it as defined in GPL v2 and GPL licenses and to a lesser degree the LGPL. They accept more permissive licenses but don't seem to like them. Look at how MySQL used the GPL with a dual license get companies to buy their commercial license. They are not the only ones that have done this... There are some using the AGPL to do this as well.

My heart is on the other side of the open source camp the one that believes in a license that is more permissive, smaller, easier to understand, asserts less ownership and less restrictive. 

Other projects that feel similarly as shown with their licenses... PHP, GO Lang, PostgreSQL, SQLite, Bootstrap, JQuery, SIPjs, Apache, Android (apache license), FreeBSD, NetBSD, OpenBSD, OpenSSH and many other projects. Check out this article if interested https://www.infoworld.com/article/2611386/open-source-software/github-ceo-backs-mit-open-source-license.html
Don't like this side of the open source world then we aren't interested in your code.

markjcrane:
PHP developers removed MySQL drivers from PHP when the drivers became GPL. FreeSWITCH license is MPL to include MySQL drivers as embedded in FreeSWITCH would cause a license conflict unless compiled separately. ZFS is another example of opensource incompatibility with the GPL. In FreeBSD the BSD license does not conflict with ZFS license and so it can be tightly integrated.

Permissive licenses give more freedom to people as it provides less restrictions for the code to benefit as many people as possible. Contribution of source code changes are encouraged but not forced. Code can be re-used in other projects with less restrictions.

PR #3021, opened 2018-03-29
Update install_fusionpbx.php

powerpbx:
Fix so install works with MySQL.  

`create_database_mysql` does not create schema
but `create_odbc_database_connection` tries to insert records into non-existent `v_database` table and
installation fails.

### DIFF
--- a/core/install/resources/classes/install_fusionpbx.php
+++ b/core/install/resources/classes/install_fusionpbx.php
@@ -545,7 +545,7 @@ include "root.php";
 			//add the database structure
 				require_once "resources/classes/schema.php";
 				$schema = new schema;
-				//echo $schema->schema();
+				echo $schema->schema();
 
 			//add the defaults data into the database
 				//get the contents of the sql file


markjcrane:
If I remember right this caused other problems. Plan to test this with an install before accepting the change.

powerpbx:
This appears related since it causes exactly the same symptoms on PostgreSQL

https://github.com/fusionpbx/fusionpbx/commit/d84bebc1d58d1c51ccef0a855246ff47c833fed1

DigiDaz:
-1 to committing this

Why not just create and maintain your own fork if you want mysql? FusionPBX delivers an excellent PBX appliance with the existing script. Reintroducing mysql compatibility means that unless everything is tested for mysql too then those that use mysql could potentially not get the smooth ride they may expect and then could end up painting the project in a bad light.

markjcrane:
PowerPBX I'll consider your commit I just need to have time to test it before I accept it.

powerpbx:
Thanks.  I have tested on CE7 but not Debian.

markjcrane:
I turned this off because I use the install as a way to recover admin password on a system that is already installed. Would need to add a test to see if schema is already there.

markjcrane:
Accepted this but recommend using PostgreSQL as it is the official supported database for FusionPBX.

PR #3032, opened 2018-04-22
Adding sip_h_Diversion channel variable to list of conditions in dialplan editor.

rstevens011:
Hello first time PR here in this project...

The idea behind this PR is to add the sip_h_Diversion header in the list of conditions for the dial plan editor.  A use case could be, for example, to detect that diversion header is present and then playback some kind of pre media such as "please wait while we connect your call" or what have you.


markjcrane:
Looks like a good idea. Thanks!

PR #3036, opened 2018-04-23
Updates for CDR archives database server

konradSC:
Adds the ability to have CDR archives on a separate database rather than your primary fusion database. To enable the functionality set the CDR archive values in Default settings. 

archive_database=true
archive_database_name
archive_database_username
archive_database_port
archive_database_password
archive_database_host
archive_database_driver


konradSC:
All the files are now updated.

PR #3039, opened 2018-04-26
Prevent NILs for 4.4

daniel-lucio:
When updating from 4.2 to 4.4. Because there are new fields, some of them can return a NIL. this verifications prevents a LUA script failure


markjcrane:
This fix will mask the real problem that the Upgrade was done wrong. This will make it harder to spot the problem and numerous issues will occur.

PR #3043, opened 2018-05-01
Fix tone_stream

konradSC:
This error was being generated in freeswitch:
[ERR] switch_core_session.c:2683 Invalid Application play tone_stream

Needs to use application "playback" for tone_stream.

Had to use a colon instead of a space as the delimited. Otherwise it would parse the dialpan action as "playback tone_stream" and the data as "//%(274,0,913.8);%(274,0,1370.6);%(380,0,1776.7)"

Now shows -->  <action application="playback" data="tone_stream://%(274,0,913.8);%(274,0,1370.6);%(380,0,1776.7)"/>


konradSC:
Now shows the following:

`<action application="playback" data="tone_stream://%(274,0,913.8);%(274,0,1370.6);%(380,0,1776.7)"/>`

PR #3055, opened 2018-05-18
Fix Mysql/MariaDB installation 1/2 [4.4]

daniel-lucio:
Same as devel


markjcrane:
This doesn't test if the schema is already there.

PR #3058, opened 2018-05-21
4.4

f1lint:
Hi,  in /resources/functions.php **function software_version()** return '4.5.0', but last software version 4.4.0
It's normal?


markjcrane:
Pull requests should be submitted for specific features so we can keep track of the changes by topic. If you have any changes please re-submit in this way.

PR #3065, opened 2018-05-23
Add language, dialect and voice selection

emaktech:
This scans the FreeSWITCH sounds directory to find installed languages and then allows the user to select the language, dialect and voice from a dropdown.  The selected option will then be built into the dialplan and saved in the database when saving the IVR.

This fixes many cases of inconsistent language problems with email templates, multilanguage voicemail support, etc. by setting an appropriate installed dialect and voice. This variable is then passed along to the next applications correctly rather than relying only on the default settings on the switch and creating inconsistent behaviors.

example of poor behavior before: Setting language to "fr" on the IVR with default switch settings as "en, us, callie" would set the language to fr-us callie which is an obviously incorrect configuration. By forcing all variables to be set to together in the IVR dialplan it avoids such inconsistent behaviors.

### DIFF
--- a/resources/install/scripts/resources/functions/mkdir.lua
+++ b/resources/install/scripts/resources/functions/mkdir.lua
@@ -1,15 +1,15 @@
 --add the mkdir function
-        function mkdir(dir)
-                api = freeswitch.API();
-                dir = dir:gsub([[\]], "/");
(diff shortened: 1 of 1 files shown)


markjcrane:
This commit would be best to talk about over the phone please call FusionPBX support number so we can talk over the changes in more detail.

PR #3073, opened 2018-05-29
 Fix Yealink Programmable Key Template Conflicts 

emaktech:
Some Yealink templates have hard-coded programmable key options in {$mac.cfg}.

A recent pull request was made to allow provisioning options to be configured on all phones through the  y000000000000.cfg files. This pull request resolves a conflict when configuring the programmable keys in the templates by simply removing the old hard-code options that exist in the {$mac.cfg} files for all devices.

### DIFF
--- a/app/ivr_menus/ivr_menu_edit.php
+++ b/app/ivr_menus/ivr_menu_edit.php
@@ -85,6 +85,7 @@
 		//get ivr menu
 			$ivr_menu_name = check_str($_POST["ivr_menu_name"]);
 			$ivr_menu_extension = check_str($_POST["ivr_menu_extension"]);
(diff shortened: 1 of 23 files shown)


emaktech:
Mainly because:

{$mac.cfg} - contain settings specific to accounts
y000000000000.cfg - contain settings specific to the device.

Line keys, programmable keys, etc. are only assignable per-device, not per-account so it makes more sense to put it in the more general "device" file.

So there's that, plus the programmable key settings for all T5 series phone templates are also in the y000000000000.cfg files even in Yealink's base templates so I just followed what they did.

markjcrane:
Accepting it because it follows Yealink latest templates.

PR #3074, opened 2018-05-30
Fixed gxp2170 template value register_timeout 

kusznir:
Somehow this was static'ed to 60 minutes.  Set with variable so it pulls from device settings now.


kusznir:
Also fix pc passthrough vlan default back to sane value

digitalcrisis:
This does need to be fixed, it caused a lot of headaches.

kusznir:
Fixed more bugs in the templates.

markjcrane:
Some of the changes look good some not so good so a live conversation over a call would be best.

PR #3085, opened 2018-06-05
Add Toll Allow Text Field

andycol:
Makes it easier for "Users" to set toll allow without editing the dialplan


markjcrane:
Please send changes in one pull request that doesn't conflict with each other.

PR #3192, opened 2018-06-15
Changed VM authentication to store user ext

inventor96:
From an auditing perspective, it's more useful to have what extension was used to authorize the call, rather than the extension's password that was used.


markjcrane:
This change would change behavior and interrupt backwards compatibility. It would be better to set a variable that allow it to use the alternative behavior. Please resubmit giving the administrator the choice for the alternative behavior.

PR #3194, opened 2018-06-19
Update app_languages.php

jrenaud007:
modify call timeout in french


markjcrane:
Thanks for the help with improving translations.

PR #3197, opened 2018-06-22
Fix wildcard search for Source

konradSC:
Match updates in 4.4 branch to fix search options


markjcrane:
Looks to me like  'and is_uuid($caller_extension_uuid' should be okay. It was added to improve security. I see how the other change causes the issue so will accept the change and add the is_uuid back.

markjcrane:
I exchanged the 'and' with && to fix the is_uuid.

PR #3201, opened 2018-06-27
Update call_broadcast_edit.php

vindaloovoip:
Functionality added to add phone number by upload CSV or TXT file. 

@markjcrane We have an option to download Sample.csv file so can you please let me know how I can upload that Sample.csv file because I tried upload but not able to upload because of permission.


markjcrane:
You forgot to translate items in this and instead hard coded some English. 
"(Upload TXT- Plain Text, CSV- Comma Separated Values file format only.)"

PR #3230, opened 2018-06-29
Fix Pin Numbers

andycol:
This fixes the dropdown to allow Pin numbers in the outbound route and changes the name to Pin numbers from Pin Codes, it also fixes the permission as per Mark's suggestion on Continued Education


markjcrane:
Made an adjustment to app_languages.php and some of the indentation on a different commit.
https://github.com/fusionpbx/fusionpbx/commit/e8db611998c29d3ec5d3c58ef1a24e9bb870f7e2

PR #3231, opened 2018-06-30
Add wildcard LIKE support to hostnames for dialplans

daniel-lucio:


markjcrane:
This will slow down the SQL query and see no real to support it.

PR #3232, opened 2018-06-30
Same as 4.5

daniel-lucio:


markjcrane:
A like statement will slow down the Query compared to an exact match. I don't see any reason for this change.

markjcrane:
This will slow down the SQL query and see no real to support it.

PR #3236, opened 2018-07-03
Update app_config.php

TekMason:
Addition of device Vendors that already have provisioning folders.


markjcrane:
Reviewed this again and haven't moved this section yet to app/yealink, app/polycom, and app/snom  will do that in the future. Thanks for additional models.

PR #3242, opened 2018-07-14
List does not display with escape

konradSC:
List does not display with escape


markjcrane:
I'll accept your pull request to give credit for bringing this to my attention. However the problem is that it would leave FusionPBX vulnerable to javascript injection. Escaping the uuid doesn't change it at all so if it really is a UUID it will not be changed.

The code that seems to have broken it was:
 implode(', ', escape($contact_name)) 

The escape did not work inside the implode function with a little change it worked with: escape(implode(', ', $contact_name))

PR #3284, opened 2018-08-26
Wired up configurable firmware versions for Grandstream Devices

danry25:
In this pull request the firmware field in the Devices app in core is used as a way to specify what folder and base URL should be served to a device, so that users can easily test beta firmware on multiple devices, or walk devices like phones through a series of firmware upgrades in cases where a direct upgrade to the latest firmware isn't possible.

There is also a [readme](https://github.com/AccelerateNetworks/fusionpbx/blob/d8736a0efd82a60cc219bf103498cc7dbb2e7d6c/resources/templates/provision/grandstream/README.md) in this pull request to help explain how to use this feature.

### DIFF
--- a/resources/templates/provision/grandstream/README.md
+++ b/resources/templates/provision/grandstream/README.md
@@ -1,4 +1,3 @@
-======================
 Device Specific Firmware Versions
 ======================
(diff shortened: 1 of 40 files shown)


markjcrane:
Changes look good.

PR #3403, opened 2018-09-11
Fix PHP out of memory error on App Defaults

shaunjstokes:
If 400+ domains exist it's not possible to Execute App Defaults without getting PHP out of memory error on line 630 in 'app\dialplan\resources\classes\dialplan.php'.

It would be better to first build XML for each domain_name (context) then for all else.


markjcrane:
Please sign up as a contributor and I will immediately accept your changes.

PR #3472, opened 2018-09-13
Allow editing 48 lines on extension_edit page

emaktech:
When provisioning 48 port analog gateway, (Grandstream GXW4248 for example) need ability to configure up to 48 lines per device.


markjcrane:
Thanks!

PR #3473, opened 2018-09-13
Update extension_edit.php

dougstamp:
Added code to update device_key_label on devices that have the extension set as BLF/etc. when an extensions effective_caller_id_name is changed.


markjcrane:
We need a default setting to enable or disable this. If needed I will do that when I get a chance.

markjcrane:
Accepting this for now but will update it to make it more flexible to work according to people's preferences.

PR #3553, opened 2018-09-21
Fix CDR search for extension-restricted users

samael33:
Also introduced variable page_skip_active_users.
If set, then "page" will skip active users.

Uses custom SQL query, not 'channels show like ...' API.


samael33:
This pull request is actually for commit ` [5a218dc](/fusionpbx/fusionpbx/pull/3553/commits/5a218dc09a73726eb84df79365a086f88c2b062b)`

PR #3628, opened 2018-10-05
GXP2170 Provisioning File Improvements

danry25:
Committed this on behalf of Jim Kusznir (who signed the [CLA here](https://github.com/fusionpbx/open-source/blob/master/contributors/JimKusznir.md))

Below are his notes on the modifications he made to this provisioning file for the GXP2170. I have reviewed the changes but not modified what he emailed me (as I saw no need, it looks fine).

# Jim's Changes

Added $grandstream_outbound_proxy_2 to backup proxy server

Removed meaningless Mandatory comments (from Grandstream, they are NOT TRUE)

Removed Broadsoft feature configuration

Made default ring tone a variable $randstream_default_ringtone with no default (makes it user-controlled if var is not set)

commented out call log so user can change setting if desired.

Used variable grandstream_disable_call_wating in place of hard coded default of 0.  Variable exists from past template design.

Deleted disabled call feature code specification

Set RTT estimate to default (100, was 50)

Set distinctive ringtone settings in additional accounts (account 1 was set)

(Due to time constraints, I ignored Accounts > 2 )

Commented out static IP address settings (overridden by dhcp seting)

Proplery separated phone VLAN out to its own variable (grandstream_phone_vlan); default is 0 if unspecified (factory default)

Changed Bluetooth setting: if not set through (existing) grandstream_bluetooth_power, then it is unset (user controlled through phone menus)

Completely deleted unconfigured OpenVPN section (anyone wanting to use that must put extesnive customizations in)

Note: left existing settings to cause it to check for firmware every 60 minutes
Note: left existing firmware path MODIFICATION to {$grandstream_firmware_path}/gxp2170 (P192)

Deleted TR-069 settings

Fixed default ringtone at P8398

Removed defaults if grandstream_city_code is not set (allows users to set themselves on their phone and not have it overwritten if it is not specified in the server)


markjcrane:
There are things about this that I think are good and some things not so good.

* Used variable grandstream_disable_call_wating in place of hard coded default of 0. Variable exists from past template design.

We have a variable called grandstream_call_waiting  and it follows the same naming convention that is used on other templates

* grandstream_phone_vlan 

We have two variables for VLAN that were added before 4.4 release they are: grandstream_lan_port_vlan  and grandstream_pc_port_vlan

Completely deleted unconfigured OpenVPN section (anyone wanting to use that must put extensive customizations in)

Should be commented out. OpenVPN can be an important feature especially at some point in the future so I really don't want it deleted.

I think you have some good improvements. As they currently are there a multiple changes that need to happen to get it ready to accept. Welcome to work on them if or I will likely have to cherry pick the changes.

danry25:
Take a look at the changes I made, pretty sure that should correct the issues you mentioned.

PR #3631, opened 2018-10-12
Setup Grandstream GXP2100 Template and Phonebook

killsudo:
Adds support for GXP2100 and properly provisions a GXP2100 using the last available firmware from grandstream. Successfully configures line and memory buttons along with additional grandstream options only available in later firmware.

Phonebook.xml relies upon the builtin Contacts APP user/group permissions and does not use the 'private function contact_grandstream' to generate the $contacts array. Instead this phonebook works with the current 4.4 $contacts array.


markjcrane:
* Fix private function contact_grandstream so it can work until the code path is removed entirely

This is going to be removed.

PR #3632, opened 2018-10-12
Fix grandstream private

killsudo:
Make 'grandstream_global_contact_groups' usable until the private function contact_grandstream can be removed entirely.


markjcrane:
I'll accept this as I don't accept variables that are gs_ and they should use grandstream_ however I'm planning to remove contact_grandstream entirely unless there is a good reason for keeping it that I'm not aware of yet.

PR #3710, opened 2018-10-25
Add fax_enable_t38=True to dialstring in fax_to_email.php

false-vacuum:
Add fax_enable_t38=True to dial string so when a user sets a Forward Number in a Fax Server, the channel can negotiate t38.

Some of our ATA's are configured to only use T.38 and not fall back to G711, so they weren't receiving faxes that were being forwarded from the fax server.

T.38 seems to be very reliable so hopefully, this will help someone else.

### DIFF
--- a/app/call_block/call_block_cdr_add.php
+++ b/app/call_block/call_block_cdr_add.php
@@ -17,7 +17,7 @@
 
 	The Initial Developer of the Original Code is
 	Mark J Crane <markjcrane@fusionpbx.com>
-	Portions created by the Initial Developer are Copyright (C) 2008-2012
+	Portions created by the Initial Developer are Copyright (C) 2008-2018
 	the Initial Developer. All Rights Reserved.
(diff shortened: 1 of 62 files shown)


markjcrane:
Not everyone wants t38 so this has to be done in a way that is optional.

emaktech:
In my now unfortunately vast experience with faxing, you are by far best off leaving fax_enable_t38 not defined in the fax server configurations and instead setting the variable in your outbound and inbound routes.

The reason is because not all origination and termination providers will correctly support T.38 all the time and you will not always use the same provider for origination and termination. Doing it with your inbound and outbound routes allows you to do proper interop testing and then based on what route a call takes you set the variables as you wish without them being "hardcoded" in the fax dialplans.

markjcrane:
The t38 option can be added currently by defining the variables in the default settings -> fax -> variables.

PR #3772, opened 2018-11-09
Variables for DP750 Handset Assignment, Ring Type

emaktech:
New variables for the DP750 to allow the following settings:

$grandstream_repeater_mode_enable
(Enables or disables repeater mode on DP750 base)

$grandstream_hanging_group_mode
(Allows users to set hanging group mode (e.g. ring handsets at the same time, sequentially, etc.)

$grandstream_handset_X_line_Y_enable
(Allows setting of handset to line assignment)


emaktech:
For the handset line settings, I had originally put an _if_ statement on line 1 for each handset and left the others to be blank but the base was not provisioning without there being a 0 on each p-code field so I had to make this giant mess.

PR #3860, opened 2018-11-19
BLF of agent status "agent+username@domain.com

konradSC:
Add the ability to monitor the availability status of a call center agent with "agent+agentname@domain.com".


konradSC:
This would need to be added to /etc/freeswitch/autoload_configs/lua.conf.xml for the above to function properly. 

```
    <!-- FusionPBX: Support BLF for Call Center Agent Login/Logout Satus -->
    <param name="startup-script" value="blf_subscribe.lua agent"/>    
```

PR #3895, opened 2018-12-04
Add new default_settings for Fanvil products

killsudo:
Work in progress to expand default settings for Fanvil provisioning.

Add multiple useful options for plug and play provisioning for DHCP/LLDP/VLAN/QOS and Firmware


markjcrane:
All default settings are lower case.

PR #3897, opened 2018-12-04
Add initial Fanvil i30 device support

killsudo:
ToDo: i30 template still needs to reflect the master repo sip line variable name changes.


markjcrane:
Once the provision variables are lower case this will be ready to commit.

PR #3900, opened 2018-12-04
Show enabled extension only on operator panel

sergey-mz:
Add e.enabled = 'true' to SQL where and show only enabled extensions on the operator panel.


markjcrane:
Good idea!

PR #3938, opened 2018-12-07
Yealink WiFi settings, Improve Consistency

emaktech:
* Added WiFi settings to all Yealink templates that support WiFi settings.
* Improved consistency across templates
* Add lang.wui = {$yealink_lang_wui} to all templates
* Added lang.gui and lang.wui values to templates that had static.lang.xxx (deprecated in firmware x.84.0.15)
* Added new conference variables to support settings added in firmware x.82.0.20


markjcrane:
Incremental changes are far better.

emaktech:
Incremental in what sense?

markjcrane:
When you change so much all at once that we can’t even tell what is
different then the changes are not safe to accept. Maybe github is making
it look more than it really is I’ll have to use a different tool to compare
the changes.

emaktech:
Oh, I see. These were rebased off of GitHub versions today. There's only about 10 or so lines of code changed on each template, but there are so many templates it's quite a hassle... Also not sure why GitHub seems to think many of the files are completely deleted and re-added, that's not correct at all.

We used software called "Beyond Compare" internally to check differences from live master on every one and it looks very different on GitHub for some reason...

markjcrane:
I use beyond compare too its a nice tool. Your pull request needs to be fixed so it doesn't look like every line changed on multiple files.

markjcrane:
We need a clean change log.

markjcrane:
Will accept this once the pull request is cleaned up.

PR #3939, opened 2018-12-07
4.4 /xml_cdr/xml_cdr.php

Tricom114:
When a basic user checks their call logs, they can not see their outbound calls.  They do see the outbound calls on their dashboard, but when they go to the Call Detail Records, all but outbound calls are shown.


markjcrane:
This pull request is not clean so we don't know what changed. Please try again.

PR #3948, opened 2018-12-13
Fix device_all permissions

emaktech:
Users without device_all permission were able to see all devices on the devices page. This PR adds additional logic to the SQL query to ensure users without device_all permission can only see devices assigned to them.


markjcrane:
Devices were not designed to be visible to a regular user in the user group. However it probably is a good idea to restrict more if they are given access.

This looks like a good start however an admin should be able to see all devices on the domain. This appears that it if the show all permission was not assigned that it would restrict the admin from seeing devices on the local domain.

I will help with this hopefully in the next few days.

emaktech:
Hmmm, yes you're right in this case if an admin has device_all they will not actually see all domain devices which is bad behavior.

My use case for this is to make it possible for normal "user" level users to have a simplified Devices page for them to manage select settings on their own devices. I have also been working on the code for the new GS Wave QR code app and similar concept needs to be built for that so it would be possible to allow regular users to provision their own extensions without having access to every extension on the domain.

For that I was floating the idea of adding a new permission type like device_own or extension_own for example which could be assigned to a user group if you wanted to control permissions that way. Would it make sense to you if I added this to the Device and GS Wave apps?

PR #3961, opened 2019-01-09
Initial template GXV3370 (New style)

StratoSwitch:
Config template V2 from Grandstream. first model with new style of XML template.  
Does not use "P" codes.

MPK settings are not programmed in this commit.

Options with FusionPBX variables are set, but all other options are still set to default.


markjcrane:
Awesome

PR #3963, opened 2019-01-16
New GS templates and update 2170

StratoSwitch:
This will complete the list of current Grandstream telephones.
last remaining template to build is "surveillance general" for cameras.


markjcrane:
Looks good thanks!

PR #3972, opened 2019-01-25
Parent-Child Domain Capability

daniel-lucio:
Gives admin (not only superadmins) the ability to manage more than one domain. Super admins will still need still to create the domain and link it (child of another domain). When this is done, for example, an admin account under a domain like reseller.mydomain.com will be able to manage customer1.mydomain.com with a single login once the superadmin makes customer1 domain a child of reseller domain.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,5 +1,9 @@
 RewriteEngine    On
 
+#RewriteCond %{HTTPS} off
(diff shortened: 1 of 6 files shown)


benwiesner1:
I’ve been looking for this solution for a while. Any news on release?

PR #3995, opened 2019-01-30
Update y000000000077.cfg

mloiterman:
Added variable custom.handset.time_format which is required to properly set the time format for firmware above 81.  This is probably applicable to other yealink configuration files, but this was only tested with a w60b.


markjcrane:
Looks good.

PR #3999, opened 2019-02-03
Show fax file date in user's timezone

emaktech:
Previously timezone was showing in GMT, this now shows fax time in user's timezone. Code to display timezone was carried over from fax_logs.php page.

### DIFF
--- a/app/fax/fax_files.php
+++ b/app/fax/fax_files.php
@@ -367,8 +367,10 @@
 			else {
 				echo "&nbsp;\n";
 			}
+			$fax_date = ($_SESSION['domain']['time_format']['text'] == '12h') ? date("F d Y H:i", $row['fax_epoch']) : date("F d Y H:i", $row['fax_epoch']);
+			
 			echo "  </td>\n";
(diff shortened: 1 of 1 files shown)


emaktech:
Honestly I don't know why it works, I just noticed that the time shows correctly on fax_logs.php so I copied the code that spits out the date from that page and updated fax_files.php to do the same thing while following the date format that was already on fax_files.php

Maybe it's because it's using $row['fax_epoch'] and converting that rather than $row['fax_date']. Either that or $_SESSION['domain']['time_format']['text'] is getting the user's timezone. Or both... Again, I'm not 100% sure why the code from fax_logs.php spits out the correct timezone, it just does.

You committed the code I copied here not too long ago here:
https://github.com/fusionpbx/fusionpbx/commit/e3ceb8ba8feeae121671b02bddc028b1c36a6723#diff-bdb37d988c2109d11c4b88653db8cc05

Hopefully this makes sense, I'm not much of a developer, I only know enough to get myself into trouble.

PR #4000, opened 2019-02-07
Azure getSubscriptionKey error

weltongbi:
old:
   return $_SESSION['azure]['key']['text'];
new:
 return $_SESSION['azure]['key']['text'];


markjcrane:
If you still want to commit these changes please send a new updated pull request.

PR #4003, opened 2019-02-16
Fix Ring Group Delay Timing

emaktech:
This is going to sound really stupid but I have tested this extensively, submitted a Jira on it months ago (was told FS 1.6.20 was no longer supported) and it's still an issue. Please do not take my word for this and create a ring group to test the delay settings with a stopwatch and you should see the real vs set delay discrepancy.

In FreeSWITCH (both on 1.6.20 and 1.8.5) when sending leg_delay_start values, for whatever reason the actual time is double the value sent. The result of this is that if you send 1000ms as leg_delay_start the actual time the call will be delayed is 2000ms.

Because of this bad behavior, ring group delay settings end up being exactly double what is set. e.g. if you set 10s, you will have to wait 20s for the call to be initiated on leg b.

The easiest way to fix this behavior is to simply multiply leg_delay_start by half as much to get the right "real" delay time. Ugly, I know... I'm not sure if leg_delay_start value is passed elsewhere, I'm thinking this behavior may also be present in find me/follow me. If this gets accepted I will look for other locations where this behavior occurs and submit separate PRs if I find any other instances of this.

### DIFF
--- a/resources/install/scripts/app/ring_groups/index.lua
+++ b/resources/install/scripts/app/ring_groups/index.lua
@@ -645,7 +645,7 @@
 				--leg delay settings
 					if (ring_group_strategy == "enterprise") then
 						delay_name = "originate_delay_start";
-						destination_delay = destination_delay * 1000;
+						destination_delay = destination_delay * 500;
 					else
 						delay_name = "leg_delay_start";
 					end


markjcrane:
I just tested this on a server running on vultr.com and the delay was correct without this change. I suspect  the virtualization technology used is affecting the speed of the delay you are experiencing. Would be great if other people would test this.

emaktech:
Thanks for testing this. Interesting suggestion that it may be tied to virtualization. I'll try on other platforms and check with a few other people as well.

DigiDaz:
I doubt this at all being anything to do with virt. If the timing was out so would audio be mangled.

DigiDaz:
What virtualisation are you using?

mtghr:
Mark said that he was Testing it on vultr

emaktech:
All my testing on this has been on OpenStack cloud instances personally. I never thought to try this on bare metal but certainly will.

emaktech:
So we've moved over to dedicated servers on bare-metal now and I've repeated my testing. The delay is still broken but only with enterprise ring strategy - simultaneous is fine. Could anybody please test this on their end and see if they are seeing the same results? Would be really helpful to get feedback on this.

markjcrane:
This time you gave more details noting ring group enterprise. I will repeat my test.

emaktech:
Hi Mark,

We just went through a process of measuring this internally on a fresh master install. We tested a variety of different ring group setting combinations with simultaneous and enterprise ring strategies. We recorded everything in a spreadsheet with all the test conditions explained in the file.

I graphed the data in the spreadsheet so you can see GUI set value vs. actual result. There were some interesting trends:

1) Confirmed that enterprise strategy always seems to lead to double the set ring delay
2) Simultaneous strategy does not allow delay beyond the timeout of the previous extension's timeout.

#2 is not much of an issue as it fundamentally just causes extensions to ring immediately sequentially but it was an interesting trend we picked on that is worth noting.

Test conditions and data here:

[](https://emak.tech/wp-content/uploads/2019/03/fusionpbx-ring-group-time-vanilla.xlsx)

If you have any questions about this let me know!

PR #4004, opened 2019-02-16
Retrieve gateway names with variables and failover

emaktech:
Previously, if a bridge statement was written with prepended variables to be passed, the gateway name would not be parsed and replaced. Additionally, if multiple bridges were used in failover in a single bridge statement separated by a pipe, only the first would be parsed and replaced.

I noticed this after upgrade to FreeSWITCH 1.8.5 (from 1.6.20) which changed failover behavior with multiple separate bridge statements. When network returns 486 BUSY, it will proceed to second bridge statement, initiating a second call. To allow failover and not have this behavior, you must write a single statement and separate your failover bridges with a pipe, this change allows the UI to show all gateway names on the line once saved.

Examples:
{fax_use_ecm=on}sofia/gateway/99b1234c-667c-4b31-9d2f-31d4f396abfa/$1
=> {fax_use_ecm=on}sofia/gateway/Gateway_Name/$1

sofia/gateway/0e70e123-00fd-4f0c-b525-bc7f6283cfa1/$1|sofia/gateway/99b1234c-667c-4b31-9d2f-31d4f396abfa/1$1
=> sofia/gateway/Gateway_One/$1|sofia/gateway/Gateway_Two/1$1

### DIFF
--- a/app/dialplans/dialplan_edit.php
+++ b/app/dialplans/dialplan_edit.php
@@ -730,19 +730,30 @@
 							if ($element['hidden']) {
 								$dialplan_detail_data_mod = $dialplan_detail_data;
 								if ($dialplan_detail_type == 'bridge') {
-									// parse out gateway uuid
-									$bridge_statement = explode('/', $dialplan_detail_data);
-									if ($bridge_statement[0] == 'sofia' && $bridge_statement[1] == 'gateway' && is_uuid($bridge_statement[2])) {
-										// retrieve gateway name from db
-										$sql = "select gateway from v_gateways where gateway_uuid = '".$bridge_statement[2]."' ";
-										$prep_statement = $db->prepare(check_sql($sql));
-										$prep_statement->execute();
-										$gateways = $prep_statement->fetchAll(PDO::FETCH_NAMED);
-										if (is_array($gateways)) {
-											$gateway_name = $gateways[0]['gateway'];
-											$dialplan_detail_data_mod = str_replace($bridge_statement[2], $gateway_name, $dialplan_detail_data);
+									// split up failover bridges and get variables in statement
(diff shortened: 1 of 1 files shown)


markjcrane:
sofia/gateway/0e70e123-00fd-4f0c-b525-bc7f6283cfa1/$1|sofia/gateway/99b1234c-667c-4b31-9d2f-31d4f396abfa/1$1

I put the call in seperate bridge statements so that when it fails it goes to the next one and then shows in the call detail records which one was used. Does your solution offer anyway to see which gateway was used?

emaktech:
Hi Mark,

To be clear, I'm not changing any behavior with respect to how people write their outbound route dialplans, I'm only adding additional logic to parse all the gateway names in the UI after it's been saved here if a user chooses to write their dialplans this way.

But to answer your question, I previously was using multiple bridge statements sequentially but that broke over the weekend when I upgraded to FreeSWITCH 1.8.5 because now when 486 BUSY is returned the second statement is executing. This means if you call a busy party, the system will then call them again.

This did not used to happen in FreeSWITCH 1.6. As such, I looked at the docs and saw that FreeSWITCH now recommends we do failover like this with a pipe and this approach is working on 1.8. I don't like it as much for the same reasons you outlined but it just seems to be how FreeSWITCH is handling calls now unfortunately. This new code just makes it easier to see what you've done once you've saved your dialplan because I was having a hard time reading the UUID.

Internally for quite some time now, I actually set my own variable that I call gateway_used to make troubleshooting easier when looking in the CDRs, but that's just my own personal approach to know which dialplan executed. I'm not sure how to know which gateway truly bridged the call now that FreeSWITCH handles calls this way, it would be interesting to look into that for sure.

PR #4005, opened 2019-02-19
autocomplete="new-password" for devices

konradSC:
Chrome is no longer supporting autocomplete="off". The new standard for Chrome is autocomplete="new-password". This also seems to work for Firefox as well.


markjcrane:
This is great news! Good Job.

PR #4006, opened 2019-02-24
 Update {$mac}.cfg

greenbea:


greenbea:
I think $device_label makes more sense than $account.x.display, as yealinks has 2 settings label and display name, label corresponds to the device label and display corresponds to the account display name on the sip headers

PR #4008, opened 2019-02-28
FreeSWITCH phrases database recordings fix

jpattWPC:
The FreeSWITCH phrases module play function is not able to deal with executing a Lua script as is required when recordings are stored in the database, however FusionPBX is storing this as a play-file function in the database. This change forces the phrase_detail_function field to be execute when using database stored recordings as a lua script has to be run in order to extract the recording. After making this change the UI shows the action correctly and FreeSWITCH executes the phrase correctly.


jpattWPC:
Cancelled as this pull request has issues.

PR #4010, opened 2019-03-04
To see the number at the phone which was dialed

ednt:


markjcrane:
This was reported that is has been fixed.

PR #4040, opened 2019-03-07
GXP21XX - 1.0.9.127 Update

digitalcrisis:
These files update the provision P-codes for the newest firmware. We also add options in the device profile pages to reflect the new features provided by the new firmware.


digitalcrisis:
Error on this pull req. Sending correctly shortly.

PR #4041, opened 2019-03-07
 GXP21XX - 1.0.9.127 Update

digitalcrisis:
These files update the provision P-codes for the newest firmware. We also add options in the device profile pages to reflect the new features provided by the new firmware.

UPDATE: changed provisioning settings to original defaults provided by FusionPBX and fixed a double hyphen, which was causing a parsing issue.


fmarcoux96:
I've gone through the templates for Grandstream phones and I notice that the GXV3240 doesn't set the Outbound Proxy.

You might wanna update that too! :)

markjcrane:
Please send changes in smaller more bit sized chunks so its easier to verify the changes are working.

PR #4042, opened 2019-03-09
Fix typo in switch.conf.xml

trideeindhoven:
Minor typo in the path for the mailer.


markjcrane:
Majority of installs are Debian and they use /var/www/fusionpbx. I can't accept this change because it will break more installs than it fixes.

markjcrane:
FusionPBX official install script for CentOS uses /var/www/fusionpbx 

So I'm glad you got it working for your system and appreciate the offer of the change but I can't use the change because of the issues it will cause for almost everyone.

PR #4043, opened 2019-03-18
Update {$mac}.cfg

greenbea:
Yealink label setting is not the same as the display name setting, label, is the device label whereas the display name is the account name


markjcrane:
I don't think the change really fits. You are proposing changing it to device label which is one label for the entire device. Except its getting added to a field that is account.1.label and each account has its own label. Your proposed change will make every account have the same label. So likely the right way to do it is to add a device line label. So that there is one for every account just likely there is in Yealink's web interface.

greenbea:
So is it possible to add a device line label for each line? I was proposing this change because the device label display on yealink phones is always the first enabled line, that's usually line 1

PR #4044, opened 2019-03-25
Update orm.php

DigiDaz:
Not certain but I think the find method should be taking in $array here.


markjcrane:
The orm.php is an old php class that has been replaced and no one should be using it.

markjcrane:
No code in master branch is actually using it that I could find.

PR #4045, opened 2019-03-26
Update devices.php

afanucchi:
Remove all non-hex characters on search.
Allows searching from pastes such as:
00-00-00-00-00-00
00:00:00:00:00:00


markjcrane:
This change was not through completely. It would break searching for other things other than the mac address.

PR #4046, opened 2019-03-29
Update call_center.php

StratoSwitch:
Add pause before greeting if greeting assigned


markjcrane:
Looks good accepting your change now.

PR #4047, opened 2019-04-01
Update 860_bind_digit_action.xml

borisk1976:
In my description A - user registered with FusionPBX, B - outside user
Original bind_digit_action works properly only in one situation - when A calls B. When B calls A after the A pressed *4 (for example) A hears MOH!!! and B gets tone.

What I was trying to achive - we're checking where call is coming from and set bind_target accordingly.

What my patch does not solve but we need find a solution: A and B are both local registered users and one is calling another. In traditional PBX no matter A or B are doing a call. Each may press transfer key and transfer call to C. With my patch only A can do that and I have no idea how to permit B too.


markjcrane:
This will take a good amount of testing to make sure it doesn't create a security risk.

borisk1976:
Of course. But this should be fixed indeed. Current bind_digit_action doesn't work as expected. And unfortunatelly my freeswitch knowledge isn't enought to completely solve this problem.

markjcrane:
This pull request is old if you would like to see the changes get added then update the changes with latest code and send a new  pull request.

PR #4048, opened 2019-04-11
Paging Controls Fix

konradSC:
Bug where you couldn't enter a page number and press ENTER to jump to that page.


markjcrane:
Looks good.

PR #4050, opened 2019-04-17
disable email checks from disabled domains

ednt:
To avoid checks of, maybe broken, email accounts from older disabled domains.
The check tooks so long, that the db connection was closed and the rest domains get wrong values.
I hope that all databases accept f.*
We use postgresql and it works.


markjcrane:
This pull request is old if you would like to see the changes get added then update the changes with latest code and send a new pull request.

PR #4058, opened 2019-04-25
Update fax_files.php

mtghr:
i did this in order to show the correct fax time on the page based on the TImezone set in the default settings 

i have notices that someone post it a different patch for the master branch i am not sure which one is a better way of doing this

### DIFF
--- a/app/fax/fax_files.php
+++ b/app/fax/fax_files.php
@@ -199,8 +199,16 @@
 	list($paging_controls, $rows_per_page, $var3) = paging($num_rows, $param, $rows_per_page);
 	$offset = $rows_per_page * $page;
 
+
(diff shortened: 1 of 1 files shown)


markjcrane:
Please sign up as a contributor here https://github.com/fusionpbx/open-source

markjcrane:
Contributor agreement helps provide some level of safety for the project
I will only accept changes from contributors that agree to it. This will be closed by 1st of June if not accepted. BSD went into legal battles with AT&T because someone added Unix code to BSD. The contributor agreement has a clause to have contributors agree to only contributing code that is theirs to contribute. There are other provisions designed to help protect users and contributors. This is why its required.

PR #4061, opened 2019-04-28
Update {$mac}.cfg

mtghr:
Polycom will not support wild  card cert only if you disable strictCertCommonNameValidation

also polycom is missing many root  certs so we need an option to push them via the template 

bot stuff was tested by me in  production and its working fine


markjcrane:
Please sign up as a contributor here https://github.com/fusionpbx/open-source

markjcrane:
Contributor agreement helps provide some level of safety for the project
I will only accept changes from contributors that agree to it. This will be closed by 1st of June if not accepted.

mlaradji:
Hi @mtghr. I am facing the same issue and I believe this will fix it. What is `$polycom_root_cert` in this case?

As a note, I usually "fix" this problem by updating the firmware on the device, which got rid of my certificate issue.

markjcrane:
I added an isset so that the polycom_root_cert  settings are added if its provided.

mtghr:
Thank you Mark I missed the isset

PR #4067, opened 2019-05-01
Add another working way to use Azure Speech to Text service

daniel-lucio:
This patch adds BING (legacy) some extra fault tolerance support. For those who still use Bing, Bing sometimes fails
It adds another way, which I think it is easier and faster (Azure). Among the outstanding things it has; just configure the zone, and use Memcache to store the access token as per Azure documentation, it saves a query (faster). It also returns the most likely right transcription (digging in the JSON)


daniel-lucio:
???

PR #4097, opened 2019-05-08
[4.5] Fix the ALTER TABLE

daniel-lucio:
when $apps[$x]['db'][$y]['fields'][$z]['name'] is an array, you dont try to get the ['text'] index therefore fields such as v_xml_cdr.xml_uuid, v_destinations.destination_type and v_destinations.destination_number wont be created when upgrading.

This patch fixes it


markjcrane:
Reverting this change because it causes a big problem. It adds fields right before they are supposed to be renamed. So then you can't rename the field because another field with the same name already exists.

daniel-lucio:
I understand it needs to check if it is not renaming a field before adding a new one. mmmm but then I guess there is a problem in the renaming part of the code.

PR #4100, opened 2019-05-08
[4.4] Same as 4.5

daniel-lucio:
I think you should declare 4.4.6 as the upgrading is a critical thing


markjcrane:
Reverting this change because it causes a big problem. It adds fields right before they are supposed to be renamed. So then you can't rename the field because another field with the same name already exists.

PR #4166, opened 2019-05-20
added the ability to notify users their vm_quota

greenbea:
I added the ability when a user gets their voicemail to email it should also notify him how much space he still has in his VM box

need to add in default settings category voicemail, subcategory send_quota, boolean true

and an email template category voicemail, subcategory voicemail_with_quota. It should look like something like  this:
<html>
<body>
Voicemail from ${caller_id_name} <a href="tel:${caller_id_number}">${caller_id_number}</a><br />
<br />
To ${voicemail_name_formatted}<br />
Received ${message_date}<br />
Length ${message_duration}<br />
Message ${message}<br />
<hr />
<b>Your Voicemail box has ${remaining_space} of remaining spacel</b>
</body>
</html>

### DIFF
--- a/resources/install/scripts/app/voicemail/index.lua
+++ b/resources/install/scripts/app/voicemail/index.lua
@@ -211,6 +211,10 @@
 						not_found_message = settings['voicemail']['not_found_message']['boolean'];
 					end
 				end
+					
+				if (settings['voicemail']['send_quota']) then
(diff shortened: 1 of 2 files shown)


greenbea:
I've added the ability to notify user the amount of new and saved messages in his voicemail box

PR #4168, opened 2019-05-22
Update contacts.php default sort order

AdrianFretwell:
Added a default sort order that can be set by the session, if unset defaults to last_mod_date descending.  Added a secondary sort on contact_organization ascending.


markjcrane:
Looks like a goo improvement I'll review it a bit more before accepting it.

PR #4170, opened 2019-05-22
Updates for Operator Panel

danbgds:


markjcrane:
This code does not maintain the capabilities of the original code. Introduces some regressions in feature. Primarily loose a major feature of changing the agent status. Tightens security in one area and then introduces risk in another area. As it stands I can't loose the features you commented out.

danbgds:
Thanks for taking a look at it.  Can you clarify what you mean by "introduces risk in another area"?

All of the commented code appears to be unused/unreachable through normal use, because the front end code never uses the "action" parameter.  For example, changing the agent status still works with these changes because that is handled by code in index.php rather than exec.php, with a request like /app/operator_panel/index.php?status=Do%20Not%Disturb.  The code that makes the request is here: https://github.com/fusionpbx/fusionpbx/blob/cea3e5cd24d9c6ecb9a25f7c874d5193d0fcf5a3/app/operator_panel/index_inc.php#L80-L102 and the backend code that processes it is here: https://github.com/fusionpbx/fusionpbx/blob/cea3e5cd24d9c6ecb9a25f7c874d5193d0fcf5a3/app/operator_panel/index.php#L46-L56

Is that incorrect?

markjcrane:
You got me to look at the entire operator panel I'm rewriting a portion of it at this moment not sure if this file is going to survive the rewrite.

markjcrane:
I’ll review it some more.

markjcrane:
Sorry my changes made the code conflict with your changes.

danbgds:
If you're open to these changes, we can resubmit them against basic_operator_panel.  Even if you're planning on refactoring the module in the future, it'd be beneficial to lock this stuff down in the meantime.

markjcrane:
Sure sorry about causing the conflict. I do see that your changes are
needed.

markjcrane:
Thanks for correcting the conflict changes have been accepted.

PR #4175, opened 2019-05-24
Added the ability to cut outbound prefixes from the CDR

greenbea:


greenbea:
When a user creates an outbound route it will add to the outbound dialplan set prefix=the_prefix
when displaying the CDR in xml_cdr.php it checks in the database for the  for the prefix variable and cuts it from the destination_number

PR #4229, opened 2019-05-31
Added support for Digium D-Series phones.

bryanredeagle:
You can configure accounts, speed dials, and BLF keys. Downside is that the BLF keys do not work. The phone will subscribe to them correctly and receive all the right NOTIFYs, but the phone doesn't do anything with them. I'm sure it's some Asterisk-specific format that Freeswitch doesn't use.


markjcrane:
Thanks for contributing back to the project.

PR #4268, opened 2019-06-06
Security Fixes

blackwolfresearch:
Fixes for CVE-2018-19094 and CVE-2018-19096


markjcrane:
This is a messy pull request that includes an already accepted change. Please send a clean pull request. That only shows the latest changes.

markjcrane:
I'm not sure if this message was saved so writing it again. Company called up and inbound fax came in and the email was not coming in. Commented out this command and inbound fax to email worked again.
cmd=shell_esc(cmd);
You might be on the right track protecting the code better but as is it broke it bad enough that at the moment it had to be commented out. Maybe it should be used around the variables instead that are used in the command.

markjcrane:
I worked on this more and made it use the shell_esc function the correct way and updated both 4.4 and master branch.

PR #4270, opened 2019-06-07
Update destination_edit.php

greenbea:


markjcrane:
Keeping this around as a reminder to work on it. I don't think simply commenting it out is the right solution.

greenbea:
What's the purpose of this line?

markjcrane:
He wants to customize the destination context

greenbea:
isn't the same thing done on line 152 but without forcing the domain name to be the context name?

PR #4275, opened 2019-06-10
Fix ocurrences of https://fusionpbx.com

mat1010:
Since the websites SSL / TLS certificate does not include `fusionpbx.com` as a SAN the https links should point directly to `www.fusionpbx.com` since otherwise the browser will throw an error regarding an insecure connection. An alternative fix would be to get the non-www name into the SSL / TLS certificate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 What is [FusionPBX](https://www.fusionpbx.com/)?
 --------------------------------------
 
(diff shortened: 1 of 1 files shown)


markjcrane:
Please sign up as a contributor and I will accept your changes immediately.

Its a good change thanks for sharing it.

markjcrane:
Great appreciate the helps accepting your changes.

markjcrane:
I'll also work on adding fusionpbx.com in the SSL along with the www.fusionpbx.com.

PR #4276, opened 2019-06-11
Update dialplans.php

chansizzle:


daniel-lucio:
I don't think this is a good idea.

markjcrane:
Agreed I haven't committed yet because I don't think it shows a full understanding of what is being changed.

PR #4280, opened 2019-06-15
Update app_config.php

Len-PGH:


markjcrane:
user_password is the variable for this setting so closing this pull request.

PR #4288, opened 2019-06-18
Add the description element for import

ednt:
To make it possible to use descriptions for own dialplan xmls


markjcrane:
A good improvement accepting the change.

PR #4303, opened 2019-06-22
Ability to set permission required to execute save and delete methods.

reliberate:
Works as expected on Email app, but please review changes.


markjcrane:
The change makes it accept any string that is sent as a permission to the database method. To me this is almost equivalent to just removing permission checks all together or similar to having a switch that tells it to disable the check. I do fully understand the concept. Maybe there is another way to accomplish the other benefits you want to get from this it would be helpful to discuss further over the phone.

PR #4311, opened 2019-06-26
Fix custom variables in the CDR importer

daniel-lucio:
If you specify a custom column in $_SESSION['cdr']['field'], and the :;array variable is set somewhere else (class heritance?), the variable doesn't exist in the JSON, but it is already specified in the ::field and in the ::array variables.

This patch prevents the reset of those variables if they already exist.


markjcrane:
Array was designed to be overridden from the beginning so it is not a flaw but part of the plan.

daniel-lucio:
You are wrong in that design, you need to verify if the element already exists. Your code just add, if someone put there something like destination_number, pdd (or any other repeated field), the SQL will create an insert query listing twice the filed: INSERT INTO xml_cdr(field1, field1, field2, field3....)

Postgresql and MariaDB, the both, complain about having a field listed twice. The insert wont be do it.

PR #4312, opened 2019-06-26
Same as 4.4

daniel-lucio:
Same rationale


markjcrane:
Array was designed to be overridden from the beginning so it is not a flaw but part of the plan.

PR #4315, opened 2019-06-28
Update {$mac}.cfg

mtghr:
in order for poly come to work on TLS on multi tenant we need 2 steps 

1. we need o push down the root cert ( for many certs since polycom i find to have much less then other vendors )
2. we need to disable strict domain validation


markjcrane:
Closing pull requests with merge issue. If you want to add change please update the changes to the new code base and submit a new pull request.

PR #4323, opened 2019-06-30
{$mac}.xml

GetAGripOnIt:
Fix empty password bug & other updates & additions
Full path: /var/www/fusionpbx/resources/templates/provision/grandstream/gxp17xx/{$mac}.xml


markjcrane:
This template is following older 4.4 branch syntax.
Master branch uses a provisioning syntax like {$account.1.display_name}

markjcrane:
I see there are parts that are definitely an improvement but needs to have updated syntax.

markjcrane:

PR #4337, opened 2019-07-07
fixes _SESSION[user][contact_uuid] emtpy

daniel-lucio:
Oddly, you fix this in 4.5 but not in 4.4.


markjcrane:
Not that odd just the developer that committed it didn't commit it to 4.4.

PR #4362, opened 2019-07-18
Little Optimization

daniel-lucio:
I was doing the following test, which is similar to the authentication methods array.
<?php
$a['authentication']['something']['text'] = 'abc';
if (!is_array($a['authentication']['methods'])) {
        print 'no array'.PHP_EOL;
        $a['authentication']['methods'][] = 'database';
}
if (!isset($a['authentication']['methods'])) {
        print 'not set'.PHP_EOL;
        $a['authentication']['methods'][] = 'database';
}

print_r ($a);

The output was:
PHP Notice:  Undefined index: methods in /tmp/t.php on line 7
no array
Array
(
    [authentication] => Array
        (
            [something] => Array
                (
                    [text] => abc
                )

            [methods] => Array
                (
                    [0] => database
                )

        )

)

Therefore, no need to verify twice, a simple OR should do the same.
Also this patch protects against wront default settings, lets say someone put $_SESSION['authentication']['methods']['boolean'] = true

### DIFF
--- a/core/authentication/resources/classes/authentication.php
+++ b/core/authentication/resources/classes/authentication.php
@@ -44,18 +44,28 @@ class authentication {
 	public function validate() {
 
 		//set the default authentication method to the database
-			if (!is_array($_SESSION['authentication']['methods'])) {
+			if (!is_array($_SESSION['authentication']['methods']) || !isset($_SESSION['authentication']['methods'])) {
 				$_SESSION['authentication']['methods'][]  = 'database';	
+			}
+			else {
+				// This protects against $_SESSION['authentication']['methods']['text'] and any simillar which is wrong set.
+				// It must be an array.
+				$do_fix = false;
+				foreach ($_SESSION['authentication']['methods'] as $i){
+					if (is_array($i)){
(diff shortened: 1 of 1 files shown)


markjcrane:
I don't see any reason why this is being done says the same thing twice.
if (!is_array($_SESSION['authentication']['methods']) || !isset($_SESSION['authentication']['methods'])) {

daniel-lucio:
If you see the code before this patch, you had it split in 2 IF's. I just joined and add more support to other cases.

is_array() and isset() evaluate different things, the variable could be set without being an array. But more important than that is the else{} part which prevents to put this default setting to another thing different than type=array

daniel-lucio:
If not clear, text me I can do a depth explanation on Skype

markjcrane:
I've been thinking about this for weeks off and on. It looks to me like it adds code complexity with very little benefit. Maybe more relevant to code you haven't shared. Next thing about it I really don't like $do_fix as a variable name. The name says fix but its still not intuitive. Next I think its better to fix the data as its going into the database rather than when its coming out.

PR #4363, opened 2019-07-18
Same as 4.5

daniel-lucio:
Same


markjcrane:
I've been thinking about this for weeks off and on. It looks to me like it adds code complexity with very little benefit. Maybe more relevant to code you haven't shared next I really don't like $do_fix as a variable name. Variable names are best when they are intuitive.

PR #4388, opened 2019-07-25
New variables added for Aastra provisioning

saraxxin:
Applied updates for new variables to Aastra phones.
(Added in wrong directory first hence it thinks it was added and deleted once.)


markjcrane:
Looks good.

PR #4395, opened 2019-07-29
Update {$mac}.xml

benwiesner1:
working / updated cisco spa112 template


markjcrane:
Please signup as a contributor here and then I will accept the spa112 template changes.

PR #4396, opened 2019-07-30
Duplicate uuid

ednt:
The duplicate uuid of ring_text_5 and _6 produces a lot of errors in postgres log file.
Also ring_text_6 is not available in the settings.


markjcrane:
Thanks for bringing this to our attention on 4.4. However apparently this has been fixed already on master branch with a new UUID. I will try to update your pull request with the UUID used on the master branch. If I go with your UUID and master branch next release uses a different UUID then that variable would be duplicated.

markjcrane:
Updated your pull request with the correct UUID and now can accept it.

PR #4403, opened 2019-07-31
Update and rename 860_bind_digit_action.xml to 045_bind_digit_action.xml and fix the dialplan

greenbea:
This pull request is a fix to the ticket I've sent in regarding bind digits dialplan problem

1 - ${sip_athorizesd} always evaluate to true even when coming from the outside  (probably because of the domains acl) so bind_target is always set to both

2 - it's only working for inbound calls going to an extension but not working for outbound calls


markjcrane:
This was disabled more than two years ago because of security concerns. Will review your changes especially in regards to security.

greenbea:
changed to improve security

PR #4414, opened 2019-08-03
Disable azure transcription memcache by default

emaktech:
In FusionPBX Master memcache is not used by default so the default behavior to use memcache for Azure voicemail transcription causes it to fail due to invalid access token being fetched.

This disables caching by default.

Should probably wait for @daniel-lucio to respond before merging as he is the original author of this feature and may have a better idea.

### DIFF
--- a/app/access_controls/app_config.php
+++ b/app/access_controls/app_config.php
@@ -55,6 +55,9 @@
 		$apps[$x]['permissions'][$y]['groups'][] = "superadmin";
 		$y++;
 
+	//cache details
+		$apps[$x]['cache']['key'] = "configuration:acl.conf";
+
 	//schema details
 		$y=0;
 		$apps[$x]['db'][$y]['table']['name'] = "v_access_controls";
--- a/app/access_controls/app_config.php
+++ b/app/access_controls/app_config.php
@@ -56,7 +56,7 @@
 		$y++;
 
 	//cache details
-		$apps[$x]['cache']['key'] = "configuration:acl.conf";
+		$apps[$x]['cache']['key'] = "configuration.acl.conf";
 
 	//schema details
 		$y=0;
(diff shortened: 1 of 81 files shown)


daniel-lucio:
OK, so let me understand. Fusion 4.5 (devel) doesn't use Memcache as default caching engine, instead, it uses a filesystem one.

The problem here is that because the filesystem cache doesn't expire, it keeps giving you the same key always. Am I right?

emaktech:
Yes, I'm not 100% sure but I think that's what's happening.

daniel-lucio:
Thinking....

markjcrane:
Something like this should work. cronjob to clear file cache when the file reaches certain age.
find /vac/cache/fusionpbx/*  -name '*part-of-cache-file-name' -mmin +5 -exec rm {} \;

daniel-lucio:
Yes, but that's not elegant, and it makes confusion for newbies. I am thinking to put some old code that asks for the key, like:

authkey =  trim(api:execute("memcache", "get " .. key));
if (XML_STRING == "-ERR NOT FOUND") or (XML_STRING == "-ERR CONNECTION FAILURE") then
-- do it manually
end

so, it uses memcached for this always. Some freeswitch modules use memcached. If memcache if off the CONNECTION FAILURE error will get it and all will be working okay.

daniel-lucio:
Sent #4415

markjcrane:
The voicemail application is database driven so I'm doubting whether caching this is of much use.

daniel-lucio:
It doesn't matter if it is file driven, the caching saves HTTP queries to azure. Azure requires 2 HTTP queries, first to get an auth token (this is the cacheable one), and the service query (which uses the auth token).

by caching the auth token you save resources, mainly on systems with a lot of VM's

PR #4415, opened 2019-08-03
Dont use the FS cache on Fusion 4.5+

daniel-lucio:
As explained, the filesystem cache doesn't expire. This doesn't use the LUA cache object and instead goes for memcache directly.

### DIFF
--- a/resources/install/scripts/app/voicemail/resources/functions/record_message.lua
+++ b/resources/install/scripts/app/voicemail/resources/functions/record_message.lua
@@ -140,13 +140,10 @@
 			                -- search in memcache first, azure documentation claims that the access token is valid for 10 minutes
 			                local cache = require "resources.functions.cache";
 			                local key = "app:voicemail:azure:access_token";
-			                local access_token_result = cache.get(key)
+					-- in this only case, only memcached
+			                local access_token_result = trim(api:execute("memcache", "get " .. key));
+					if access_token_result:sub(1, 4) == '-ERR' then
 
-			                if access_token_result then
-			                        if (debug["info"]) then
-			                                freeswitch.consoleLog("notice", "[voicemail] Azure access_token recovered from memcached\n");
-			                        end
-			                else
 			                        access_token_cmd = "curl -X POST \"https://"..api_server_region.."api.cognitive.microsoft.com/sts/v1.0/issueToken\" -H \"Content-type: application/x-www-form-urlencoded\" -H \"Content-Length: 0\" -H \"Ocp-Apim-Subscription-Key: "..api_key1.."\"";
 			                        local handle = io.popen(access_token_cmd);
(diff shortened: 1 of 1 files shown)


markjcrane:
The voicemail application is database driven so I'm doubting whether caching this is of much use.

daniel-lucio:
It doesn't matter if it is file driven, the caching saves HTTP queries to azure. Azure requires 2 HTTP queries, first to get an auth token (this is the cacheable one), and the service query (which uses the auth token).

by caching the auth token you save resources, mainly on systems with a lot of VM's

emaktech:
Agree it definitely makes sense to reduce HTTP queries if possible on a busy system.

Not sure I like needing to have memcache just for this though, it won't be hit so often that it really matters. Maybe instead of using memcache, simply save to a local file and every time the script runs check the age of the file. If the file is older than set time, get a new key if it's newer then simply read the value from the file and move on.

The location could use the existing FusionPBX default setting value stored in:
category: cache
subcategory: location

In this way I suppose it would be agnostic on the OS and would not require any additional modules to work.

daniel-lucio:
Memcache is installed by default in the fusion installer, and it is used by other modules at freeswitch level. you were using it without even know it

daniel-lucio:
On the other hand, as you say, the "expiring" capability is needed. This is a missing part of the FS caching, should be implemented in the cache object.

emaktech:
@daniel-lucio I tested this PR and I'm unable to leave voicemails with this new code in place (call disconnects immediately when transferred to the VM box). I haven't done a lot of testing yet but I will try to investigate further why this is happening from this submission.

The other thing I noticed is that when using Azure transcription engine with files saved as mp3 (vm_message_ext == "mp3") this is ignored and the file gets saved and emailed as WAV only. This is a problem for users that choose to use mp3 using this variable, so it will also be necessary to add in some additional logic to send WAV to the API and then save/email as mp3.

I will try to take a deeper look at this over the next few days to hopefully put something together that's ready for production with more testing and is compatible with all FusionPBX settings.

markjcrane:
I'm not switching any code to be forced to memcache by default. If someone wants to use memcache and not file cache it is possible on latest version by changing a setting.

daniel-lucio:
I agree with the logic, must use the cache object. However, your filesystem cache is way broken, objects do not expire. You need to fix than by comparing file modification against the expire (lua) array because https://github.com/fusionpbx/fusionpbx/blob/master/resources/install/scripts/resources/functions/cache.lua (line 145) does check for expiration.

markjcrane:
Way broken is your opinion. I have learned over years we don’t agree on just about anything. Database, operating system, or web server. So I expect we will disagree on other things too. Cache works very good for my needs and way I use it. I’m not going to try and convince you otherwise just use memcache if you want to... case closed.

PR #4424, opened 2019-08-07
A fax always used the first gatway which was found regardless of the …

ednt:
…conditions.

When more than one outbound rule is available, the fax uses always the first which was found.
But this was the wrong one which was clear when we look at the conditions.
So we have to respect the conditions to find the right gateway.
I already 'pulled' this 2018 for 4.2. Now for the master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,23 +32,30 @@ How to Install FusionPBX
 ----------------------------
 * As root do the following:
 
-```bash
-apt-get update && apt-get upgrade && apt-get install -y git
+Debian Install
 ```
-```bash
-cd /usr/src
+wget -O - https://raw.githubusercontent.com/fusionpbx/fusionpbx-install.sh/master/debian/pre-install.sh | sh;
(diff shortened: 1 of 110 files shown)


markjcrane:
That is why we have had the fax prefix. So you create an outbound route for the fax server put a prefix on the server so it picks the right outbound route using the prefix.  A Simple solution...

ednt:
Wrong. We needed a prefix, because the right gateway was not found.
And also in click_to_call, because this uses also outbound_route_to_bridge().

markjcrane:
I added that prefix I know why it was added. There is more than one reason for it being added. Your message fails to explain your commit. I won't add the code unless I see why it should be there. 

Your description states we have to respect the conditions which is impossible to do perfectly because there are numerous conditions that can use values that are only accessible during the call. Outbound route to bridge is run before the call so it does not have access to these values that are only available during the call.

This pull request is waiting here because I need to do further testing to get a better idea of what the changes are doing. A better explanation of your changes could help otherwise it will wait until I have more time to look at it.

PR #4432, opened 2019-08-11
Update user_dashboard.php

benwiesner1:
fix user dash welcome: link

### DIFF
--- a/core/user_settings/user_dashboard.php
+++ b/core/user_settings/user_dashboard.php
@@ -2,31 +2,24 @@
 /*
 	FusionPBX
 	Version: MPL 1.1
-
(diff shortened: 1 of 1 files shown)


benwiesner1:
this issue also exsist under Menu Home>Accounts but i cant fig out how to fix that one.

benwiesner1:
if you notice when you select the currently logged in user at the top of dash it simply just opens window for new account rather then the current users account setttings. this behavoir is also happening if you select account settings under menu item accounts.

benwiesner1:
from what i can tell its just not loading the current user/session UUID

benwiesner1:
committing simpler version as //get user was not needed.

PR #4433, opened 2019-08-11
Update app_menu.php

benwiesner1:
fixed user account setting edit under Home>Account settings


benwiesner1:
this is the same fix as the user dashboard pull request that was commited

benwiesner1:
any update on getting this pushed out? customers getting annoyed

PR #4434, opened 2019-08-11
Update app_menu.php

benwiesner1:
seems more logical to move this menu item to the Status category


benwiesner1:
yea i understand that just kind made sense to me to be under status for super admins as its a tool to check status / monitor this service

PR #4436, opened 2019-08-11
Update user_dashboard.php

benwiesner1:
fixed dashboard user account edit link at top right hand corner next to welcome.


benwiesner1:
I also submitted the same fix for the main menu

PR #4437, opened 2019-08-13
Fix Grandstream error

powerpbx:
Fixes smarty template engine error when using Granstream gxp2140.  This may also affect other templates.  To reproduce, assign provisioned extension device as gxp2140 and try save.  We only tested this on v4.4 branch. We are assuming the problem still exists and the fix still works on master.


markjcrane:
Closing pull requests that Shows 3903+ changes.

PR #4453, opened 2019-08-23
Update destinations.php

chansizzle:


benwiesner1:
can this same code work for the ext show all issue with registered ext not showing?

PR #4455, opened 2019-08-27
Not all DB support COUNT FILTER

daniel-lucio:
Since Freeswitch has opened their support to MariaDB, this worths the pain. This code builds an equivalent code when DB type is not psql.

Maybe later, would be better to move from FILTER to SUM as I have tested the SUM queries on PostgreSQL and they work without issue.


markjcrane:
We are thoughtful about the code we write and that we accept in order to make it easier to maintain the code base and to improve the readability.

These two projects removed MySQL their reasons are interesting and we have experienced similar limitations that are mentioned in the links below.

- https://about.gitlab.com/2019/06/27/removing-mysql-support/
- https://www.openproject.org/deprecating-mysql-support/

daniel-lucio:
None of these reasons are valid, we use MariaDB (MySQL is just a standard way of calling the protocol). That TEXT restriction doesn't exist in MariaDB. The second article doesn't back up their decision.

FreeSWITCH 1.10.x has now mod_mariadb, which shows how they are adopting the technology.

markjcrane:
This pull request is out dated and cannot be merged. Closing pull request if still interested in adding changes update them and submit a new pull request.

PR #4456, opened 2019-08-27
Same as Devel [4.5]

daniel-lucio:


markjcrane:
If you want to support MySQL probably be better to maintain two separate SQL queries. Rather than if statements mixed in with the existing SQL query. At least that would keep it simple and easier to follow.

PR #4483, opened 2019-09-03
Update dialplan_xml.php

chansizzle:


reliberate:
As you probably discovered, the app_uuid is used as it's passed when clicking on Inbound Routes and/or Outbound Routes in the menu (acts as a filter for the dialplan entries). If your PR was accepted, it would fix the Back button when visiting through Dialplans in the menu, but would break the Back button when visiting through Inbound or Outbound Routes.

That said, the issue you experienced with the Back button should be resolved with the recent commit.  Thanks.

PR #4511, opened 2019-09-06
Update app_config.php

benwiesner1:
Polycom syslog settings


benwiesner1:
re-commiting with some other changes

PR #4512, opened 2019-09-06
Update {$mac}.cfg

benwiesner1:
Polycom syslog settings


benwiesner1:
re-commiting correct code

PR #4518, opened 2019-09-06
Update app_config.php

benwiesner1:
added - syslog, updated descriptions
removed bluetooth radio as this is only for vvx 600 and can be set as per device


benwiesner1:
rewriting this pull
close

PR #4526, opened 2019-09-09
Fix delay for external follow-me calls

konradSC:
This is related to https://github.com/fusionpbx/fusionpbx/commit/9dcaddd81422bb46d940a2a3105112b30debcafa#diff-b1f5588538149bd825603176ff81d714.

For internal calls the delay needs to be "Delay In Seconds * 500".
For external calls the delay needs to be "Delay In Seconds * 10000". 

For external calls I'm am just doubling the value set prior in the script.

### DIFF
--- a/resources/install/scripts/app/ring_groups/index.lua
+++ b/resources/install/scripts/app/ring_groups/index.lua
@@ -753,6 +753,9 @@
 						dial_string = "[sip_invite_domain="..domain_name..",call_direction="..call_direction..","..group_confirm.."leg_timeout="..destination_timeout..","..delay_name.."="..destination_delay.."]" .. row.destination_number;
 					else
 						--external number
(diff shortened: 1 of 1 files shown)


yois615:
@konradSC - What bug was this intended to fix?
I'm using 1.10.7 and this line is causing incorrect behavior by doubling the delay time.

greenbea:
There used to have a bug that caused the delay to be half of the time https://github.com/fusionpbx/fusionpbx/blob/master/app/scripts/resources/scripts/app/follow_me/index.lua#L394

If this is fixed on 1.10.7 we have to adjust the delay only for older Freeswitch versions

yois615:
Do you have a reference for the FS bug so we can know if it's fixed?
I don't have any systems to refer back to, all I know is what's happening now.  I commented out this line in both ring_group and follow_me and now it's working (for me) as expected.

PR #4528, opened 2019-09-09
Calculate timeout for Follow-me

konradSC:
Need to calculate the timeout for Enterprise RG members that have extensions with follow-me. 

The RG timeout should always take precedence over a follow-me timeout value. What we do is take the delay of the follow-me destination and subtract that from the ring group timeout to give us the total timeout of the destination. 

Example: 
RG 1: x1000 (Delay=0, Timeout=10)
RG 2: x2000 (Delay=10, Timeout=10)

x2000 has follow-me enabled
FM 1: x2000 (Delay=0, Timeout=15)
FM 2: x3000 (Delay=5, Timeout=20)

In this example we would want x2000 ring for 10 seconds and x3000 to ring for 5 seconds. 

What if we changed this... FM 2: x3000 (Delay 15, Timeout=20)

In this example we wouldn't want x3000 to ring at all because it would start to ring after the RG timeout has expired. Our calculated value would be a negative value, -5. These negative values don't work as leg_timeouts in the dialstring, so we need to test for them.

### DIFF
--- a/resources/install/scripts/app/ring_groups/index.lua
+++ b/resources/install/scripts/app/ring_groups/index.lua
@@ -561,7 +561,11 @@
 								destinations[new_key]['domain_name'] = field.domain_name;
 								destinations[new_key]['destination_number'] = field.destination_number;
 								destinations[new_key]['destination_delay'] = field.destination_delay + row.destination_delay;
-								destinations[new_key]['destination_timeout'] = field.destination_timeout;
+								if (tonumber(field.destination_timeout) < (tonumber(row.destination_timeout) - tonumber(field.destination_delay))) then
+									destinations[new_key]['destination_timeout'] = field.destination_timeout;
+								else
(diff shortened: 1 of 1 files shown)


konradSC:
Lines 564-568 are the key. 

Line 620 is an "If" statement to check for negative values. 

The rest are just indentations.

konradSC:
Description has been added to the code.

FozzyGo4:
Hi Mark,
Quick correction on index.lua.
Line 122 should be v_follow_me table not v_extensions
sql = sql .. "from v_extensions ";
Line 125 param should be follow_me_uuid = follow_me_uuid and not destination_number = destination_number
local params = {domain_uuid = domain_uuid,destination_number = destination_number};
Line 346 is wrong syntax for the variable = 'destination_uuid' should be = :destination_uuid
sql = sql .. "and destination_uuid = 'destination_uuid' ";

PR #4563, opened 2019-09-11
Update do_not_disturb.lua

greenbea:
The SQL query was broken it always set the dial_string to error/user_busy should be null when disabling DND


greenbea:
Mark was faster than me in fixing this

PR #4564, opened 2019-09-11
Fix follow me caller ID query and variable 

FozzyGo4:
I found the wrong table was used in the query to get caller ID from follow me table as well as the incorrect variable used.

Also incorrect syntax used for the set called ID query further down.


FozzyGo4:
Sorry for the double commits. This is my first time.

markjcrane:
Pull requests should be one topic that is changed per pull request. This pull request has multiple different things. Not just fixing caller id in follow me.

markjcrane:
For example adding South Africa to outbound routes select list should be its own pull request.

FozzyGo4:
So sorry markjcrane. I am brand new at Githup and not to sure yet of how it all works.

Should I perhaps delete these commits and do them separately? Not sure why it added all the above commits to the fix proposal commit.

FozzyGo4:
Closing and resubmit clean commit.

PR #4600, opened 2019-09-13
Re-enable follow me if follow me exists for DND

FozzyGo4:
Follow me is disabled after DND is enabled but does not re-enable follow me if DND is disabled.
Only applies if follow me exists for the user.


markjcrane:
Problem here again is an assumption that may not be true. What proof do you have that follow me should be reactivated? What happens if someone disabled it but didn't remove the destinations and this code automatically re-enable it.

FozzyGo4:
Again agree with you 100%. Think the idea of my comments on the call forward commit will give the proof that that follow me was previously enabled or not.

FozzyGo4:
Closing this and resubmitting new code

PR #4602, opened 2019-09-13
Update call forward to re-enable follow me

FozzyGo4:
If call forward is enabled then it disables follow me. When CF disabled it should re-enable follow me if exists.

### DIFF
--- a/resources/install/scripts/call_forward.lua
+++ b/resources/install/scripts/call_forward.lua
@@ -252,13 +252,33 @@
 			session:streamFile(sounds_dir.."/"..default_language.."/"..default_dialect.."/"..default_voice.."/ivr/ivr-call_forwarding_has_been_cancelled.wav");
 	end
 
---disable the follow me
+--toggle the follow me
 	if enabled == "true" and not empty(follow_me_uuid) then
 		local sql = "update v_follow_me set ";
 		sql = sql .. "follow_me_enabled = 'false' ";
 		sql = sql .. "where domain_uuid = :domain_uuid ";
 		sql = sql .. "and follow_me_uuid = :follow_me_uuid ";
 		local params = {domain_uuid = domain_uuid, follow_me_uuid = follow_me_uuid};
+		local sql = "update v_extensions ";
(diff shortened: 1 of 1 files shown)


markjcrane:
This is assuming that follow me was disabled by call forward. Without something that really indicates it was then it will introduce its own bug when call forward was not the thing that disabled follow me.

FozzyGo4:
I agree 100%. The current code only disables follow me (FM) if a FM UUID exists. To re-enable after call forward (CF) is enabled may not be the desired effect if the user does not want FM enabled or had it disabled or never set it up in the first place. This would cause issues. Perhaps there should be a second variable like a "what the current state of FM was at the time" when CF was enabled. Will try and explain a scenario quick of how I think it could work and please correct me if I'm wrong.

A user has FM enabled by default with some destinations. The user then decides to have calls forwarded to a mobile for the weekend. User dials *73, enters number, FM is disabled and CF is now enabled. After the weekend then user disables CF and because they originally had FM enabled before, it then gets re-enabled. If the user had FM disabled before hand then it should do nothing to FM.

Maybe this could be achieved with a double "follow_me_enabled" field in the DB like "follow_me_enabled" and "follow_me_default_enabled". If a user enables FM from the UI or from feature code *21 then both fields will be set to true. If disabled or never setup then both fields are set to false. If the user enables call forward, it then disables the FM (follow_me_enabled = false) field but leaves (follow_me__default_enabled = true) set as it was. In the LUA script the code checks `if (follow_me_caller_id_uuid ~= nil AND follow_me__default_enabled == 'true')` then re-enable follow me. If not, do nothing.

What do think?

I personally feel this is important as currently if the user has FM setup and enables DND or CF then if they don't remember to enable FM again manually, it will remain disabled and the use thinks the system is not working as it should. I have had customers call me regularly in this regard.

Hope this makes sense.

FozzyGo4:
I have now put in to practice what I have just thought of and seems to be working as desired. I will close this commit and re-submit with the new changes.

PR #4603, opened 2019-09-13
Update {$mac}.cfg

benwiesner1:
this should satisfy exsisting using polycom_provision_url and allow to have it global so each domain does not need to have a path attached to it.


benwiesner1:
closing and re-commiting cleaner version

PR #4609, opened 2019-09-15
Fix table name, variables and syntax Follow me Lua

FozzyGo4:
Correct the wrong table name on line 123. Correct the param on line 126 and correct the syntax on line 347.


markjcrane:
Thanks for the changes. We have validated your changes and now accepting your pull request.

PR #4617, opened 2019-09-17
Registrations page: Parse for LAN IP in "Contact" field

afanucchi:
1. Registrations XML from Freeswitch contains invalid characters '&gt' and '&lt' in the contact field; stripped. 
2. Registrations does not parse the Freeswitch 'Contact' field for a Yealink (or other) LAN IP address; added condition.


emaktech:
We have tested this PR and can confirm that it works - it correctly parses LAN IP from all Yealink phones on all firmware versions. We have not noted any adverse affects on the registrations.php page.

PR #4633, opened 2019-09-19
Update {$mac}.cfg

benwiesner1:
forgot to add "set" for mac address tag


benwiesner1:
copied wrong, re-submiting

PR #4753, opened 2019-10-06
Update index.lua

greenbea:


greenbea:
we want only to increment the timeout if it rang a destination and it didn't answer, but if a destination was busy it shouldn't increment the timeout as it didn't ring yet

PR #4754, opened 2019-10-07
Update {$mac}.xml

greenbea:


markjcrane:
In your pull request I see two p codes for call waiting why are you removing the variable off of one of the p codes? My guess it will fix it for one firmware and break it on different firmware?

greenbea:
The other p-code is for  Disable Call-Waiting Caller ID, I looked at other templates and it doesn't contain the variable for that p code

markjcrane:
I see I missed the Caller ID part of that.

PR #4793, opened 2019-10-18
Adding new Mitel 5320e template

tricksol:
Mitel's are very finicky and I needed to create the exact model to provision. Because all models download the MN_Generic.cfg  I had to update other models to not clobber the one needed for the 5320e. The previous committer of these templates hardcoded the admin and user passwords. I set them to use the variable {$admin_password} and {$user_password}


tricksol:
Screenshot at 2019-10-18 00-25-15
Screenshot of the template.

PR #4798, opened 2019-10-25
Create {$mac}.xml

Len-PGH:


markjcrane:
Isn't there a non p code version of the configuration for these phones?

PR #4802, opened 2019-10-28
sip auth check if domain is enabled

Reltik:
This prevents sip users to sign in if the domain is disabled.


daniel-lucio:
Why this hasn't been accepted?

markjcrane:

PR #4808, opened 2019-10-30
Update 400_recordings.xml

mtghr:
the system didnt ask me for the ID


markjcrane:
I made this code and changed it intentionally to use recording id instead of recording slots this was not an accident.

mtghr:
ok,

now i see that the reason it didnt work is cause recording_id was = true i need it to remove "true"

mtghr:
i am just wondering where that "true" came from 
is there any other place that will change this code from the dial plan?

markjcrane:
The true was an old value. That was there since the dialplan was created.

PR #4810, opened 2019-10-30
Update destination_edit.php

greenbea:
fixed SQL query to check for destinations duplicates if the number already exists with a prefix


markjcrane:
This looks like a nice little improvement.

PR #4819, opened 2019-11-01
Update xml_cdr_details.php

chansizzle:
remove double escaping


markjcrane:
Reviewed the code doesn't look like it is double escaped. What I saw was urldecode and wordwrap. The like escape function is still needed.

PR #4820, opened 2019-11-03
Call forward problems

borisk1976:
From my point of view we have a big problems with call forwarding. How the current situation looks like: 
A calls B. B set unconditional (for example) forward to C. If C is inside the same domain there is no problem. C sees the number of A is calling. 
(we has sip_from A, and sip_to C)

Ok. What happens if C is outside number? Mostly the call fails as external PBX will not pass original A number as it is waiting for number of B. (sip_from B, sip_to C). Yes, we can achive this by Outbound Caller ID Number, but we loose A number then.

So all we know there is Diversion header which can solve the problem. I wrote a dirty patch for failure_handler/index.lua and call-forward application. It adds sip_h_Diversion variable. It is very dirty as it only solves one situation when A and C are both outside. But I want to write a complex patch for all situations. So I need a help - forward system of FusionPBX is too comples (LUA, XML). What files should I explore? What situations should I consider?

Regards,
Boris


markjcrane:

PR #4836, opened 2019-11-06
Update disa.lua

greenbea:
- enabled to set a caller id for disa

Was there a reason this was commented out?https://github.com/fusionpbx/fusionpbx/commit/f50b7be40a814e62110f0ef507a69e7209cf48fd#diff-16f9c832e4c7d11b52601e9d1f732c79


markjcrane:
Look down lower in the code for caller id name and number.

greenbea:
line 158
```lua
if (caller_id_name) then
 --caller id name provided do nothing
```
So it gives you the option to set that variable so it has be defined somewhere

markjcrane:
I reviewed it again and agree with you. Accepting your pull request.

PR #4842, opened 2019-11-08
Create {$mac.xml}

tnssfairchild:
This is a template for poly-obi-vvx450

Buttons 1-24 are left side pages 1 - 4
buttons 25-48 are right side pages 1 - 4

Variables used in default settings (all text) 
poly_obi_dst_enabled   (Day Light savings enabled True / False)
poly_obi_logo_450 (URL TO LOGO FILE)
poly_obi_timezone (For example: "GMT-10:00(Hawaii)" )


markjcrane:
I think it might be better to create a new vendor called poly.

tnssfairchild:
My reasoning for putting under Obihai was that that the phone is technically running the Obihai firmware. I do agree a new vendor "poly" would be better. 
These phones can run both UC firmware and Obihai firmware.

Ill work on  submitting an updated pull request with a new vendor.

ill keep my variables poly_obi (since these are for poly phones running obi firmware)

Should the folder structure look like this?

poly/vvx_450_obi   (for obi config) 
poly/vvx_450_uc   (for standard UC config)

let me know what you think.

tnssfairchild:
I have made new vendor poly, created app_config.php for it, and updated the template for the proper "poly" folder.
l
let me know how it looks. 
Thanks

PR #4843, opened 2019-11-10
Ability to set inbound route condition with sip_to_user

greenbea:
Some carriers are not sending the destination number in the INVITE line, only in the To header field so ```destination_number``` condition won't work in such case  so need instead ```${sip_to_user}```.

I see support for this is already halfway there 
```php
if (strlen($_SESSION['dialplan']['destination']['text']) > 0) {
    $dialplan["dialplan_details"][$y]["dialplan_detail_type"] = $_SESSION['dialplan']['destination']['text'];
}
else {
    $dialplan["dialplan_details"][$y]["dialplan_detail_type"] = "destination_number";
}
```
But with that approach the user has to set the ```['dialplan']['destination']['text']``` in default settings. This limits the user to use the same kind condition for all his destinations (Beside only people reading the code know that this is even possible to do).

I added support for being able to select that the dialplan_detail_type should be sip_to _user by adding a field "Number from To header" and selecting it to true

### DIFF
--- a/app/destinations/app_config.php
+++ b/app/destinations/app_config.php
@@ -169,6 +169,10 @@
 		$apps[$x]['db'][$y]['fields'][$z]['type'] = "text";
 		$apps[$x]['db'][$y]['fields'][$z]['description']['en-us'] = "Regular Expression version of destination number";
 		$z++;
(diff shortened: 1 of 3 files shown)


markjcrane:
Because you didn't know it was there doesn't mean no one else does. And the change suggested could be better.

PR #4846, opened 2019-11-12
Update app_config.php

Len-PGH:


markjcrane:
Set the default to 0.

PR #4856, opened 2019-11-14
UI Suggestion: Increase Data Area Width in Tables

emaktech:
Currently in the UI, there is a "loss factor" of 30% of screen real-estate for data tables which house a large number of columns/fields just for row labels.

Reducing the width of the label column to 5% leaves 95% of the space available for the important data input fields that users interact with. Also especially helpful for mobile users.

Thoughts?

### DIFF
--- a/app/devices/device_edit.php
+++ b/app/devices/device_edit.php
@@ -784,10 +784,10 @@
 
 	echo "<table width='100%' border='0' cellpadding='0' cellspacing='0'>\n";
 	echo "<tr>\n";
-	echo "<td class='vncell' width='30%' valign='top' align='left' nowrap='nowrap'>\n";
+	echo "<td class='vncell' width='5%' valign='top' align='left' nowrap='nowrap'>\n";
 	echo "	".$text['label-device_mac_address']."\n";
 	echo "</td>\n";
-	echo "<td class='vtable' width='70%' align='left'>\n";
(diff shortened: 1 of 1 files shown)


markjcrane:
I'm not sure there is really a loss factor here on a smaller screen the area is automatically smaller. I will discuss this with some additional people and see what they think.

reliberate:
I see what you're referring to -- the dead space on the left side of the form labels (the shaded areas in the attached screen capture below).  

screen_20191121_094515

Unfortunately, in an effort to maintain consistency throughout the application, the column of field labels on this page is set the same as other form pages.  I agree, it's less than ideal, considering the form itself, as it does result in taking up quit a bit of the available horizontal space.  The better solution would be to change the layout of the page to accommodate the greater-than-usual number of fields.  At the moment, however, we are focused on other higher-priority changes.  It's always a challenge to do what works best, but maintains consistency.  We'll likely revisit the layout of this page in the future, and appreciate your input.

In the meantime, if you want to utilize a little more of the available horizontal space, check out the **Theme** > **body_width** Default Setting.  If you set it to **100%**, the left and right margins of the page will shrink allowing you to view a bit more on the screen.

Hope that helps.

markjcrane:
The UI adapts and provides less space to the Label column if the screen size is smaller.

PR #4881, opened 2019-11-19
Update disa.lua

greenbea:
- fixed caller_id bug on disa line 50 and 51
```lua
caller_id_name = session:getVariable("caller_id_name");
caller_id_number = session:getVariable("caller_id_number");
```
was always getting set because it's a freeswitch variable

- commented unnecessary lines


daniel-lucio:
Can you explain the rationale of adding new disa_caller_id_* variables?

greenbea:
I didn't add it it was there before, it's meant to be able to able to set seperate diss caller id 

1.

PR #4885, opened 2019-11-21
Fix SQL in cidlookup script.

sergey-mz:
1. There should be a space at the end of line. Otherwise, resulting SQL has "callerWHERE" error.
sql = sql .. "INNER JOIN v_destinations ON v_destinations.domain_uuid = v_contacts.domain_uuid AND v_destinations.destination_number = :caller";
2. Destination number will never match with caller's number. The result will be nil.
        v_destinations.destination_number = :caller
3. There is no reason to join destination table at all. It just multiplies the number of rows in the result.

### DIFF
--- a/resources/install/scripts/cidlookup.lua
+++ b/resources/install/scripts/cidlookup.lua
@@ -87,7 +87,6 @@
 		sql = "SELECT v_contacts.contact_name_given || ' ' || v_contacts.contact_name_family AS name FROM v_contacts ";
 	end
 	sql = sql .. "INNER JOIN v_contact_phones ON v_contact_phones.contact_uuid = v_contacts.contact_uuid ";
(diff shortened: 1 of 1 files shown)


daniel-lucio:
The missing space was a typo. I will send a fix for that.

however, I totally disagree with taking out the last INNER JOIN. The purpose of this is to identify the incoming call domain_uuid using the destination table, then it will use only the agenda of that given domain.

If domain1 has "ABC" for number 9999999999, and domain2 has "ZXY" for the same domain the query would return 2 entries. If number 9999999999 calls a DID linked to domain1, he will see ABC, if number 9999999999 calls a DID linked to domain2, he will see ZXY.

sergey-mz:
Ok. I forgot about multiple domains. Still, filter "AND v_destinations.destination_number = :caller" is wrong. Let say I have DID 9050000000 in destinations (destination_number) and I receive a call from 6470000000  ( :caller) then the query will return zero records. We have to get domain UID and filter by it.

sergey-mz:
I created another request to fix the issue.

PR #4887, opened 2019-11-21
Add grandstream_call_waiting and sip_port variables to Grandstream AT…

demonspork:
Added grandstream_call_waiting and {$account.x.sip_port} variables to the HT802 template.

Does anyone have any thoughts on the newer "call features" and whether or not they should be disabled? *72, *73 etc are processed locally on the ATA unless we default P191 to 0 (the template currently sets this variable to 1)


digitalcrisis:
Definitely use the FusionPBX rather than the ATA, especially if it's for a client. So you can control call forwarding from the WebUI.

markjcrane:
I just checked and didn't see you signed up as a contributor. Please sign up as a contributor here.

demonspork:
I signed the CLA so we should be good to go.

PR #4888, opened 2019-11-22
Fixed Thanksgiving Time Condition to work every year

digitalcrisis:
Thanksgiving is always the fourth Thursday on November ever year.


markjcrane:
Thanks for the attempt to improve the code. However mweek doesn't work they way you think it does. Your suggestion would actually not work for this years Thanksgiving but it would work next year and break intermittently on other years.

PR #4889, opened 2019-11-22
Update Adminer to v4.7.5

powerpbx:
Could not get the current version to work with Postgres 12 so we forked the repository and created our own branch to compiled our own custom file.  It may make changes easier in the future.  

This file was compiled as multi-language.  If that is a problem or if anything else needs to be changed it should be easy to compile another file.


markjcrane:
Thanks will give your version a try.

markjcrane:
Login still failed on my server.

powerpbx:
I presume you are referring to autologin.  That is a separate issue.  That may require some changes on the fusionpbx autologin functions.  For now you just need to press the login button.  I haven't found any documentation about it but there was a commit that prevents direct logins for security reasons.  So that might have something to do with it.

markjcrane:
No I tried manual login and it failed. I understand how to manually login and the previous version allowed me to autologin but it wasn't consistent and didn't not always work either like this one.

markjcrane:
Adminer login is flaky on the last couple versions.

powerpbx:
I added a new commit which rolls back the upstream commit that I think causes the problem.  Let me know if this fixes the autologin problem.

https://github.com/powerpbx/fusionpbx/commit/54daaf214b2385385d44d08840896d072d741e31

markjcrane:
Would be nice if you sent it as a pull request without that it won't get added to the code and likely adminer will be removed.

markjcrane:
Never mind I'm removing adminer. In favor of more secure solutions.

PR #4911, opened 2019-11-25
Get domain_uuid and use it in SQL to filter

sergey-mz:
1. Get domain_uuid of the call using API. argv[2] is not used in dialplan - remove the reference.
2. Remove JOIN with destinations.
3. Use domain_uuid to filter contacts.


daniel-lucio:
You can't get rid of the last INNER JOIN, also I am not sure if the domain_uuid variable is set the way you are getting it (need to verify)

sergey-mz:
You can make the changes and enable SQL debugging.
The resulting SQL will include filter by v_contacts.domain_uuid and it will return only one record or nil.
2019-12-02 10:53:13.429026 [NOTICE] switch_cpp.cpp:1341 [cidlookup] SQL: SELECT CASE WHEN contact_name_given = '' THEN v_contacts.contact_organization ELSE v_contacts.contact_name_given || ' ' || v_contacts.contact_name_family END AS name FROM v_contacts INNER JOIN v_contact_phones ON v_contact_phones.contact_uuid = v_contacts.contact_uuid WHERE v_contacts.domain_uuid = :domain_uuid and v_contact_phones.phone_number = :caller
2019-12-02 10:53:13.429026 [NOTICE] switch_cpp.cpp:1341 [cidlookup] params:{"domain_uuid":"41b4f79b-2c47-42ab-8356-a0beca120d9f","caller":"6471234567"}

If you don't get rid of the last JOIN then it will return a record per destination, but you will use only the last one. What's the point of doing this?

markjcrane:

PR #4914, opened 2019-11-25
Incorrect Phrases variable used

jpattWPC:
This prevents any phrases from showing up for the 'invalid sound' setting in an IVR.


markjcrane:
I just confirmed it is fixed.

PR #4916, opened 2019-11-26
Allow permission to export devices and extensions to be assigned to groups

saraxxin:
Add new permissions extension_export and device_export which may be assigned under group manager. Default only allows superadmin.


markjcrane:
Good improvement thanks... Accepting your pull request.

PR #4919, opened 2019-11-27
Update fax_retry.lua

greenbea:
Tweaked for loopback support


greenbea:
Did some testing and loopback doesn't seem to be working with api_hangup_hook

PR #4924, opened 2019-12-02
Send 404 as a WARNING instead of DEBUG

daniel-lucio:
New system use the journal (some still use syslog). the "log" action default level is DEBUG and it is a bad idea when loading mod_syslog to send all the debug, it would kill the system.

This change moves the 404 error code to WARNING, the mod_syslog will report to the journal (or syslog).


markjcrane:
This pull request says it 422 files changed need to be broken down in smaller chunks.

daniel-lucio:
oh I did a mistake, i will resend it

PR #4937, opened 2019-12-11
Some Loopback endpoint support

daniel-lucio:
Leg B won't have the variable domain_name filled correctly, regardless if you do the right loopback/user/domain values.
however, the context variable is filled correctly. This will allow us to let all the dial plans run correctly.


markjcrane:
The context is usually the domain name but not always.

daniel-lucio:
yes, they are 2 independent variables. But in fusionpbx, we have made them match (except public context).

This only auto-fills the domain_name variable when it is empty, I have seen this only happens when a loopback is involved.

daniel-lucio:
I have found a better way, let me update the pull requests. Hold a moment.

daniel-lucio:
Fixed, I found a better way. When there is a loopback endpoint involved, the variable ${loopback_leg} exists. Then when loopback-b is executing, we can make this correction

PR #4942, opened 2019-12-13
Fix time conditions display bug

jpattWPC:
When displaying time conditions on an existing time condition the last preset in the array would be overwritten with the previous preset. For example, if the last 2 presets were black friday and christmas, then black friday would appear twice and the christmas actions would show as a manually set time condition in the 'Settings' section. This is because $preset was being reused by PHP. Renaming the $preset variable for loops resolved this issue.


markjcrane:

PR #4954, opened 2019-12-16
Conference Sessions - Use Database Recording Path

emaktech:
Previously, the recordings on this page were being fetched based on the recording directory default setting. This instead uses the recording field in the v_conference_sessions table.

It is necessary to use the field in the database for users that are moving their call recordings to another storage medium and modifying the recording path in the database.


markjcrane:
Can you create a new pull request for this if it hasn't already been applied?

greenbea:
This was already applied a few months ago

PR #4971, opened 2019-12-19
Update domains.php

greenbea:
the delete button was commented out


fusionate:
The ability to bulk delete domains was disabled intentionally.  A Delete button is now displayed (instead) on the Domain Edit page.

PR #4972, opened 2019-12-19
Incorrect SQL query in groups

jpattWPC:
view_groups does not exist, changing this to v_groups resolves the issue of no groups showing in the group manager

### DIFF
--- a/core/groups/groups.php
+++ b/core/groups/groups.php
@@ -92,7 +92,7 @@
 	}
 
 //get the count
-	$sql = "select count(*) from view_groups ";
+	$sql = "select count(*) from v_groups ";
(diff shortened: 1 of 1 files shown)


fusionate:
Just FYI, "view_groups" is a database _view_, that should be created by /core/groups/app_defaults.php when Advanced > Upgrade > App Defaults is run.  It's probably supposed to be as it was... but if the page isn't working, there may be a different issue.

jpattWPC:
When did view_groups get added? I am wary of running app_defaults for anything, and I ran a schema check and this did not resolve the issue.

markjcrane:
This would create a bug and it is supposed to use the view.

PR #4973, opened 2019-12-19
Update message_delete.php

greenbea:
message media wasn't in the array


fusionate:
Thanks, good catch.

greenbea:
It's missing as well from the messages.php class,

PR #4976, opened 2019-12-19
Update index.php

greenbea:
Added the ability to set a user setting to control if we want to get the messages forwarded to a sip endpoint


greenbea:
Added it forwarding to email support

greenbea:
Will resend the pull request a little cleaner

PR #4978, opened 2019-12-20
[4.5] Fix call limit

daniel-lucio:
Same as 4.4


app/dialplans/resources/switch/conf/dialplan/020_call_direction.xml
```diff
@@ -1,7 +1,16 @@
 <context name="{v_context}">
 	<extension name="call-direction" number="" continue="true" app_uuid="3780f814-5543-4350-b65d-563512d1fe71" enabled="true" order="20">
+		<condition field="context" expression="^(public$|outbound@)" break="on-true">
+			<action application="set" data="call_direction=inbound" inline="true"/>
+		</condition>
+		<condition field="destination_number" expression="^([\d\D]{1,6})$" break="on-true">
+			<action application="set" data="call_direction=local" inline="true"/>
+		</condition>
```
borisk1976: IMHO this is dangerous to treat call as "local" with this regexp. For example in Russia there are many small towns where phone numbers are 5 or 6 digits.

daniel-lucio: I didn't decide that. I took the logic from the lua script $script_dirs/app/dialplan/index.lua which governs all the Fusion dialplan feeding.

I think specific country (cities) cases must be addressed locally.

borisk1976: Yes, I see.... that's very strange and the logic of this decision eludes me. May be Mark can explain more? Also I found other scripts where call_direction variable is set: follow_me, dialplan, ring_groups. Isn't better to create one common function to set call_direction var? And it will be more simple to maintain one peace of code.

markjcrane:
I don’t think the assumption is safe to use as a default for everyone. Here is what the dialplan looks like by default currently.
https://github.com/fusionpbx/fusionpbx/blob/master/app/dialplans/resources/switch/conf/dialplan/020_call_direction.xml

borisk1976:
Mark, but the question still here - what piece of code is responsible for setting call_direction? In your example this code only set "local" when call_direction is none of inbound,outbound,local.

daniel-lucio:
So, could it be an alternated, disabled dialplan?

daniel-lucio:
> I don’t think the assumption is safe to use as a default for everyone. Here is what the dialplan looks like by default currently.
> https://github.com/fusionpbx/fusionpbx/blob/master/app/dialplans/resources/switch/conf/dialplan/020_call_direction.xml

The problem here is that I found outgoing calls where the call direction variable was not set properly.

borisk1976:
IMHO we should not use "quick patches". I think we should 
1) Discuss what calls we can treat as inbound/outbound/local
2) Create "one place" solution where call_direction variable will be set. I think this will be a LUA script which we can use either from XML or LUA

From my point of view:
1) Calls placed into public context are inbound
2) Calls leaving current context are outbound
3) Other calls are local

But where is the right place to set the variable? Can we rely on freeswitch direction variable?

markjcrane:
This pull request cannot be merged because of a former change in the code. If you would like to add your change then please add the changes to the new version of FusionPBX and add a new pull request. Sorry about this problem.

PR #5009, opened 2020-01-07
Fix the CDR Archive

daniel-lucio:
The archive works almost out of the works, but it is missing some little things:
- xml_cdr_archive permission is never defined
- button-archive is never defined
- variable $archive_request is never defined (but evaluated).

This patch enables de "archive" button, and then it allows us to use the DB specified in the ['cdr']['archive_database'] default settings set.

No SQL modifications are required.


markjcrane:
4.4 is considered deprecated but its a bug fix so accepted the change.

PR #5010, opened 2020-01-07
[4.5] TouchIcon Support

daniel-lucio:
When accessing Fusion from a table/celphone, if the touchicon tag is enabled, it will allow it to be displayed when bookmarked.

This won't replace favicon on common browsers.


markjcrane:
Closing pull requests that Shows 5000+ changes.

PR #5011, opened 2020-01-07
[4.4] TouchIcon Support

daniel-lucio:


daniel-lucio:
What's the problem with this, this is a really nice now that tablets are popular

markjcrane:
Not adding new features to old release.

PR #5045, opened 2020-01-09
[4.4] Enable caller_id_name and number in ring group

mat1010:
Currently the options `caller_id_name` and `caller_id_number` are not used within the dialplan for ring_groups, even though they are properly set and inserted into the database. This change will add those options into the dialplan xml for a ring group if set.


markjcrane:
The Caller ID name and number are set in the LUA script
https://github.com/fusionpbx/fusionpbx/blob/4.4/resources/install/scripts/app/ring_groups/index.lua

mat1010:
Thanks for the update. Looking at the script, I see that the LUA script is using the `ring_group_cid_name_prefix` and `ring_group_cid_number_prefix`. But those variables cannot be defined within the form for the ring groups. Or am I missing something? The form only fills the values for `ring_group_caller_id_name` and `ring_group_caller_id_number`

PR #5052, opened 2020-01-12
Improve recordings base64 efficiency

jpattWPC:
The current method of dealing with Base64 recordings does not scale. It requires querying ALL recordings from the database into memory (including the full base4 text) every time the recordings.php page was loaded. This was causing the default PHP memory limit of 128MB to be exhausted which would cause the recordings section to become inaccessible. The solution is to have the database indicate whether something is stored in the base64_recording field rather than have PHP make that determination, and if there is something then we pull individual recordings in and deal with them one at a time. This reduced page load times for recordings.php from close to 10 seconds down to less than 1 second when dealing with about 2000 recordings.

Please note: I do not have a MySQL database to test against (only PostgreSQL) so this may need further testing to determine if the SQL is compatible. From cursory research I believe it is but I can't be sure.


markjcrane:
Closing pull request if still interested in adding changes update them and submit a new pull request.

PR #5118, opened 2020-01-16
Backwards compatibility

daniel-lucio:
The right way to take out a function is first deprecating it in one release (4.6?) and eliminating it in the next stable.

### DIFF
--- a/resources/functions.php
+++ b/resources/functions.php
@@ -28,18 +28,21 @@
 	// Deprecated, backwards compatibility. This functions will be gone in 4.7+
         if (!function_exists('software_version')) {
                 function software_version() {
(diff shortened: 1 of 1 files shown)


markjcrane:
What you haven't revealed is why you need it. No other code uses this.

daniel-lucio:
The first reason is the correct process to dissolve an internal API function. Any project should deprecate a function and then in the next release dissolve. 
There are many other coders who have their own apps, and because of the differences between 4.4 and 4.5+, a version check must be done to interact correctly.

I have found my own way to figure this out, but for the sake of the project and others' apps, I think this is the correct way.

markjcrane:
Just update your code.

PR #5123, opened 2020-01-21
Added voicemail domain permission

greenbea:
The code used to force to give voicemail_delete permission in order to view all domains voicemails


greenbea:
wrong permission name

PR #5139, opened 2020-02-03
Update provision.php

PBXForums:
Make a key available in the template for whitelisting.


markjcrane:
Why could this not be done with a device setting?

markjcrane:
I have lots of concerns with this pull request.

DigiDaz:
Yes, its a calculated value.

markjcrane:
I think there is a likely a better name for the variable. Will try to find some time come up with a name.

markjcrane:
Closing pull request if still interested in adding changes update them and submit a new pull request.

PR #5141, opened 2020-02-04
Update TLS to v1.2

z47co:
TLS v1.0 is deprecated by many SMTP providers. Use TLS v1.2 instead.

> TLSv1.1 and TLSv1.0 are also actively being deprecated in accordance with guidance from government agencies (e.g. NIST SP 80052r2) and industry consortia such as the Payment Card Industry Association (PCI) [PCI-TLS1].
> 
> The primary technical reasons for deprecating these versions include:
> 

- > They require implementation of older cipher suites that are no longer desirable for cryptographic reasons, e.g. TLSv1.0 makes TLS_DHE_DSS_WITH_3DES_EDE_CBC_SHA mandatory to implement
- > Lack of support for current recommended cipher suites, especially using AEAD ciphers which are not supported prior to TLS 1.2
- > Support for four protocol versions increases the likelihood of misconfiguration
- > At least one widely-used library has plans to drop TLSv1.1 and TLSv1.0 support in upcoming releases; products using such libraries would need to use older versions of the libraries to support TLSv1.0 and TLSv1.1, which is clearly undesirable

https://tools.ietf.org/id/draft-moriarty-tls-oldversions-diediedie-00.html


markjcrane:
This pull request cannot be merged because of a former change in the code. If you would like to add your change then please add the changes to the new version of FusionPBX and add a new pull request. Sorry about this problem.

PR #5153, opened 2020-02-18
Allow user with only contact_view permission to click on contact record to view numbers etc.

AdrianFretwell:
A user with only contact_view permissions is unable to click on the contact record to view the numbers and addresses etc.

By add allowing contact_view to have the contact_edit.php link they can now view the numbers etc.

Only showing the save button for users with contact_edit and contact_add permissions reinforces the fact that a user with contact_view only cannot edit the contact.

### DIFF
--- a/app/contacts/contact_edit.php
+++ b/app/contacts/contact_edit.php
@@ -495,7 +495,9 @@
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


markjcrane:
I'll accept the pull request for now. The page design will be changed to favor a view with an edit button in the future.

PR #5158, opened 2020-02-26
Time Conditions to respect time_format variable

igorolhovskiy:
The idea is to Time Condtitions respect Domain - Time Format - variable in a case if it's set to 24h (Not use this non-US AM/PM things)


borisk1976:
Great! +1

markjcrane:
This pull request is old if you would like to see the changes get added then update the changes with latest code and send a new pull request.

PR #5159, opened 2020-02-27
Update app_config.php

mtghr:
option to disable the DND softkey


markjcrane:
Notice all key names in default settings are lowercase and most often separated with an underscore. The change is fine if it’s consistent with the rest of the project.

mtghr:
i fixed it 

thank you

PR #5167, opened 2020-02-28
Update index.lua

borisk1976:
We must call session:hangup("USER_BUSY") in all causes. Without it freeswitch (according to https://freeswitch.org/confluence/display/FREESWITCH/Hangup+Cause+Code+Table) add additional header "Reason: Q.850; cause=16; text=NORMAL_CLEARING" and it causes problems with some SIP<->PRI gateways. They just translate to PRI Q.850 header they see.


borisk1976:
Sorry, this patch breaks voicemails, so I close it.

PR #5173, opened 2020-03-02
Added ability to have custom url for password reset links

greenbea:


markjcrane:
This is a good change need to decide whether reset_url is the best variable name and also whether we should be using a single variable for specifying a centralized domain rather than having one for every feature.

markjcrane:
This pull request will need to be updated. Please resubmit.

PR #5177, opened 2020-03-10
Remove hard-coded sounds path

racitup:
There should be a way of reading the sound_prefix path from the switch variables, so this is a bit of a hack. Better way needed.

It appears there are more places en/us/callie is hard-coded too:
https://github.com/fusionpbx/fusionpbx/search?q=%2Fen%2Fus%2Fcallie&type=Code


racitup:
Will remove all hard-coded paths on new pr for correct international use

PR #5180, opened 2020-03-11
Remove uuid.php there not use in project code

roramirez:


markjcrane:
It is used by the install script.

PR #5182, opened 2020-03-14
Attempt to add en-gb and fix non-default language issues

racitup:
I have introduced a new 'variables' class that is used by files->sounds() (and can be used in other places) to retrieve configured switch variables from the db, and expand the ${}'s, so the interface can be truly multi-lingual to match the sounds files.


markjcrane:
Pull requests should be one topic only. This pull request 3 or 4 different changes. Please resubmit them separately.

PR #5188, opened 2020-03-16
Yealink CP930W initial config files

bajizs:
Initial template for Yealink cp930 conference phone


bajizs:
This template not neaded, its a dect version, use w60b dect base...

PR #5189, opened 2020-03-17
Setting a conference room sounds=false

konradSC:
Right now settings a conference room's to 'sounds=false' doesn't do anything. This fixes that bug. 

Additional updates to the script most likely need to be made after this so that conference_enter_sound and conference_exit_sound honor the value specified in the conference profile. I didn't see any code referencing the conference profile values for sounds.


markjcrane:
Thanks for this bug fix accepting your changes.

PR #5190, opened 2020-03-18
Update All Mitel Templates to use variables

tricksol:
Updating the 5320 / 5324 / 5330 / 5340 templates to use account variables like the 5320e template that I committed on "6f81b62"


markjcrane:
Looks good accepting your changes.

PR #5193, opened 2020-03-18
[4.4] Cosmetic fix

daniel-lucio:
This is fixed in 4.5 but not in 4.4. selected='selected' shouldn't be scaped in the <option> tag


markjcrane:
Accepting the fix.

PR #5196, opened 2020-03-20
Improve multi-lingual support

racitup:
Refactor languages and modify ivr_menu & phrase apps


racitup:
One more change to languages.lua inbound before merge for proper multi-lingual ivrs built from phrases

racitup:
All finished, I don't expect any more commits. Tested and working.

racitup:
Actually, hold off on this. I will need to reverse the language override logic so that phrases do not require language to be set, which means they can be overridden by the dialplan (default) language if left blank. Will do it within 24 hours but not now.

racitup:
Now fixed and working

racitup:
😆 Making your contributors rebase all their work is one way to pi$$ them off I suppose!

PR #5204, opened 2020-03-24
[4.5] Fix fax email Subject when not using sofia/gateway

daniel-lucio:
Same as 4.4


siacali:
This would solve a huge problem for us.  We get Subject: Fax to:  0 SENT, which makes the faxserver less than useful for email to fax users.

PR #5206, opened 2020-03-25
Add one more fallback option to get domain_name

samael33:
In `XML_HANDLER`.
Also done a small refactor.
`Caller-Context` is added as an additional fallback for getting `domain_name`
due to some complex scenarios (mine was IVR as a Ring Group member)
this variable did not survive all loopback transfers


markjcrane:
I fixed domain_name for ring group members yesterday.

samael33:
Ah. Ok. I'm not using master, fixed it here in 4.4, but found same code in master.

PR #5210, opened 2020-03-26
Update {$mac}.xml

jasonkinner:
P2916 (wallpaper source) was defined twice - comment out first definition outside of if/else


markjcrane:
Will need to update your pull request and resend.

PR #5211, opened 2020-03-26
Update {$mac}.xml

jasonkinner:
P2916 (wallpaper source) was defined twice - comment out first definition outside of if/else


markjcrane:
Please update your pull or re-base your changes so that the commit doesn't show 5,606 files changed. Then resend request again.

PR #5213, opened 2020-03-27
adding icon to raise hand 

hershyheilpern:
adding a hand icon when hand is raised


markjcrane:
We use font-awesome its likely that there is already an icon for hand raise without having to add a new icon.

hershyheilpern:
i will look in to it

hershyheilpern:
where do i find the font details

hershyheilpern:
changing to font-awesome

PR #5216, opened 2020-03-27
[4.5] REQUESTED_CHAN_UNAVAIL option in xml cdr

daniel-lucio:
same as 4.4


markjcrane:
just another option for searching calls

PR #5218, opened 2020-03-29
Fixed background colors

greenbea:
The ``login_background_color`` and ``login_background_color`` got mixed up


fusionate:
Thanks for bringing this to our attention.  The fix wasn't actually quite as involved - just needed to adjust the logic a little.  See if the latest commit works for you.  My tests show it working correctly now.

greenbea:
Yes, I see, I missed one part there, looks good. Thanks

PR #5220, opened 2020-03-29
add count to conference center hand raised

hershyheilpern:
gives you a count on how many hands are raised 

i know this isn't written in the most optimized way if you have any suggestions please let me know


markjcrane:
I'm questioning how the hand raising is performed. Should have talked to me about the design ahead of time. Grateful for the feature but it maybe rewritten.

hershyheilpern:
i'm open for any suggestions when can we talk

PR #5221, opened 2020-03-30
Remove hard coded setting for moderator conf ctls

konradSC:
Conference Controls for moderators are hard coded. Because of this you cannot use the Conference Controls app in the web interface to create a new set of moderator controls and set them in the Conference Profile.


markjcrane:
Please call and discuss your changes. This might be good for you because you wrote additional code to set it somewhere else but for everyone else they would loose the ability to have a separate conference controls for the moderator.

markjcrane:
Next time I talk to you I plan to ask about this during that call.

PR #5223, opened 2020-03-30
conference center announcement control

greenbea:
added ability to control member count announcement and recording announcement


greenbea:
resent with minor changes

PR #5226, opened 2020-03-31
Refactor and fix language support

racitup:
NOTE: Depends on #5225 
The original language support completely ignored locales of the same language. "en" was hard-coded to "en/us/callie" for example. Whilst the project is moving toward dynamically generated language xml, it isn't there yet and still relies on hard-coded language xml files.
This PR changes the language reference from 2 characters to the full string: en-us-callie.
This allows anyone to mix and match languages, dialects and voices with ease.

The PR also fixes phrase support which was broken.
Whilst this is a step in the right direction and works, it is not complete, and is merely less broken than it was. The phrase support for instance generates (dynamically into the cache dir using lua) a language file and maps it to the language specified by the dial-plan. However, the phrase itself has a language set which may not match the dial-plan. This is intentional, since a German IVR may wish to play French phrases for a multi-lingual menu. But the phrase php app has no idea what language is being used by the dial-plan, so when changing the phrase, it needs to clear the cache, but in the old cache key naming scheme it had no idea what that key was. To fix this issue, the dial-plan language was removed from the cache key. This will need to be put back once the cache clearing problem is solved.


markjcrane:
Project relies on hard coded languages files with intent. I don't think this is a bad thing. Unless we want to make it easier to break the system with the GUI. Phrases tool was added in the GUI for user or custom phrases.

Large sweeping changes like this need more explanation by phone call and remote desktop to discuss reasons for various changes. Otherwise smaller pull requests with bite sized changes are preferred as they are easier to discuss over text.

racitup:
> Project relies on hard coded languages files with intent. I don't think this is a bad thing.
That's because you live in North America and it's hard-coded to en-us-callie. For the rest of the world, it is a bad thing.

> Large sweeping changes like this need more explanation by phone call and remote desktop to discuss reasons for various changes. Otherwise smaller pull requests with bite sized changes are preferred as they are easier to discuss over text.
I contributed a significant amount of work to your open-source project to make it truly multi-lingual. I'm not about to waste more of my time and money explaining the fixes more than I already have. If you don't want to review and merge it, it's your project, your decision. Just don't make excuses. All the changes are related so can't be split.

PR #5229, opened 2020-04-02
Daylight Saving changes

bajizs:
- enabled on server auto.DST.xml configuration for phones
- added new autoDST.xml file to fusionpbx server
- corrected autoDST.xml, addde balkans countries


bajizs:
Helo, when can merge this request? Tjis is neaded to finalize DST on yealink devices....

markjcrane:
This setting in default settings yealink_auto_dst.url is unlikely to work. The period usually used in the template engine to indicate an array and what is after that period in the named array in this case would be url.

markjcrane:
If you want to get this committed then call our support number and discuss it over a phone call. Each commit I accept I need to understand the reason for the changes because I will be maintaining them year and after year.

bajizs:
The reason is simple:

Lot of Balkans country (and I don't known, maybe other) don't have his
DST time programed in Yealink phones at the moment. You can Use Serbia
or Belgrade or whatever for example this not work. Time zone works
because You add +1, but if You want to use DST You must to select for
example Hungary as time zone, or every 6 month change Time zine from +1
to +2 and vice versa.

The Yalink added support to custom DST file, where You can add all Time
zones, but Yealink not updated it in last years... With tis option You
can over provisioning support this updates. And if something changes
with DST (I don't known at the  moment but i read news last year to be
changed maybe not used in future), You can it push to Yealink phones
without factory support.

Thanks.

markjcrane:
Thanks for the explanation.

PR #5232, opened 2020-04-02
Add enable disable field to call flows

greenbea:


markjcrane:
Your changes are being reviewed to see if the toggle is being done the same as other features in FusionPBX.

greenbea:
I've done it the same way as conference_rooms.php

PR #5237, opened 2020-04-19
use path to redirect login.php

hershyheilpern:
use the path parameter if you we'er loged out and went to a page to redirect you after loging in


markjcrane:
We had this it was disabled because it was considered a security risk. Would need security improvements in order to be safe.

PR #5238, opened 2020-04-20
Set new channel variable

saraxxin:
New channel variable added so that one may hook on to DISA calls which are not local.
New variable: disa_outbound


markjcrane:
Good idea!

PR #5241, opened 2020-04-23
Update index.lua

Redbehrend:
-Fix bug where confirm didn't work if you used random in ring groups.
-Made random more random when you have a small amount of extensions in a random ring group.


markjcrane:
Looked at your changes and they look good. Then looked here https://github.com/fusionpbx/open-source and didn't see you as registered as a contributor. Please signup as a contributor and I will accept your pull request.

PR #5243, opened 2020-04-24
Specify table for Accountcode Search

konradSC:
I had to add "c." in order for accountcode CDR searches to work under Advanced CDR Search. 

I suspect there may be more that require this, but I haven't checked every field under advanced search.


markjcrane:
There might be more but there are several in fields that already have the c. on them.

PR #5247, opened 2020-04-28
Update mute.lua

greenbea:
more accurate description


greenbea:
I was wrong with this as it's not necessary to pass the channel variable to the lua script. Simple mute and unmute action from the conference controls will do it

PR #5254, opened 2020-05-02
Update {$mac}.xml

Len-PGH:


markjcrane:
This removes grandstream_wallpaper_source was that an accident?

PR #5256, opened 2020-05-06
Phrases add permission

greenbea:
Added permission to disallow users without this permission  to POST phrase execute function


greenbea:
I'll resend for small fix and better permission name

PR #5257, opened 2020-05-06

benwiesner1:
add drop down for announce sound


benwiesner1:
anyway to get this pushed? or should i just add to our code base only

markjcrane:
This will actually cause a regression and loss of features. Its also not the most efficient way to do it.

markjcrane:
Its more efficient to put the announce sound into the music file.

PR #5259, opened 2020-05-07
Fix moderator controls

greenbea:
* hard coding the moderator controls doesn't allow setting up custom controls 
* it isn't necessary to set the controls as it's already set in the profiles


markjcrane:
There is already a pull request for this.

PR #5265, opened 2020-05-11
security - disallow accessing time conditions of other domains

greenbea:


markjcrane:
They would also have to know the time condition UUID which is likely more difficult than your password.

PR #5266, opened 2020-05-12
Yealink Combined Provisioning + for loops

Redbehrend:
Yealink combined provisioning + foreach loop to cut down and simplify account settings.
Created series group which the configuration files go by using the manual, admin manual and yealink configuration editor instead of individual phones.
Would like to do the same for loop to expansion modules but don't have any to test.


markjcrane:
Not a 100% sure if every t2x and t5x could be summed up in one template. However I'll accept them if you change the naming convention from t2series to t2x and so on. This would follow naming convention better.

Redbehrend:
Updated per your request.

markjcrane:
Been trying to decide whether to add these new templates that duplicate current templates but with a loop. Will talk this over with some people tomorrow and over the next few days.

demonspork:
I'm a fan of them. Most of the GXP templates for Grandstream are basically like this, not exactly sure why they are separate in fact I have them set up symlinked to each other to simplify minor customizations.

At the moment with Yealink when I want to add something custom like create a variable for something that isn't available already or just change a default there are waaaay too many separate templates to look through and edit because I want to change it for the entire product line, not a single phone.

greenbea:
I think it's a great idea

PR #5267, opened 2020-05-12
Fixed error when deleting moh as user

anux-linux:
When deleting moh as user the sql statement was corrupt and moh was not deleted.


markjcrane:

markjcrane:
Thanks for fixing this error.

PR #5285, opened 2020-05-27
Checking returnvalue of PDO::execute for errors in database->save()

mhoogveld:
Caught this while importing extensions csv with first line that contained column headers while leaving the "from_row" to 1.
In that case, saving the first row generates a sql error, because of an invalid value for the extension_uuid column (value not being a uuid). This causes execute() to return false which is not handled/checked.
Saving the second row generates a PDOException because of an invalid transaction state, which is displayed to the user.
I throw and catch a RuntimeException, which doubles the catch clause, because I did not want to misuse the PDOException

There are more cases where execute() is used and the return value is not checked. Some, if not all, could be dealt with in similar fashion.

### DIFF
--- a/resources/classes/database.php
+++ b/resources/classes/database.php
@@ -1227,10 +1227,23 @@ include "root.php";
 									if ($prep_statement) {
 										//get the data
 											try {
-												$prep_statement->execute();
+												$res_exec = $prep_statement->execute();
+												if ($res_exec === false) {
+													$errInfo = $prep_statement->errorInfo();
+													if (isset($errInfo[2])) {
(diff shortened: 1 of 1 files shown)


markjcrane:

markjcrane:
Will be adding a way to turn this on or off as it is a security risk to have it on 100% of the time.

mhoogveld:
The way an error is dealt with can indeed be improved to prevent sensitive information to leak and the best way depends on the context, but I think the return value of PDO::execute idealy should be checked in most (if not all) cases and dealt with.

PR #5293, opened 2020-06-03
Update {$mac}.cfg for plycom 4.x

hershyheilpern:
added server address to address line instead of just the user id got issue when doing server1 server2 setup


markjcrane:
Thanks accepting the change.

PR #5295, opened 2020-06-08
$sql and $parameters never unset

frytimo:
Change order of unset statement as it is after a return statement thus unreachable.


markjcrane:
Looks good thanks for the improvement.

PR #5313, opened 2020-06-21
Fix ivr empy exit action

greenbea:
When exit action is empty it causes Freeswitch to log ``[ERR] switch_core_session.c:2697 Invalid Application
``


greenbea:
The action will be hangup even it's empty but freeswitch will generate an error in the if ``<action application="" data=""/>`` is empty

markjcrane:
If there is no action or data it should not be added to the xml and likely should be deleted.

greenbea:
You're right I didn't realize this

greenbea:
I'll resend it with removing it from dialplan

PR #5314, opened 2020-06-21
Add destination->valid check for ivr exit action

greenbea:


greenbea:
Resubmitting better version

PR #5315, opened 2020-06-21
Update destinations.php

hershyheilpern:
show destination even on show all


markjcrane:
That isn’t going to work

hershyheilpern:
worked for me a few minutes ago when i tried it 
what problems can it cause

PR #5324, opened 2020-07-01
Fix Yealink Expansion Modules in Device Profiles

konradSC:
The 'if' statements need to reference v_device_profile_keys.profile_key_vendor instead of device_key_vendor which is in a different table.


markjcrane:
Good catch accepting your pull request.

markjcrane:
This actually seems to  break the yealink expansion modules when using Device Profiles.

markjcrane:
You save the Expansion and it get turned into Line instead of Expansion.

markjcrane:
Looked like you were on the right track just may have missed changing Yealink see the commit for details.
https://github.com/fusionpbx/fusionpbx/commit/90dc96d1340ab9c9f6ed621005163615378e9138

konradSC:
Yep. Looks like I missed one. Thanks!!

PR #5337, opened 2020-07-07
Fix disa caller_id

greenbea:
disa.lua was checking if ``caller_id_name`` and ``caller_id_number`` variables are set. These 2 variables are read-only variables set by FreeSWITCH on every call it always has the values of the caller causing disa to always call out with original caller id

### DIFF
--- a/app/scripts/resources/scripts/disa.lua
+++ b/app/scripts/resources/scripts/disa.lua
@@ -47,8 +47,8 @@
 		sound_extension = session:getVariable("sound_extension");
 		pin_number = session:getVariable("pin_number");
 		sounds_dir = session:getVariable("sounds_dir");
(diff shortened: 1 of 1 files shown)


markjcrane:
You will have to talk to me about this over the phone to discuss this. This will not be backwards compatible and will break people that are already using it.

emaktech:
Interesting, thanks for this!

I always wondered about DISA and we could never use it because it was always passing the CID of the caller rather than the outbound CID of the extension. Very useful!

markjcrane:
Went a little different direction to fix caller ID with this commit. Only setting it when for an external calls.
- https://github.com/fusionpbx/fusionpbx/commit/dc858710095716fa8bf9bd877de49e7926b37274

PR #5344, opened 2020-07-12
Follow me: set domain_uuid as a channel variable

greenbea:
``${domain_uuid}``  is needed in the dialplan for a few dialplans (example: call-limit is using it for the resource)


markjcrane:
Looks good

PR #5347, opened 2020-07-13
 remove CDR Statistics from default menu for users

greenbea:


greenbea:
Users don't have view permission by default to CDR statistics

PR #5350, opened 2020-07-15
Add samesite=strict to session cookie

greenbea:
* Additional security layer for CSRF attacks is to use the same-site attribute in the cookie
more info can be at found https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#samesite-cookie-attribute

* It's only supported in PHP version 7.3 and up

### DIFF
--- a/core/install/install.php
+++ b/core/install/install.php
@@ -41,6 +41,9 @@
 
 //start the session
 	//ini_set("session.cookie_httponly", True);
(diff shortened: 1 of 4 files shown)


greenbea:
I thinking it should maybe be set to lax to allow GET requests from external links

PR #5359, opened 2020-07-21
Changed error message for password length

tony1661:
I've updated the wording for the error when a password is not long enough. I kept getting this error and found it wasn't descriptive enough and took me some time to realize what the actual error was. I also added the translations in Spanish and French.


markjcrane:
This looks great thanks for contributing this improvement.

PR #5363, opened 2020-07-27
Update {$mac}.xml

si458:
fix lcd_contrast in spa941


si458:
thanks for this,
however i removed the Back_Light_Timer line because the SPA941 doesnt support this line
and it also doesnt appear anyone in the config i pulled from our SPA941 devices we use everywhere,

PR #5373, opened 2020-07-31
feature add mask/unmask recordings in real time. white space cleanup.

gabada:
This is a feature for PCI compliance when taking credit card numbers or any personal information over a recorded phone call.
This allows the user to dial *5 before taking the personal information to play white noise instead of recording and then *6 after the information is taken to start recording the conversation again.

This solution is better than pausing and unpausing the recording because you can tell how long it took the user to gather the personal information. The recording is still in progress but white noise is being recorded instead of the conversation.

These changes allow recording masking/unmasking on inbound, outbound, and ring group calls.

1 bug is that if you dial a DID the outside party can mask/unmask but I think its a nonissue since its an obscure binding and the callee would have to know about it.

I also removed some white space in index.lua of ring groups.

NOTE: This change does not affect call center.


greenbea:
There should be a way to bind the digits only for the registered users not only for this feature, but everywhere where fusion binds digits

gabada:
Yes, that would be great but the issue is A leg vs B leg.

demonspork:
It would need conditions based upon the call direction to determine if the a leg or the b leg is the internal caller.

PR #5417, opened 2020-08-21
[4.5] Gives IVR Loopback support 1/2

daniel-lucio:
After reading the destinations class, I found that you have an issue comparing against $apps[$x]['destinations'][$y]['name'] (method destination::select()), this is the root cause why loopback is not shown although you already have the definition in extensions/app_config.php.

'name' index needs to be different so users can have both options, direct transfer and loopback bridge.

Also, the 'ivr' index was missing (just copy and paste).

This patch allows a user to have the option of selecting a direct transfer or loopback bridge when adding a menu option.

### DIFF
--- a/app/extensions/app_config.php
+++ b/app/extensions/app_config.php
@@ -47,7 +47,7 @@
 		$y++;
 		$apps[$x]['destinations'][$y]['type'] = "sql";
 		$apps[$x]['destinations'][$y]['label'] = "loopback";
-		$apps[$x]['destinations'][$y]['name'] = "extensions";
+		$apps[$x]['destinations'][$y]['name'] = "loopback_extensions";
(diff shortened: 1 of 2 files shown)


markjcrane:
Loopback is used when transfer is not possible. In the case of IVR Menu options or exist action transfer is available and loopback should not be used.

markjcrane:
If you want this committed you will have to explain why its even useful when transfer is available there is not reason I can think of to use loopback. Loopback has a lot of downsides and makes a mess of call detail records and often breaks reporting.

daniel-lucio:
I will add a default setting to allow or not this kind of routing. About the reasons:

1. Gives the user more options
2. Specific cases when an inbound route goes to an IVR, and IVR routes to an external source, let's say you want to bill both (in and out), simple bridges won't allow you. (this scenario is very common, especially when users route calls to cellphone)

PR #5418, opened 2020-08-21
[4.5] IVR loopback support 2/2

daniel-lucio:
Through a new default setting, the direct dial may be switched to use loopback or direct transfer


markjcrane:
Loopback is used when transfer is not possible. In the case of IVR Menu options transfer is available and loopback should not be used.

PR #5431, opened 2020-08-30
Update class.phpmailer.php

phoneben:


phoneben:
making templates work with utf  8 characters based on https://stackoverflow.com/questions/2491475/phpmailer-character-encoding-issues and tested in real prudution

PR #5441, opened 2020-09-08
BugFix [master] - ring_groups\index.lua

mafoo:
change to use functions.send_mail so system from address is taken into account


markjcrane:
Thanks for contributing... Looks like a good improvement.

PR #5443, opened 2020-09-08
Update index.lua

graycom-git:
Update to use effective_caller_id_name when prefixed on inbound routes as the caller_id_name when recording voicemail details in database


graycom-git:
Update to use effective_caller_id_name when prefixed on inbound routes as the caller_id_name when recording voicemail details in database

markjcrane:
This looks like a good thanks for sharing it.

PR #5458, opened 2020-09-15
Hide global dialplans. Show to superadmin group that has dialplan_all permission

FozzyGo4:
If permissions to dialplans is given to anyone other than superadmin then they should not be able to see GLOBAL dialpans unless they have dialplan_all permission as well.

### DIFF
--- a/app/dialplans/dialplans.php
+++ b/app/dialplans/dialplans.php
@@ -140,9 +140,11 @@
 		$sql .= "where true ";
 	}
 	else {
-		$sql .= "where (domain_uuid = :domain_uuid) ";
 		if (if_group("superadmin")) {
-			$sql .= " or (domain_uuid is null)";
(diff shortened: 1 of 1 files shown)


markjcrane:
Only superadmin should have permissions to dialplans beyond view permission. If anyone gives any customer  group permission to directly add or update dialplans then security has been compromised. Users with this control can send calls to other domains. Intercept calls from other domains, eavesdrop, steal caller ID from other domains and lot more.

FozzyGo4:
Agreed but how then does a reseller admin add outbound dialplans and or custom dialpans for their customers?

FozzyGo4:
This commit also does not suggest to give add /edit / delete permissions to dialplans, just removes the viewing of global dialplans from anyone not a superadmin.

PR #5486, opened 2020-09-25
Update send_email.lua

junction1153:
In reference to post: https://www.pbxforums.com/threads/mailbox-name-on-vm-to-email-template.4498/ (part A)


markjcrane:
Any website links could go down some years in the future. So prefer adding notes to the pull request instead of a link.

PR #5504, opened 2020-10-02
Update app_config.php

jamesorose:
yealink_trust_certificates should always be default FALSE. This messes up provisioning.


markjcrane:
That breaks concept of secure by default sorry you had trouble with it. But security is important.

PR #5506, opened 2020-10-03
Update {$mac}.xml

pennytone:
line 6262 - auto upgrade based on hour of day. set to 2
line 6294 - Firmware Upgrade Confirmation set to 0 no
this will allow unattended firmware upgrade.


markjcrane:
P194  Should likely use a provisioning variable so that people can choose if they want upgrade to be based on hour or not.

PR #5507, opened 2020-10-04
Voicemail : Not Found Extension (ability to transfer to a destination (extension/ivr/etc) when extension is unavailable and no voicemail is configured

JosephWBailey:
Voicemail : Not Found Extension

When an extension is unavailable and no voicemail is configured, there is an option to transfer the caller to a defined destination (extension/ivr/etc).

To enable/disable this, add a default setting of type: text, named (subcategory) not_found_extension in Advanced > Default Settings > Voicemail category and set the value field to the destination to suit your needs.

This setting can be used in conjunction with not_found_message if desired.

### DIFF
--- a/app/scripts/resources/scripts/app/voicemail/index.lua
+++ b/app/scripts/resources/scripts/app/voicemail/index.lua
@@ -219,6 +219,13 @@
 					end
 				end
 
+				not_found_extension = '';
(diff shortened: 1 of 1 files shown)


JosephWBailey:
At some point my GitHub accidently got toasted, so this pull request is showing as from user ghost.  If this is an issue I will be happy to commit again.

PR #5521, opened 2020-10-09
Update {$mac}.xml

pennytone:
<P8352>5</P8352>

When this is set, it will override the users choice of ring volume. My customer sets some phones to low by pressing the - volume button. By next morning, the volume is back to loud because the provision file sets <P8352>5</P8352> thus overwriting the users choice of volume.

### DIFF
--- a/resources/templates/provision/grandstream/gxp2170/{$mac}.xml
+++ b/resources/templates/provision/grandstream/gxp2170/{$mac}.xml
@@ -8256,7 +8256,7 @@
     <!-- # Speaker Ring Volume -->
     <!-- # Number: 0-7. Default is 5. -->
     <!-- # Mandatory -->
-    <P8352>5</P8352>
+    <!-- <P8352>5</P8352> -->
 
     <!-- # Notification Tone Volume -->
     <!-- # Number: 0-7. Default is 5. -->


demonspork:
For the sanity of the rest of us, could you put this change in all of the gxp templates instead of just the 2170?

Len-PGH:
Mandatory Pcodes commented out can leave the phone in a boot loop. I'll make a variable that can be set 0-7

demonspork:
Most of these variables that are labeled mandatory are a lie. Can we confirm this is actually mandatory? Having this variable set makes it completely impossible for the customer to set their own ring tone volume. This is also true of several other user preference settings including wallpaper source, where the template has them tagged as mandatory and I have commented them out for years.

demonspork:
Here is the core issue: You are interpreting     to mean that you have to include the variable in the template. This is patently untrue.

The Grandstream GXP phones, upon factory default, generate a valid value for all settings that are tagged as     in our templates.

When the phone downloads an XML configuration, it takes a look at all of the settings defined in the template and either adds or overwrites those values in its internal configuration. Values that are not defined in the XML file are left at their existing values in the phone.

This means that values that are marked     can be safely left out completely or commented out.

The only danger for the     is setting them to an empty value. Thus, the only invalid version of this Pcode is: `<P8352></P8352>` which when the phone reads that, it overrides the phone's existing valid value for the phone volume with a NULL/Empty value and is thus invalid and can cause other glitches.

If we simply never define <P8352>5</P8352> at all in the XML, the phone processes the configuration and never sees P8352 and leaves the current value alone. Safe and happy and with the user's preferred volume.

That being said, if we want an admin to be able to define it we create a variable, check that the variable exists, and leave out the Pcode if it is not set, like this:
``
{if isset($grandstream_speaker_ring_volume)}
  <P8352>{$grandstream_speaker_ring_volume}</P8352>
{/if}
``
notice there is no "else" statement because the Pcode is not required to be defined, it simply requires a VALID definition if it is present. Which is the true meaning of Mandatory in these Grandstream templates.

markjcrane:
This pull request would make one person happy and other people unhappy so I believe the isset is the best option for this and the variable so that it can be controlled and this should accommodate everyone's needs.

markjcrane:
I accepted this pull request. Because this is how I was planning to change the original pull request.
https://github.com/fusionpbx/fusionpbx/pull/5528

PR #5526, opened 2020-10-10
Update {$mac}.xml

Len-PGH:


demonspork:
This does not allow the user to manipulate their own ring volume. The correct solution is 

```
{if isset($grandstream_speaker_ring_volume)}
  <P8352>{$grandstream_speaker_ring_volume}</P8352>
{/if}
```

because if the Admin says "I want the users to have control over their ring volume" the admin can simple disable grandstream_speaker_ring_volume in the default settings. If the user wants full control over the ring volume, they can leave the setting enabled and define a ring volume.

This solution provides a wider degree of possibilities, it works for admins who want to control the ring volume and for admins who want the users to have this control.

markjcrane:
I agree with the isset for this case.

PR #5546, opened 2020-10-18
Update app_config.php

phoneben:
add default permissions on ring_group_cid_number_prefix ring_group_cid_name_prefix to super admin


phoneben:
https://github.com/fusionpbx/fusionpbx/pull/5045
due to this request

markjcrane:
This can be added by Using Group Manager setting a permission on the superadmin group and then setting that permission to protected. This way it doesn't need to be provided by default.

PR #5547, opened 2020-10-19
GXP17xx Template Improvements

demonspork:
Various basic variables were missing like transport type and sip_port and basic things were wrong like Accept Incoming SIP from Proxy Only was disabled.
This template P codes are almost identical to the GXP16XX except that the loops for button programming don't seem to work with it at least with my limited testing, so they are left out for now.

Tested and working with firmware 1.0.1.89


markjcrane:
Looks good thanks for sharing these changes.

PR #5555, opened 2020-11-03
Add ability to set default accountcodes

greenbea:


markjcrane:
This looks good just think we may want a better place for the get_accountcode function.

markjcrane:
Reviewed this again if the conflicts are fixed I will accept this pull request.

PR #5566, opened 2020-11-05
Fix Device Provisioning and Feature Sync across SIP Profiles

winsock:
# Context
If you have multiple registrations that are also across SIP profiles the provision button under the `device_edit.php` page and the feature syncing will not work since it calls `sofia_contact` and only uses the first profile found. This causes the feature keys to not sync or devices to not re-provision in some cases.

# Overview
- Modify `send_notify` in `feature_event_notify.php` to use `preg_match_all` to grab all of the profiles and loop the event sending for each unique profile.
- Move the profile code to `cmd.php` in the devices app and like what was completed in `feature_event_notify.php` loop over each unique profile and send the command.
  - Also removes the newly unused code from `device_edit.php`
- Renames `get_profile` to `get_profiles` in `feature_event_notify.lua` which either returns nil for no profiles or a array of unique profiles.
- Each lua function that used to take the single sip_profile was updated to take this array and send events to each profile the user is part of.

### DIFF
--- a/app/calls/resources/classes/feature_event_notify.php
+++ b/app/calls/resources/classes/feature_event_notify.php
@@ -43,38 +43,40 @@ include "root.php";
 
 	//feature_event method		
 		public function send_notify() {
-
-				$fp = event_socket_create($_SESSION['event_socket_ip_address'], $_SESSION['event_socket_port'], $_SESSION['event_socket_password']);
-				if ($fp) {
-					//get the sip profile name
-						$command = "sofia_contact */".$this->extension."@".$this->domain_name;
-						$contact_string = event_socket_request($fp, "api ".$command);
-						if (substr($contact_string, 0, 5) == "sofia") {
-							$contact_array = explode("/", $contact_string);
-							$sip_profile_name = $contact_array[1];
-						}
-						else {
-							$sip_profile_name = 'internal';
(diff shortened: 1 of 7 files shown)


winsock:
I also had another issue that I wanted to fix in this MR that I came across but I am not sure if it was needed just because of my current server setup.

For the device_edit.php provision button it was not working on my secondary domain since I am not using domain filtering and routing all registrations for secondary domains to their own SIP profiles on non-standard ports that force the correct domain. The provision button would try to send the SIP NOTIFY to the extension on the primary domain instead of the secondary domain. I was able to fix it with the following change:
```php
		echo button::create(['type'=>'button','label'=>$text['button-provision'],'icon'=>'fax','style'=>$button_margin,'link'=>PROJECT_PATH."/app/devices/cmd.php?cmd=check_sync"."&user=".urlencode($user_id)."@".urlencode($_SESSION['domains'][$domain_uuid]['domain_name'])."&domain=".urlencode($_SESSION['domains'][$domain_uuid]['domain_name'])."&agent=".urlencode($device_vendor)]);
```
If you think that this is safe for general use I will add it to this MR.

winsock:
Updated to add the same logic to the LUA part of the feature.

PR #5583, opened 2020-11-17
Update send_email.lua to choose correct email template

Andy-Seattle:
To handle scenarios where transcribe_enabled is true BUT voicemail_transcription_enabled is not true for specific voicemail boxes. Need to be able to switch between the email templates.


markjcrane:
Thanks for sharing this improvement!

PR #5584, opened 2020-11-17
Minify Provisioning XML

winsock:
# Context
For device provisioning templates that are valid XML there can be a substantial size reduction by removing all whitespace and comments from served XML files.

## Example in size reduction
GRP2613 Template (8% of original):
- Before: ‘index.html?mac=000b82354b51&pretty=true’ saved [168157/168157]
- After: ‘index.html?mac=000b82354b51’ saved [14837/14837]

Custom Template (42% of original):
- Before: ‘index.html?mac=c074ad288040&pretty=true’ saved [426295/426295]
- After: ‘index.html?mac=c074ad288040.2’ saved [178341/178341]

# Overview
- Add a new request parameter of `pretty` and if it is not empty do not minify output
- Add a new setting called `minify_xml` to allow disabling this behavior.
- Using DOMDocument parse the output of the provisioning template and minify the XML if requested and was parsed as valid XML.
- Remove the call to `simplexml_load_string` for determining the Content-Type and use the result of the `loadXML` call.

### DIFF
--- a/app/devices/device_edit.php
+++ b/app/devices/device_edit.php
@@ -683,7 +683,7 @@
 			if ($_SERVER['HTTPS'] == 'on') { $_SERVER['HTTP_PROTOCOL'] = 'https'; }
 			if ($_SERVER['SERVER_PORT'] == '443') { $_SERVER['HTTP_PROTOCOL'] = 'https'; }
 		}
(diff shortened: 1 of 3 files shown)


demonspork:
I particularly like getting the DOMDocument code working on the provisioning files. It opens up some possibilities for defining settings without editing the templates, it would allow us to have either an Array of P values that get added/updated to the DOMdocument pretty easily or just a setting format like "grandstream_p1455" to set P1455 to a specific setting

PR #5597, opened 2020-11-18
Only send MWI notifications for enabled voicemail boxes.

winsock:
# Context
FusionPBX would continue replying to `MESSAGE_QUERY` events or send `MESSAGE_WAITING` events periodically even when voicemail boxes were disabled. This can also cause issues if you have another system responding to MESSAGE_QUERY events for the extensions that have their voicemail boxes disabled.

# Overview
- For the main MWI notifier that just sends notifications every X seconds add an extra clause to check that `v.voicemail_enabled = 'true'`
- For the MWI subscription responder add the same check in the UUID lookup.
- Modify the logic and SQL to instead of embedding the uuid lookup SQL as a subquery to just use the value queried earlier.
  - This is needed otherwise the `(SELECT COUNT(*) ...) AS ...` statements would return a valid row with zeros for even a disabled voicemail account causing the `MESSAGE_WAITING` event to be sent to FreeSWITCH.

### DIFF
--- a/app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
+++ b/app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
@@ -13,18 +13,18 @@ local vm_message_count do
 
 local vm_to_uuid_sql = [[SELECT v.voicemail_uuid
 FROM v_voicemails as v inner join v_domains as d on v.domain_uuid = d.domain_uuid
-WHERE v.voicemail_id = :voicemail_id and d.domain_name = :domain_name]]
+WHERE v.voicemail_enabled = 'true' and v.voicemail_id = :voicemail_id and d.domain_name = :domain_name]]
 
 local vm_messages_sql = [[SELECT
 ( SELECT count(*)
 	FROM v_voicemail_messages
-	WHERE voicemail_uuid = %s
+	WHERE voicemail_uuid = :voicemail_uuid
 	AND (message_status is null or message_status = '')
 ) as new_messages,
 
 ( SELECT count(*)
 	FROM v_voicemail_messages
-	WHERE voicemail_uuid = %s
+	WHERE voicemail_uuid = :voicemail_uuid
 	AND message_status = 'saved'
 ) as saved_messages
 ]]
@@ -42,38 +42,32 @@ function vm_message_count(account, use_cache)
 	local dbh = Database.new('system')
 	if not dbh then return end
 
+	-- Get the UUID from the cache if enabled/supported
 	local uuid
 	if use_cache and cache.support() then
-		local uuid = cache.get('voicemail_uuid:' .. account)
-		if not uuid then
-			uuid = dbh:first_value(vm_to_uuid_sql, {
-				voicemail_id = id, domain_name = domain_name
-			})
+		uuid = cache.get('voicemail_uuid:' .. account)
+	end
 
-			if uuid and #uuid > 0 then
-				cache.set('voicemail_uuid:' .. account, uuid, 3600)
-			end
+	-- If no UUID cached or if cache is not enabled/supported get the UUID from the database
+	if not uuid then
+		uuid = dbh:first_value(vm_to_uuid_sql, {
+			voicemail_id = id, domain_name = domain_name
+		})
+
+		if uuid and #uuid > 0 and use_cache and cache.support() then
+			cache.set('voicemail_uuid:' .. account, uuid, 3600)
 		end
 	end
 
+	-- Get the count of unread and read messages from the database if there is a valid voicemail account
 	local row
 	if uuid and #uuid > 0 then
-		local sql = string.format(vm_messages_sql, ":voicemail_uuid", ":voicemail_uuid")
-		row = dbh:first_row(sql, {voicemail_uuid = uuid})
-	else
-		local uuid_sql = '(' .. vm_to_uuid_sql .. ')'
-
-		local sql = string.format(vm_messages_sql,
-			uuid_sql, uuid_sql
-		)
-
-		row = dbh:first_row(sql, {
-			voicemail_id = id, domain_name = domain_name
-		})
+		row = dbh:first_row(vm_messages_sql, {voicemail_uuid = uuid})
 	end
 
 	dbh:release()
 
+	-- This condition can be hit if the voicemail box is either disabled or non-existent
 	if not row then return end
 
 	return row.new_messages, row.saved_messages
--- a/app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi.lua
+++ b/app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi.lua
@@ -76,6 +76,7 @@
 				FROM v_voicemail_messages as m, v_voicemails as v, v_domains as d
 				WHERE v.voicemail_uuid = m.voicemail_uuid
 				AND v.domain_uuid = d.domain_uuid
+				AND v.voicemail_enabled = 'true'
 				GROUP BY v.voicemail_id, v.voicemail_uuid, v.domain_uuid, d.domain_name;]];
 			if (debug["sql"]) then
 				freeswitch.consoleLog("notice", "[voicemail] SQL: " .. sql .. "\n");


app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
```diff
@@ -42,38 +42,32 @@ function vm_message_count(account, use_cache)
...
+	end
+
+	-- If no UUID cached or if cache is not enabled/supported get the UUID from the database
+	if not uuid then
+		uuid = dbh:first_value(vm_to_uuid_sql, {
+			voicemail_id = id, domain_name = domain_name
+		})
+
+		if uuid and #uuid > 0 and use_cache and cache.support() then
+			cache.set('voicemail_uuid:' .. account, uuid, 3600)
```
moteus, line 58: You trying set the cache even if it not allowed

app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
```diff
@@ -13,18 +13,18 @@ local vm_message_count do
...
 local vm_to_uuid_sql = [[SELECT v.voicemail_uuid
 FROM v_voicemails as v inner join v_domains as d on v.domain_uuid = d.domain_uuid
-WHERE v.voicemail_id = :voicemail_id and d.domain_name = :domain_name]]
+WHERE v.voicemail_enabled = 'true' and v.voicemail_id = :voicemail_id and d.domain_name = :domain_name]]
 
 local vm_messages_sql = [[SELECT
 ( SELECT count(*)
 	FROM v_voicemail_messages
-	WHERE voicemail_uuid = %s
+	WHERE voicemail_uuid = :voicemail_uuid
```
moteus, line 21: ODBC does not supports params in such queries. This is why i did not use them here

app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
winsock: I do not use LUA daily but I would assume that `uuid and #uuid > 0 and use_cache and cache.support()` would resolve to false if any one of the four conditions is fase? That means the cache code would not be hit unless I am mistaken.

I'll add further logging messages to double check this.

app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
winsock: But the UUID lookup code uses parameterized queries originally, it was inserted as a sub-query if there was no cached UUID. If we are working with a DB that does not support parameterized queries they should be removed from all the SQL queries then?

moteus: Sorry. I did not see second part somehow

winsock: No worries!

app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
winsock: Results of testing seem to be fine. cachetype was set to memcache in default settings, I used this for testing:
```lua
		if uuid and #uuid > 0 and use_cache and cache.support() then
			log.notice('cache set hit')
			cache.set('voicemail_uuid:' .. account, uuid, 3600)
		else
			log.notice('cache set miss')
		end
```

Results with use_cache = false and memcached loaded:
```
2020-11-18 15:10:10.202312 [NOTICE] switch_cpp.cpp:1447 [mwi] cache set miss
2020-11-18 15:10:10.202312 [NOTICE] switch_cpp.cpp:1447 [mwi] voicemail 100@10.100.106.108 has 0/0 message(s)
```

Results with use_cache = true and memcached loaded:
```
2020-11-18 15:13:38.008614 [NOTICE] switch_cpp.cpp:1447 [mwi] cache set hit
2020-11-18 15:13:38.008614 [NOTICE] switch_cpp.cpp:1447 [mwi] voicemail 100@10.100.106.108 has 0/0 message(s)
```

Results with use_cache = true and memcached unloaded:
```
2020-11-18 15:30:03.073218 [NOTICE] switch_cpp.cpp:1447 [mwi] cache set miss
2020-11-18 15:30:03.073218 [NOTICE] switch_cpp.cpp:1447 [mwi] voicemail 100@10.100.106.108 has 0/0 message(s)
```

app/scripts/resources/scripts/app/voicemail/resources/scripts/mwi_subscribe.lua
moteus: My preview comment was about cache.
About this parameter. Original FS dooes not use native query parameters and fusion jus umelates them by string replace . So, ypur solution woll work.
But I use patched version of FS which uses native parameters and PG ODBC driver does not support params in queries with the `union`

winsock: Understood, resolved the cache conversation instead of this one.

winsock: I am still a bit confused on what the issue is regarding the SQL queries and ODBC. Do you want me to switch everything to use string replacement? The original SQL queries use parameters. Here is the original code:

```lua
local vm_to_uuid_sql = [[SELECT v.voicemail_uuid
FROM v_voicemails as v inner join v_domains as d on v.domain_uuid = d.domain_uuid
WHERE v.voicemail_id = :voicemail_id and d.domain_name = :domain_name]]

local vm_messages_sql = [[SELECT
( SELECT count(*)
	FROM v_voicemail_messages
	WHERE voicemail_uuid = %s
	AND (message_status is null or message_status = '')
) as new_messages,
( SELECT count(*)
	FROM v_voicemail_messages
	WHERE voicemail_uuid = %s
	AND message_status = 'saved'
) as saved_messages
]]
```

The vm_messages_sql would be transformed to then include the `:voicemail_uuid` if there was a cached UUID:
```lua
		local sql = string.format(vm_messages_sql, ":voicemail_uuid", ":voicemail_uuid")
```
Resulting in a query that looks like this:
```sql
SELECT
( SELECT count(*)
	FROM v_voicemail_messages
	WHERE voicemail_uuid = :voicemail_uuid
	AND (message_status is null or message_status = '')
) as new_messages,
( SELECT count(*)
	FROM v_voicemail_messages
	WHERE voicemail_uuid = :voicemail_uuid
	AND message_status = 'saved'
) as saved_messages
```

And if there was no cached UUID it would replace the "%s" with the UUID lookup query:
```lua
		local uuid_sql = '(' .. vm_to_uuid_sql .. ')'

		local sql = string.format(vm_messages_sql,
			uuid_sql, uuid_sql
		)
```
Resulting in a query that looks like this:
```sql
SELECT
( SELECT count(*)
	FROM v_voicemail_messages
	WHERE voicemail_uuid = (SELECT v.voicemail_uuid
FROM v_voicemails as v inner join v_domains as d on v.domain_uuid = d.domain_uuid
WHERE v.voicemail_id = :voicemail_id and d.domain_name = :domain_name)
	AND (message_status is null or message_status = '')
) as new_messages,
( SELECT count(*)
	FROM v_voicemail_messages
	WHERE voicemail_uuid = (SELECT v.voicemail_uuid
FROM v_voicemails as v inner join v_domains as d on v.domain_uuid = d.domain_uuid
WHERE v.voicemail_id = :voicemail_id and d.domain_name = :domain_name)
	AND message_status = 'saved'
) as saved_messages
```

Both possible queries in the current version of the code in master use the query parameters. My changes just split the queries into two distinct lookups to allow for exiting early if the voicemail box is disabled in FusionPBX.

I can replace all of the parameters in the queries with "%s" and use `string.format` to substitute if that is what you think is best. It just deviates from the existing code in master and I wanted to match what was being used already originally.

winsock: Isn't your use case covered by database.lua to begin with? From that code it supports emulating support for parameters and that should happen before hitting the ODBC driver correct?

```lua
    if params then
      -- backend supports parameters natively
      if self.__base.parameter_query then
        return self.__base.parameter_query(self, sql, params, callback)
      end

      -- use emulation of parametes
      local err
      sql, err = self:_apply_params(sql, params)
      if not sql then
        log.errf('can not bind parameter: %s', tostring(err))
        return nil, err
      end
    end
```

PR #5609, opened 2020-11-24
Call Center Loopback voice record

sergenkabakci:
When I forward calls in this way, no voices recorded. Can you fix it?

loopback/100/domain.com


markjcrane:
Please fix your pull request so that it only show the changes you have made.

markjcrane:
If you are asking for help fixing something don't ask here through pull requests. Instead submit requests from an account in www.fusionpbx.com or ask in irc on freenode #fusionpbx.

PR #5612, opened 2020-11-30
Improve the valet_park_auto dialplan to support auto out and whole lot BLF monitoring

winsock:
# Context
There were a couple of features of valet_park that were not exposed before by the default FusionPBX dialplan. This adds support for those features.

# Overview
- If the Referred-By header is set call the normal `auto in` code. If it isn't set that means someone called it directly, call `auto out` and attempt to pop the calls in the lot in FIFO order.
- This also changes the parking lot name from `park` to `5900` this allows phones to monitor `park+5900` to get the status of the whole parking lot. Before you would have to monitor `park+park` to get this functionality.
  - https://freeswitch.org/confluence/display/FREESWITCH/mod_valet_parking#mod_valet_parking-UsingPresence

### DIFF
--- a/app/dialplans/resources/switch/conf/dialplan/470_valet_park_auto.xml
+++ b/app/dialplans/resources/switch/conf/dialplan/470_valet_park_auto.xml
@@ -1,0 +1,9 @@
+<context name="{v_context}">
+	<extension name="valet_park_auto" number="park+5900" continue="false" app_uuid="c192ee50-084d-40d8-8d9a-6959369382c8" enabled="false" order="470">
+		<condition field="destination_number" expression="^(?:(?:park\+\*?)|(?:\*))(5900)$"/>
+		<condition field="${sip_h_Referred-By}" expression="sip:(.*)@.*" break="never">
+			<action application="valet_park" data="5900@${domain_name} auto in 5901 5999"/>
+			<anti-action application="valet_park" data="5900@${domain_name} auto out 5901 5999"/>
+		</condition>
+	</extension>
+</context>
--- a/app/dialplans/resources/switch/conf/dialplan/470_valet_park_in.xml
+++ b/app/dialplans/resources/switch/conf/dialplan/470_valet_park_in.xml
@@ -1,7 +1,0 @@
-<context name="{v_context}">
-	<extension name="valet_park_in" number="park+*5900" continue="false" app_uuid="c192ee50-084d-40d8-8d9a-6959369382c8" enabled="false" order="470">
-		<condition field="destination_number" expression="^(?:(?:park\+\*?)|(?:\*))(5900)$">
-			<action application="valet_park" data="park@${domain_name} auto in 5901 5999"/>
-		</condition>
-	</extension>
-</context>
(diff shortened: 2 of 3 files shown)


winsock:
With the changes in #5611 valet_park_auto could be enabled by default if we want now.

markjcrane:
My concern on this 5900 the purpose of the * is used like this *5900 is to make sure the number doesn't conflict with numbers the admin uses for the system. Makes it to tempting to change the number to something like 00 - 99 or 1 to 99.

demonspork:
> 
> 
> My concern on this 5900 the purpose of the * is used like this *5900 is to make sure the number doesn't conflict with numbers the admin uses for the system. Makes it to tempting to change the number to something like 00 - 99 or 1 to 99.

Isn't this still safe because the dialplan still only contains conditions that match park+*5900, park+5900, or *5900. The regex changes don't match on just 5900 as far as I can tell in testing it.

winsock:
Correct, it should not match on plain 5900. Here is an example:
https://regex101.com/r/hYKr7N/1

The #5611 MR to match this is the same thing:
https://regex101.com/r/qpSvYz/1

I have been using this in production now for 5 months or so.

markjcrane:
Thanks for the clarification did you also test this to make sure it still prevented double parking?

markjcrane:
Merged this but its not backwards compatible anyone that updates and then updates their dialplans will break their Park BLF that are using *5900 to *5999
- They will have to remove the * prefix for BLF to work again. So you preserved * making it optional but because of removing the * it will break their BLF for anyone that programmed phone with a star like *5901.
- Also anyone that uses extensions 5900 - 5999 for something other than Park will have a conflict with these new dialplans.

For what benefits?

markjcrane:
Would be nice if we could talk on the phone Tuesday and talk it over.

winsock:
Hi Mark,

We can certainly talk about this over the phone tomorrow. It took me a bit to remember why removing the asterisk was needed and it's due to valet park's auto out and in functionality. It will not work with non-numeric parking lots. Meaning that you could not directly access auto-in parking lots from the direct access codes. 

We could have a split system where we have the "*5900" lot with *5901-*5999 parking spots. And then a "5900" lot with 5901-5999 parking spots that support auto-in and out. 

It just seems a bit weird to have two separate lots.

About 5900-5999 extensions unless I have a typo in the regex somewhere those will not match these dial plans and will not be an issue. It will only match park+59xx, park+*59xx, and *59xx

winsock:
Just to be a bit more specific it's these lines that will not work if you put *5901 *5999 in them
```
<action application="valet_park" data="5900@${domain_name} auto in 5901 5999"/>
<anti-action application="valet_park" data="5900@${domain_name} auto out 5901 5999"/>
```

winsock:
Another possible idea to resolve the incompatibility while maintaining support for auto in, auto out, and whole lot BLF would possibly be a LUA hook to listen then duplicate the BLF events to park+*59xx

winsock:
@markjcrane If you want to have a phone call about this today let me know. I have a meeting in 7 minutes but after that I can be free. Just message me on IRC.

PR #5615, opened 2020-12-07
Order by start_epoch to Speed Up Query Execution

emaktech:
In some cases, this seems to speed up xml_cdr loading times by ~10x.

It appears one big cause of this is the final ORDER_BY statements are very slow in PostgreSQL for timestamp fields. Ordering by start_epoch field improves query execution time in a dramatic way and should result in the same ordering.

### DIFF
--- a/app/xml_cdr/xml_cdr_inc.php
+++ b/app/xml_cdr/xml_cdr_inc.php
@@ -187,7 +187,7 @@
 	}
 
 //create the sql query to get the xml cdr records
-	if (strlen($order_by) == 0) { $order_by  = "start_stamp"; }
+	if (strlen($order_by) == 0) { $order_by  = "start_epoch"; }
 	if (strlen($order) == 0) { $order  = "desc"; }
 
 //set a default number of rows to show


demonspork:
I did some more research and testing on this. It seems that internally in Postgres, a "numeric" column like the start_epoch is stored as a string of 16 bit integers in base 10000 prefixed by a few bytes with the length and precision, which creates overhead dealing with tuples and blocks in the background.

The timestamp data type is stored as a 64 bit integer, so order by using the start_stamp should be significantly faster, and in all of my testing it is.

### Testing 

`explain analyze SELECT * FROM v_xml_cdr ORDER BY start_stamp` executed in 19s on my test server with ~2 million rows

`explain analyze SELECT * FROM v_xml_cdr ORDER BY start_epoch` executed in over 40s on the same server.

### Notes

I also played around with indexing the start_stamp and start_epoch columns, but when that index exists, both queries nearly doubled so I quickly ditched that idea.

I also tried `SET work_mem TO '2 GB'` and that decreased the start_stamp query down to under 10 seconds, so a big part of the xml_cdr performance issues might be simply we don't have enough working memory for the queries we are doing. When using `explain analyze` it shows the sort method, and before increasing the work_mem it showed `Sort Method: external merge  Disk: 859920kB` but that changed to 'Sort Method: quicksort  Memory: 1266896kB', which seems to be much faster.

emaktech:
@demonspork I agree, this should be reverted as it doesn't actually change performance in the end.

The history of why this change was made was because while troubleshooting performance issues on the CDR page, the sort field change resulted in a dramatic speed increase.

But in later testing with @markjcrane , I realized that I had forgotten to index the start_stamp field while start_epoch was indexed. Surprisingly, the difference in query time on the sort on this page was something like 30,000x (yes, 5 orders of magnitude!) faster with index than without.

Once the other field was indexed, there is essentially 0 performance delta between the two data types.

I just ran your test query on our production system and got these results:

```
Index  Scan using start_epoch on v_xml_cdr  (cost=0.43..669948.92 rows=7547620  width=948) (actual time=0.077..2963.269 rows=7579286 loops=1)
--
Planning time: 2.504 ms
Execution time: 3135.082 ms
```

```
Index  Scan using start_stamp on v_xml_cdr  (cost=0.43..660710.25 rows=7547636  width=948) (actual time=0.362..2946.913 rows=7579289 loops=1)
--
Planning time: 3.661 ms
Execution time: 3116.588 ms
```

So yes again I agree, this should be reverted as, at best, it does nothing. The final moral of the story is index the important fields in your big tables!

PR #5619, opened 2020-12-09
Fix accountcode not getting set on loopback

greenbea:
if an extension calls a local extension or a ring group that goes out to an external destination accountcode won't get set because of loopback being used. Fixed by exporting accountcode


greenbea:
I'll resubmit a different fix, this not good as the one that sets the follow me needs to be billed not the caller

PR #5674, opened 2020-12-13
[4.5] Gives priority to key authentication

daniel-lucio:
The current logic when authenticates verifies username variable. If that happens, it ignores the key variable (regardless if it is present). When this happens, authentication forcibly goes on user-password variables.
This could bring issues when the username and password variables are not used for authentication but editing/creating some elements. Since they are a very common name (more than one form uses in as an input field that is not authentication).

This patch gives priority to the key authentication if the key variable is present. (picked up explanation from 4.4 pull request - Mark)

### DIFF
--- a/resources/check_auth.php
+++ b/resources/check_auth.php
@@ -62,12 +62,12 @@
 
 		//validate the username and password
 			$auth = new authentication;
-			if (isset($_REQUEST["username"]) && isset($_REQUEST["password"])) {
-				$auth->username = $_REQUEST["username"];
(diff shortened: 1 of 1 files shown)


markjcrane:
Please provide an explanation for why you want this change.

markjcrane:
It seems like this would have extremely small difference in resources so minuscule as to be more of a preference than a benefit.

markjcrane:
I prefer the current priority given to users that login rather than the few that use the key.

PR #5685, opened 2020-12-23
Fix CDR Search Result order with start_epoch

demonspork:
When searching in xml_cdr.php the results would forget to sort by the Start time. It seems that the start_epoch change xml_cdr_inc.php caused this in PR #5615, this is the fix.


demonspork:
See my notes on #5615, which shows that this is much slower so the original change that required this bug fix should be reverted.

PR #5688, opened 2020-12-27
[4.5] Many Installation fixes & DB compatibility

daniel-lucio:
I will send more PRs as this is getting accepted.

The first issue I found is that the installer has hardcoded many absolute paths, if you want to install fusion under http://x.x.x.x/fusionpbx/ it is not possible. css.php is broken (this could be considered a cosmetic issue, it just made installation harder to read). This patch adds /fusionpbx/ as well, the browser will look for / and /fusionpbx, one will hit and it will look nice again.

Second, and maybe this is a big deal. The installer doesn't recognize the project path. Therefore, it has in many ways a hardcoded path /var/www/fusionpbx which as a system administrator, it is not always the best place. Using _SERVER['DOCUMENT_ROOT'] is not enough, This patch uses relative paths while still assuming things are in/as the installer does. This may need a little work about detecting the PROJECT_ROOT web path and filesystem path. 

Third, Added mysql/mariadb database installation. Users should always have options. PostgreSQL is still the pre-selected database.

4th, Fix a CAST() SQL, mysq/mariadb doesn't use the same cast as PostgreSQL. Not a big deal on this, it is just adding an if $db_type condition.


daniel-lucio:
Among the many changes, I am getting read of ||

|| is unsafe, (see the screenshot), CONCAT is safe and it is multi-platform compatible.

Screenshot_20201230_143708

PR #5690, opened 2020-12-28
[4.5] Fix the favicon and custom_css

daniel-lucio:
If you put a default setting for favicon with a value of /file.ext  the current condition will fail as it concatenates a double //
This patch fixes it

### DIFF
--- a/resources/footer.php
+++ b/resources/footer.php
@@ -108,7 +108,7 @@
 										is_array($tmp_path) && @sizeof($tmp_path) != 0 &&
 										(
 											($tmp_url['scheme'] != '' && $tmp_url['scheme'].'://'.$tmp_url['host'].$tmp_url['path'] == $tmp_path['dirname'].'/'.$tmp_path['filename'].'.'.$tmp_path['extension']) //is url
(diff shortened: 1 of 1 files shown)


markjcrane:
A better approach is likely to normalize the path by removing the /

daniel-lucio:
I thought about that, but I think that would break settings like: //yourserver/images/logo.png or relative paths (think on complex configurations where you would like a rewrite or a specific favicon/css per page/app)

markjcrane:
Allow / in the filename is a security risk.

daniel-lucio:
Then the issue is not the extra /, but the logic to detect a valid value. I will work on that later.

markjcrane:
The path you mentioned doesn't follow normal path that is recommended for FusionPBX the theme directory for images go in fusionpbx/themes/<theme-name>/images. There is no logic problem in the path.

daniel-lucio:
Anyone should be able to upload their logo and put it wherever they prefer, Including /var/www/fusionpbx path, which is the equivalent of placing it as /image.jpg

PR #5692, opened 2020-12-29
Fix changing user status via the operator panel

winsock:
# Context
Changing user status via the operator panel was broken due to an invalid permission and database update.

# Overview
- Switch from checking `user_account_setting_edit` to `user_setting_edit` as the permission
- Change the database update to how it's updated in [/app/calls/resources/classes/do_not_disturb.php#L51]( https://github.com/fusionpbx/fusionpbx/blob/master/app/calls/resources/classes/do_not_disturb.php#L51)
- Add a `user_setting_edit` permissions check in content.php to only show the buttons if there is permission to change the status.

# Outstanding Existing Issues
- Setting DND via the operator panel does not trigger the DND feature key sync found in the calls application.

### DIFF
--- a/app/basic_operator_panel/index.php
+++ b/app/basic_operator_panel/index.php
@@ -64,7 +64,12 @@
 			}
 
 		//update the status
-			if (permission_exists("user_account_setting_edit")) {
+			if (permission_exists("user_setting_edit")) {
+				//add the user_edit permission
+				$p = new permissions;
+				$p->add("user_edit", "temp");
+
+				//update the database user_status
 				$array['users'][0]['user_uuid'] = $_SESSION['user']['user_uuid'];
 				$array['users'][0]['domain_uuid'] = $_SESSION['user']['domain_uuid'];
 				$array['users'][0]['user_status'] = $user_status;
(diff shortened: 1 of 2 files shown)


winsock:
I'm not 100% sure on the best way for resolving the feature sync issue. My initial thought was to cleanup/improve the do_not_disturb.php (and other classes while I'm at it) to allow other applications to use the class for a central location for all DND control code. That way there is no duplication any more and if a change to DND happens the two applications do not get out of sync in terms of supported features. You can see an initial WIP version of that idea here:
https://github.com/winsock/fusionpbx/tree/calls_cleanup

The other idea I had but I initially ruled out was just copying the feature code sync code into the appropriate place, that is a smaller change but feels more fragile.

markjcrane:
The change to make it more consistent on the permissions is a good thing skipping database save method I'm not liking as much. You loose some functionality by skipping the database save method. Tracking changes and who made the change.

winsock:
@markjcrane That makes sense, I'll try to use it using the save method. I mis-remembered where I got the example from. I took it from do_not_disturb.php in the calls application. I switched it to this since I was having issues with it actually updating the database as-is.
https://github.com/fusionpbx/fusionpbx/blob/master/app/calls/resources/classes/do_not_disturb.php#L51

winsock:
@markjcrane I updated this MR by adding a temporary `user_edit` permission so that database save can be used. Is this the proper way or am I missing something?

markjcrane:
Looks good.

PR #5693, opened 2020-12-29
Remove the toggle dropdown in favor of three separate toggle buttons

winsock:
# Context
As discussed in our phone call this seems like the better option then the weird dropdown interface for the calls application.
image
(My test server has a custom theme that's why it's green)

# Overview
- Remove toggle select box and javascript
- Add individual buttons for each toggle type that sets the form action and then opens the confirmation modal
- Modify the confirmation modal to just submit the form when the user confirms the toggle

### DIFF
--- a/app/calls/calls.php
+++ b/app/calls/calls.php
@@ -212,9 +212,9 @@
 		echo "	<div class='actions'>\n";
 
 		if ($extensions) {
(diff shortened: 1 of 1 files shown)


markjcrane:
Green buttons are nice.

PR #5697, opened 2020-12-30
Calls application feature class refactor and cleanup

winsock:
# Context
This MR contains all of the base changes needed to allow other applications to modify extension features without having to duplicate code. This was started to allow the operator panel to use feature key syncing without having to duplicate code into the operator panel application.

The operator panel change to use this new code is in a seperate MR #5698 

# Overview
- Remove all unused functions and variables from the classes. These functions were only referenced from commented out code in the calls application.
- Move token validation to calls.php to allow the function classes to be used outside of calls.php
- Move the toggle message to calls.php for the same reason
- Pull all common code into an abstract base PHP class to allow it to be shared among all of the feature classes.
  - `get_existing_state` Gets the existing extension feature state needed for all of the classes
  - `update` Does the per extension updates needed when things change
    - Currently only clearing the cache and sending feature key sync messages
  - `save` Saves the changes and regenerates the XML if needed
- Modify do_not_disturb, call_forward, and follow_me classes to extend feature_base and clean up their logic using the new abstracted functions

### DIFF
--- a/app/calls/calls.php
+++ b/app/calls/calls.php
@@ -52,27 +52,39 @@
 
 //process the http post data by action
 	if ($action != '' && is_array($extensions) && @sizeof($extensions) != 0) {
+		//validate the token
+		$token = new token;
+		if (!$token->validate($_SERVER['PHP_SELF'])) {
+			message::add($text['message-invalid_token'],'negative');
+			header('Location: calls.php');
+			exit;
+		}
 		switch ($action) {
 			case 'toggle_call_forward':
 				if (permission_exists('call_forward')) {
 					$obj = new call_forward;
 					$obj->toggle($extensions);
+					//set message
+					message::add($text['message-toggle']);
 				}
 				break;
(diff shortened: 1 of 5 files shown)


greenbea:
nice job

demonspork:
Is there any progress/plans to implement these changes? I'm working on some integrations that will touch this code and would benefit from these updates.

winsock:
Unfortunately, I have left the company that was using FusionPBX a couple of years back. I have no plans on continuing but if it isn't too much work and going to be accepted I can rebase this PR to resolve the conflicts. I would probably need help validating the rebased changes since I haven't touched FusionPBX since then either.

demonspork:
I'll check it out and verify it, if I need to make any changes I would need edit access to your fork or I'll just apply this to my fork and make a new PR.  I'm also going to work on the lua side of this because of some upcoming integrations I'm working on will need some modification to both sides of it.

PR #5699, opened 2020-12-30
[4.5] Cache engine revamped

daniel-lucio:
I have never been a fan of file cache engine, but I find it handy in some scenarios.

- File cache engine now has an expiration mechanism
- Memcache engine does not depend on local Freeswitch

This will fix the voicemail Bing/Azure STT hash issue with File Cache.

The detailed rationale of this will be explained in my blog.

### DIFF
--- a/app/scripts/resources/scripts/resources/functions/cache.lua
+++ b/app/scripts/resources/scripts/resources/functions/cache.lua
@@ -129,20 +129,36 @@ end
 -- @return[2] error string `e.g. 'NOT FOUND'
 -- @note error string does not contain `-ERR` prefix
 function Cache.get(key)
-  local result, err = nil, 'UNSUPPORTTED'
-
+  local result, err = nil, 'UNSUPPORTTED';
   if (cache_method == "memcache") then
     result, err = check_error(api:execute('memcache', 'get ' .. key2key(key)))
     if result then
       result = memcache_decode(result)
     end
-  end
+  elseif (cache_method == "file") then
+    key = key2file(key);
+    files_string = File.find(key .. '.*', true);
+    files_table = split(files_string, "\n", true);
+    n = table.getn(files_table);
 
-  if (cache_method == "file") then
-    key = key2file(key)
-    -- log.noticef('location: %s', key)
-    result, err = File.read(key)
-    if not result then
+    if n > 0 then
+    -- there is a hit
+      last = 0;
+      for i,v in ipairs(files_table) do
+        path,file,extension = File.pathinfo(v);
+        last = max(last, tonumber(extension));
+      end
+      
+      if last > 0 and os.time() <= last then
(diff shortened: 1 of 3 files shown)


ac000:
Hmm, you seem to have forgotten to write commit messages!

Hint: "Update blah" isn't a commit message, I'd also say it's insufficient for a summary...

As it stands there is really no way to tell if the changes are even doing what they are intending...

See [here](https://chris.beams.io/posts/git-commit/) for a good overview of writing useful commit messages. Or just take a look at the likes of the [Linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/)

markjcrane:
The commit message admits you never liked the file cache. I created it because it was needed for high performance systems. File systems today are so good that they easily out perform memcache.  Reason primarily is because to access memcache do it through a TCP socket which takes time to setup and there are a limited number of listeners or workers. You can of course increase those.  I also got reports of memcache security problems so my solution was to simply replace the file cache. Other major benefit of the file cache is that you can easily look at it in a way that is much easier than memcache.

Its very easy to expire the file cache with a cron job by looking at the age of the file for this reason the changes seem to be unnecessary however I'll provide time for other people to review changes and provide their opinions before making a final decision.

daniel-lucio:
What you are saying is half true. It is clear you don't know Memcached enough, it is not a TCP only thing, it works on UDP as well which is way faster. The only "security-risk" I see, if you want to call lit like that, is that default installation on some distros listen to all interfaces, changing to listen only lo or fw work is far enough.

Your approach of deleting with a crontab is not ideal and it has a lot of issues. For starters, you can't specify an expiration time on a specific item. You could but there is a lot of extra admin work and it loses the capability of deciding the expiration time in realtime.

There is also http://memcachefs.sourceforge.net/ for those who want to navigate.

I am sure you are really smart to understand what I did. I have been following your coding style. Also, if you found a bug, comment about it, I will be happy to fix it.

Memcached shouldn't be removed, not everybody can afford a super fast 7200+  rpms HD or an SSD.

I also don't understand your defensive position. I am fixing it and supporting the project, not replacing it.

ac000:
Another advantage of using the file system is that if you are using sendfile(2) or splice(2) you may be able to attain zero-copying of data (between kernel and user space) by having the kernel simply put the file data directly onto the clients network socket.

markjcrane:
- With a poor description this commit looks like it is trying to hide functionality.
- File cache with wildcard support. I think this introduces security risk its safer to be exact.
- Changes look more significant because in lua ending ; are optional and these were added to multiple lines.

markjcrane:
Closed as don't want to increase security risk by doing wildcard includes. Expiring the cache is also possible with a cron job and a script so I don't see the need to build it into the lua.

PR #5704, opened 2021-01-06
Do not clobber caller ID names if no number is set

winsock:
# Context
By default custom caller ID names will get clobbered by this diaplan if no number is set. This is potentially problematic if you just want a custom caller ID name set for multiple extensions using the default caller ID number.

# Overview
- Split up the two groups in this dialplan to check the name values and it they are empty set them with the defaults.


markjcrane:
Looks good thanks for sharing the changes.

PR #5716, opened 2021-01-07
Fixed PostgreSQL Commands for Fax Queue

ZionDials:
Currently, the Data Type for the `task_next_time` and `task_lock_time` in the `v_fax_tasks` are of type `timestamp`. In PostgreSQL, the data type `timestamp` does not save the timezone, whereas `timestamptz` does. Therefore, since the existing database table is configured for `timestamp` data type, the `at time zone 'utc'` must be removed in order for it to work properly.


markjcrane:
Seems like we should switch it to timestamptz so that the time zone is represented better.

ZionDials:
I agree. I didn't know how the upgrade process would occur on the database (dropping columns, modifying tables, etc.), so I chose to update the code to reflect the database. If that makes sense?

PR #5722, opened 2021-01-14
Allow dragging the whole extension box instead of just the portrait.

winsock:
# Context
Some of our users were getting confused when attempting to use the drag and drop calling functionality since the whole box wasn't draggable. It's a bit more intuitive to have the whole box draggable and easier to explain. It also gives more context on which extension your dragging from over the old style.

# Overview
- Add all of the draggable elements to the main div for the box.
  - The ID changed to the call identifier that used to be part of the image
  - The second parameter to the ondrop event was set directly to the escaped extension instead of the ID that is now the call ID.
  - The same was done to ondragleave, ondragover, and ondragstart.
- Removed the now unused span container for the portrait
- Set the portrait to not be draggable, this makes the main div start dragging if the user starts dragging when over the portrait
- The portrait still controls the virtual drag and I changed the cursor to a pointer to match that it's a button now versus the main dragging element.

### DIFF
--- a/app/basic_operator_panel/resources/content.php
+++ b/app/basic_operator_panel/resources/content.php
@@ -354,13 +354,11 @@ if (is_array($activity)) foreach ($activity as $extension => $ext) {
 			$status_hover = $text['label-status_logged_out_or_unknown'];
 		}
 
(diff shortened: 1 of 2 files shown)


markjcrane:
Sorry, I've been so behind on reviewing some pull requests. This one is a good idea and was worth accepting.

PR #5741, opened 2021-01-25
Updated Polycom provisioning files

nemerald-voip:
Updated Polycom config with more features and enhanced security. This request also has a fix for Polycom TLS issue for Debian 9 not matching Polycom default ciphers.


markjcrane:
Please give me a call so we can talk over each item.

PR #5749, opened 2021-01-29
call center agent record

greenbea:
If call center recording is enabled it will record calls for all agents if disabled it won't record for anyone.  Added ability to enable recording on individual call center agents even call center recording is disabled.


greenbea:
Another use case for setting the recording on the agent level is when using loopback for an agent, call center recording seems to fail on loopack

PR #5759, opened 2021-02-08
Dont overload DB with conditions

daniel-lucio:
You already have the condition " p.dialplan_uuid = s.dialplan_uuid"

therefore, asking for s.dialplan_uuid with the same constant value just makes the DB work more without benefit.

### DIFF
--- a/app/dialplans/resources/classes/dialplan.php
+++ b/app/dialplans/resources/classes/dialplan.php
@@ -634,7 +634,6 @@
 							}
 							if (is_uuid($this->uuid)) {
 								$sql .= "and p.dialplan_uuid = :dialplan_uuid \n";
-								$sql .= "and s.dialplan_uuid = :dialplan_uuid \n";
 								$parameters['dialplan_uuid'] = $this->uuid;
 							}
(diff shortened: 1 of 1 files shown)


markjcrane:
Did you notice those were for two different tables?

daniel-lucio:
Yes, that's the reason.

You have in your WHERE condition: p.dialplan_uuid = s.dialplan_uuid, then, if later you have p.dialplan_uuid = CONSTANT, you don't need s.dialplan_uuid = CONSTANT.  

If a = b AND a = 999 then b = 999

Databases are not smart enough to reduce conditions, they just execute them. When you have JOINS (explicit or implicit), databases start correlating data and they do it in the order of the WHERE conditions. If you have as a first condition p.dialplan_uuid = s.dialplan_uuid, you already have that match, the p.dialplan_uuid = 99999 is enough, s.dialplan_uuid = 9999 is redundant, wastes CPU cycles, memory (a JOIN is always big, especially in the dialplan tables).

markjcrane:
Tested this with PostgreSQL and using a tool to tell me the execution time and cost there was no real noticeable difference. I was using a tool to help compare cost and execution time and appeared to be the same.

daniel-lucio:
Nevertheless, that query is not well built.

PR #5760, opened 2021-02-09
More DB compatibility 

daniel-lucio:
Let users choose what they want


markjcrane:
This change assumes that the mysql accounts for everything other database. I think it would be better to use an else if mysql.

daniel-lucio:
I understand, but this will help to add more databases in the future without big effort.

SUBSTRING_INDEX exist in SQLite as well (I didn't check on others, but 2 vs 1 is why I put the ELSE with the most common function.

markjcrane:
I think it is better to be specific even if it takes a little bit more effort.

PR #5764, opened 2021-02-10
Update {$mac}.cfg

cranben:
Updated format with     account.{$row.line_number}     which allows us to drop the 5 extra accounts. going from 3061 lines to 673 
also fixed some spacing issues


demonspork:
This will put the t41s template way out of sync with the rest of the templates that are similar. The t4x/t2x etc templates already use this method and are working great. If we do this, it should probably be to almost all of the templates at the same time because it will be the same change.

To do similar sweeping changes, I have used VS Code with ctrl+shift+H to find and replace through all files.

markjcrane:
"This will put the t41s template way out of sync with the rest of the templates that are similar. " Not sure that is a good reason because they could be synced up to work the same and not be out of sync. I will consider the idea.

markjcrane:
@demonspork I was actually more interested in the original templates getting updated than creating new ones called t4x and t2x.

PR #5776, opened 2021-02-18
Check for duplicate MAC addresses in the device import

winsock:
# Context
It is possible to import duplicate MAC addresses via the bulk CSV import if they do not have their `device_uuid` column in the CSV. MAC addresses are already considered globally unique and duplication should not be allowed. Needs #5775 to ensure the MAC address is normalized for a database lookup.

# Overview
- Create the database object only once throughout the import action section
- Add in a new block that will populate `device_uuid` if the `device_uuid` field was not provided by the CSV but the `device_mac_address` was provided.
  - This block will also verify that the `domain_uuid` matches and if it does not it will disallow the import with an error message displayed to the user.

# Drawbacks
- This adds a database lookup for every device line entry in the database. Looking at `database.php` it doesn't look like it would be easy to add a secondary unique field that would automatically be detected for switching between INSERT and UPDATE.

# Other Solutions
- Use the MAC address for `device_uuid` and switch the type to `char(12)`
  - While a MAC address isn't a UUID in the strictest sense, it is globally unique for this use case and could be used as a primary key.
  - We cannot switch `device_mac_address` to be the primary key since the database code requires a _uuid column for the insert/update detection to work.
  - This way we can get the same result as the MR but without doing potentially a ton of database lookups.

### DIFF
--- a/app/devices/device_imports.php
+++ b/app/devices/device_imports.php
@@ -252,10 +252,14 @@
 		//set the domain_uuid
 			$domain_uuid = $_SESSION['domain_uuid'];
 
+		//open the database
+			$database = new database;
+			$database->app_name = 'devices';
+			$database->app_uuid = '4efa1a1a-32e7-bf83-534b-6c8299958a8e';
+
 		//get the users
 			$sql = "select * from v_users where domain_uuid = :domain_uuid ";
 			$parameters['domain_uuid'] = $domain_uuid;
-			$database = new database;
 			$users = $database->select($sql, $parameters, 'all');
 			unset($sql, $parameters);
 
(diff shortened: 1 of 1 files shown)


winsock:
@markjcrane I personally think the secondary proposed solution is better but it mangles the meaning of the _uuid field for device and I'm not positive if that kind of change would cause issues elsewhere in FusionPBX. Let me know if you prefer it and I can open a MR with that change.

markjcrane:
I prefer to keep our primary keys with current naming convention of the singular application name a postfix of _uuid. If it is ever changed in the future it would be app_name_id, uuid, or id. I have code that is dependent on consistent naming convention for primary keys. Not very excited about changing the convention.

Also there is some cases where a the device mac address doesn't have to be unique like one used as a template could be all zeros. Soft phone's don't use mac addresses. We do have some features that will use the devices feature to help provision soft phones or at least to contain the data for them.

winsock:
Those are some very good points that would rule out my secondary proposed change. The original driver for this change came about issues faced by some of my coworkers.

The workflow that caused the issue was found when we imported a CSV file that had some mistakes in the line labels. The immediate instinct was to correct the labels and then re-import. This caused duplicate devices and a bit of confusion with my coworkers. The current correct method would have been re-export with the UUID column and then do the modifications, or to delete all the imported devices then re-import. 

My next idea to reduce confusion and still allow validation is to possibly have the proposed change in this MR but have it configurable in the device_imports.php between the Enclosure and File Upload inputs. It can default to false to retain current behavior but then allow users to enable duplicate MAC checking on demand. A new default setting could then be added to change the default to true if someone always wants it enabled.

markjcrane:
> My next idea to reduce confusion and still allow validation is to possibly have the proposed change in this MR but have it configurable in the device_imports.php between the Enclosure and File Upload inputs. It can default to false to retain current behavior but then allow users to enable duplicate MAC checking on demand. A new default setting could then be added to change the default to true if someone always wants it enabled.

This idea would be fine a default setting or an option on the page as you described to allow choosing to allow duplicates or not. I think most of the time for import preventing duplicates is a good idea so have decided to accept this pull request.

PR #5778, opened 2021-02-20
Improved Missed Call accuracy, cdr statistics, and hide duplicated CDRs from Enterprise Ring Groups

demonspork:
Purpose
----------
I have been getting clients wanting statistics on their calls and useful reports, but when we try to explain and validate the results of the CDR page and the CDR statistics page, I always come up short because the stats are just not accurate and broken in many ways. This is round 1 of improving this accuracy, making it more useful, and filtering out the remaining "chaff" that has no use except to clutter our views in the CDR and throw off missed call and volume statistics. I dug around with using external reporting tools, but 90% of the work is already there in fusionpbx.

Changes
-----------
  - Improve CDR Import Logic so that missed_call column is more accurate to the "missed" status. It would previously mark unanswered outbound calls as "missed", but they can be be more accurately filtered for by selecting Call status "Cancelled" and direction "Outbound".
  - Don't mark the CDRs of the "legs" of an Enterprise Ring Group call as missed, only the originating_leg will be marked (one missed call per call) - We could also just "skip" importing these call legs. Simultaneous ring groups don't have these duplicated CDRs for every ringing phone and are fine without them. The "Skip" approach might make most of the rest of this work irrelevant. The only benefit of having them is really for diagnostics.
  - Create `originating_leg_uuid` column in v_xml_cdr and import it into the database during CDR imports so it is available for filtering Enterprise Ring Group calls out of CDRs and reports.
  - Create `voicemail_message` boolean column in v_xml_cdr and a voicemail_message_seconds channel variable in voicemail lua app. On CDR import, voicemail_message is set to true if voicemail answered but a message was left, and false if the caller hung up before leaving a message.
  - If Voicemail answers, but the caller hangs up without leaving a message that call shows up in the "Cancelled" filter
  - Move logic that hides the agent leg of CC calls, LOSE_RACE calls, and the Enterprise Leg  from xml_cdr.php into xml_cdr_inc.php in the SQL query WHERE clause so the CDR page looks more consistent. The logic is the same, but these calls are now excluded from the query result entirely instead of having to "skip" rendering them in the list on the xml_cdr.php page.
  - Improved CDR statistics page to use the missed_call variable instead of relying upon billsec and answer_stamp/answer_epoch. Added the same logic as the xml_cdr pages to the query so it excludes enterprise ring group call legs.
  - Fixed a display bug for TTA in xml_cdr.php
  - Reimplemented the "Failed" Call status so it actually works to search the database
  - Laid the query groundwork in xml_cdr_statistics to report on Average TTA (No UI changes yet to include that statistic)

Retroactive Change Applications
-----------------------------------
There are a few changes going back in time to bring everything in line with this better reporting accuracy:
  - If you want to populate the `originating_leg_uuid column` in `v_xml_cdr` for calls prior to this update, it will rely upon having the `json` column and not having deleted the data from it like I know some people do for space saving (including me).
  - If you don't have the json column,  you are mostly out of luck for hiding the duplicate legs of Enterprise ring group calls historically. It might be possible, but it isn't going to be easy.
  - On Newer Versions of postgres, this works (I tested it on 13):
```
UPDATE v_xml_cdr SET originating_leg_uuid = (json->'variables'->>'originating_leg_uuid')::uuid WHERE json->'variables'->>'originating_leg_uuid' IS NOT NULL;
```
  - For some reason on postgres 9.4, I had to UPDATE every single record because I couldn't get it to allow the json syntax properly after the WHERE. This is fine, it doesn't change the end result it just means it has to run the UPDATE on every record, which will take a while
```
UPDATE v_xml_cdr SET originating_leg_uuid = (json->'variables'->>'originating_leg_uuid')::uuid;
```  
- To remove the `missed_call = true` on all your previous originating_leg/lose_race calls after they have had their leg column imported. If you don't have the json column you just can't trust your missed_call statistics for a while.
```
UPDATE v_xml_cdr SET missed_call = false WHERE originating_leg_uuid IS NOT NULL;
```
  - To remove the `missed_call = true` on all your previous outbound records so that they don't show up when you filter on missed (outbound unanswered calls can be accurately listed with TTA max 0 and direction outbound)
```
UPDATE v_xml_cdr SET missed_call = false WHERE direction = 'outbound' AND missed_call = true;
```

Future Improvements
-----------------------
There is still some work to do to make everything accurate and more useful. I haven't touched the extension summary yet, but it should benefit from some of this work already. I am also going to make a few minor feature improvements to the cdr statistics page such as including the average TTA like in the PDF export.
  - Bug 1: I still know about in the call detail records is with Follow Me. If an extension has follow me enabled and only has a single external destination, the CDR shows up as "outbound" even if the original caller was an external caller dialing the DID. If you add multiple Follow Me destinations, the CDR shows up as inbound as expected, even if all destinations are external numbers.
  - Bug 2: Extension Summary counts calls directly to voicemail as "missed" for the caller on internal calls. If 425 dials *99426, the extension summary seems to count that as a "missed" call for 425.
  - Bug 3: Calls that were put on park and picked up display their caller ID name in the CDR list prefixed by PARK

### DIFF
--- a/app/scripts/resources/scripts/app/voicemail/resources/functions/record_message.lua
+++ b/app/scripts/resources/scripts/app/voicemail/resources/functions/record_message.lua
@@ -587,7 +587,7 @@
 
 		--if the recording is below the minimal length then re-record the message
 			if (message_length > 2) then
-				--continue
+				session:setVariable("voicemail_message_seconds", message_length);
 			else
 				if (session:ready()) then
 					--your recording is below the minimal acceptable length, please try again
--- a/app/xml_cdr/app_config.php
+++ b/app/xml_cdr/app_config.php
@@ -564,6 +564,12 @@
 		$apps[$x]['db'][$y]['fields'][$z]['type'] = "text";
 		$apps[$x]['db'][$y]['fields'][$z]['description']['en-us'] = "Save the last application data.";
 		$z++;
+		$apps[$x]['db'][$y]['fields'][$z]['name'] = "voicemail_message";
+		$apps[$x]['db'][$y]['fields'][$z]['type']['pgsql'] = "boolean";
+		$apps[$x]['db'][$y]['fields'][$z]['type']['sqlite'] = "text";
+		$apps[$x]['db'][$y]['fields'][$z]['type']['mysql'] = "text";
+		$apps[$x]['db'][$y]['fields'][$z]['description']['en-us'] = "";
+		$z++;
 		$apps[$x]['db'][$y]['fields'][$z]['name'] = "missed_call";
 		$apps[$x]['db'][$y]['fields'][$z]['type']['pgsql'] = "boolean";
 		$apps[$x]['db'][$y]['fields'][$z]['type']['sqlite'] = "text";
--- a/app/xml_cdr/app_config.php
+++ b/app/xml_cdr/app_config.php
@@ -197,6 +197,9 @@
 		$apps[$x]['permissions'][$y]['name'] = "xml_cdr_lose_race";
 		$apps[$x]['permissions'][$y]['groups'][] = "superadmin";
 		$y++;
+		$apps[$x]['permissions'][$y]['name'] = "xml_cdr_enterprise_leg";
+		$apps[$x]['permissions'][$y]['groups'][] = "superadmin";
+		$y++;
 		$apps[$x]['permissions'][$y]['name'] = "xml_cdr_cc_agent_leg";
 		$apps[$x]['permissions'][$y]['groups'][] = "superadmin";
 		$y++;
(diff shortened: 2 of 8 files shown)


demonspork:
I was doing my dev on a test environment with Postgres 13 and this works and fast. Using this with Postgres 9.4, loading the xml_cdr.php with no arguments takes several minutes on a well indexed database and sometimes doesn't load at all (5 minute timeout). I pushed a commit that fixes it but I haven't the faintest why it works.

Also, anyone have any input on the "Calls per minute" column in CDR stats? I'm thinking about hiding it unless you enable it in a default setting. It will only really come into play for higher volume than your average bear.

greenbea:
for enterprise originates there's a variable `ent_originate_aleg_uuid` on all originated legs

greenbea:
You want to filter out every A-leg that got originated from an A-leg not only enterprise originates so `originating_leg_uuid` is fine to use it filters out every originated A-leg that got originated from an A-leg but it's not meant only for enterprise originates. so you should change the permission name to a different name not only enterprise

greenbea:
reading the freeswitch code looks like `originating_leg_uuid` gets only set on enterprise originates :)

emaktech:
This looks like some pretty great work and some much needed changes. I particularly like the addition of `originating_leg_uuid`.

I always thought it would be useful to group all call legs together in the GUI such that they could be expanded later. e.g An inbound CDR would show the A-leg record and you could click a + or similar to show all the other resulting legs.

I assume this variable allow for this kind of grouping?

demonspork:
It absolutely would allow this kind of grouping, the originating_leg_uuid always corresponds to its "Parent" call's uuid. No call that has an originating_leg_uuid was answered and bridged to the originating call because if a "leg" answered it shows up as part of the originating call's CDR record and doesn't have its own.

demonspork:
Prompt works correctly, it shows up as "Call rejected" if rejected, but the originating_leg_uuid filters apply correctly to it. 

If anyone knows of any other scenarios that should be tested with this, let me know. I want to have this well tested before it gets merged.

emaktech:
> If anyone knows of any other scenarios that should be tested with this, let me know. I want to have this well tested before it gets merged.

I've tried to put together a list of scenarios that inbound calls can follow across a few different categories.

**"Regular" Destinations:**

- Extensions
- Voicemails
- Fax Servers
- Recordings/Phrases
- Custom dialplans

**Scenarios that create one or more subsequent call legs:**

- Calls sent to Call Center agents
- Ring groups (Enterprise)
- Ring groups (Simultaneous)
- Call forward (permanent, on busy, no answer, not registered)
- Follow Me
- Setting an outbound destination as the primary action for a destination

**Scenarios where CANCEL can be sent before timeout reached or hangup sent from FreeSWITCH:**

- Call ends on ring group
- Call ends on recording/phrase
- Call ends on IVR
- Call ends in Call Center queue

This should be a good start of the most common destinations that can create a CDR. Obviously there are an infinite number of combinations depending on what path the call took before getting to the final destination...

If anybody can think of scenarios I didn't list, I can edit this comment and add them above.

PR #5786, opened 2021-02-26
GRP261x configuration using the new v2 XML format

winsock:
# Overview
This adds in a new configuration template that works for all GRP261x phones with newer firmwares that support the v2 XML format that uses human readable names. I have been using this template in production now for a few months without issue currently on a GRP2616 and a GRP2613.

### DIFF
--- a/resources/templates/provision/grandstream/grp261x/phonebook.xml
+++ b/resources/templates/provision/grandstream/grp261x/phonebook.xml
@@ -1,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<AddressBook>
+
+	<pbgroup>
+		<id>1</id>
+		<name>Users</name>
+		<ringtones>default ringtone</ringtones>
+	</pbgroup>
+	<pbgroup>
+		<id>2</id>
+		<name>Groups</name>
+		<ringtones>default ringtone</ringtones>
+	</pbgroup>
+	<pbgroup>
+		<id>3</id>
+		<name>Extensions</name>
+		<ringtones>system</ringtones>
+	</pbgroup>
+
+	{$start_id=0}
+	{foreach $contacts as $row}
+		{if $row.category == "users"}
+		<Contact>
+			<id>{$start_id++}</id>
+			{if $row.contact_name_given != ""}
+			{if $row.contact_organization != ""}
+				<FirstName>{$row.contact_name_given} {$row.contact_name_family}</FirstName>
+				<FirstName>{$row.contact_name_given} {$row.contact_name_family}</FirstName>
+				<Company>{$row.contact_organization}</Company>
+			{else}
+				<FirstName>{$row.contact_name_given}</FirstName>
+				<LastName>{$row.contact_name_family}</LastName>
+			{/if}
+			{else}
+				<FirstName>{$row.effective_caller_id_name}</FirstName>
+			{/if}
+			<JobTitle></JobTitle>
+			<Frequent>0</Frequent>
+			{foreach $row.numbers as $number}
+			{if $number.phone_number != ""}
+				{if $number.phone_label == "work"}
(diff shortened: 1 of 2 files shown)


markjcrane:
This looks great in fact if it works good would like for it to replace the GRP2613 - GRP2616 templates rather than making a new template and not replacing the old ones.

winsock:
I like the idea of replacing the other templates with this, it would remove confusion on having two different templates for the same phones and the new format is actually readable by a human.

The counter argument that I can think of is that if someone with a GRP261x phone on a really old firmware tries to provision with the new style template it will not provision. Maybe keeping around a old style `grp261x-old` or event `grandstream-old` template to allow mass upgrading?

One thing that I have never tried before is it may be possible to have both V1 and V2 format XML within the same file that way we could provide some basic settings to help migrate old phones. Maybe something like this if it works okay on new and old firmwares:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<gs_provision version="1">
	<!-- Minimal V1 template that only sets config server and firmware server paths for the phone to upgrade -->
	<config version="1">
		<!-- Set the config and firmware paths and set it to always check for new versions -->
		<P238>0</P238>
		<P8375>0</P8375>
		<P6767>1</P6767>
		{if $grandstream_config_server_path=="none"}
		<P237></P237>
		{elseif isset($grandstream_config_server_path)}
		<P237>{$grandstream_config_server_path}</P237>
		{elseif isset($domain_name)}
		<P237>{$domain_name}{$project_path}/app/provision</P237>
		{/if}
		{if isset($grandstream_firmware_path) && isset($firmware_version)}
		<P192>{$grandstream_firmware_path}/{$firmware_version}</P192>
		{elseif isset($grandstream_firmware_path)}
		<P192>{$grandstream_firmware_path}</P192>
		{else}
		<P192>{$domain_name}{$project_path}/app/provision/resources/firmware/</P192>
		{/if}
	</config>
	<config version="2">
		<!-- Full V2 config here -->
	</config>
</gs_provision>
```

Or maybe I'm overthinking it and just supporting the newer firmware versions is good.

demonspork:
The grandstream phones put their firmware version in the User-Agent. the provision php code is already finding the manufacturers/models from the user-agent, we could add some code in there to check for a minimum firmware version and if it is lower, use a barebones file named "legacy.xml" that contains that same v1 config. That approach could be re-used across multiple phone types to get them up to a new enough version, especially with stepped upgrade requirements like with the GXP phones that have to be on 1.0.8.65 or something like that before anything newer than 1.0.9 will load.

winsock:
It almost sounds like it may be a good idea to have a dedicated "firmware.php" in provision that we can centralize firmware lookups/upgrade chains if needed outside of the method that is chosen to serve different config files to different firmware version phones.

markjcrane:
The most efficient way to share the firmware with the web server rather than a PHP pass thru. I typically do this when possible from app/yealink/resources/firmware directory and app/<vendor>/resources/firmware when the phone may not support a custom path we use app/provision/ as the directory for the firmware (not the preferred path). I think its useful to spread out the firmware to their own vendor directories because for example the polycom firmware can be a large number of files and supporting files.

If people need to support older firmware they can create a new template or look in the git history to pick up the old templates.

winsock:
@markjcrane I updated this MR removing the other old templates(as individual commits incase one needs to be reverted) and renaming this one from `grp261xv2` to `grp261x`. Let me know if you prefer it the other way with each model having its own duplicated template copy and I'll do that instead.

@demonspork I believe I have added in all of your recent template changes/new default settings to this template. But if you can double check I would appreciate it!

markjcrane:
I don't want to break backwards compatibility as much as this pull request would do.

winsock:
Sounds good, I probably misinterpreted your previous message. I'll revert the removal commits.

PR #5789, opened 2021-03-02
Add variables to set codec priority for Grandstream devices

demonspork:
If you don't like this approach it wouldn't be too hard to make the template work with a comma separated list for the codec order, it would be less new settings.


demonspork:
Also added the ability to hide remote BLF status.

winsock:
Could this potentially utilize the `array` setting type, at first thought it would seem to fix the shortfalls of both a comma separated value or a bunch of individual settings.

demonspork:
I haven't looked to make sure the provision templating can handle the array setting type, but I might play with it. The array type wouldnt actual reduce the amount of individual settings. The comma separated list would allow you to reorder the codecs with a single edit instead of having to edit 8 separate settings, so I'm leaning towards just trashing this version and going that route for that reason alone.

winsock:
True, it still appears as individual settings under the same name. I think the template system should handle arrays just fine, I've used ad-hoc arrays inside of the templates before. But I have never tried to access an array from the settings before.

markjcrane:
Yesterday for other needs in provisioning phones we made the array type set in default and domain settings so they would actually create an array that could be used in the templates. Probably would be nice to talk about this pull request over a phone call sometime.

demonspork:
My core concern is about how labor intensive it is to set the codec order using the array type or individual priority settings like the yealink phones have. If I want to re-arrange the 8 priorities for Grandstream, I have to edit a setting, save, search again, edit a setting, save, search again, and keep track of which one I have put in which order without actually seeing them.

This also carries over to if I want to customize a single domain's codec priority, I have to copy over 8 settings, or create 8 separate settings for a device.

The solutions I can see are 

1. The approach I am using here where I just use a comma separated list expanded in the templates.
2. Update the "Settings" UI so that "Array" settings types list all of them in a single page, so when I edit setting that is of type "array" the form shows a list of each array element and the value of each of them so I can fill them in all at once or re-order them easily.
3. Update the "Default Settings" list to support editing the value of the default settings without clicking through to default_setting_edit.php, just being able to set those values on default_setting.php in the list view.

PR #5791, opened 2021-03-03
CDR - Don't Check Filesystem for Recording on Load

emaktech:
I have been debugging slow loading on our CDR pages for the last few days now.

One issue that we have encountered is that currently as the page loads, it checks the filesystem for each file one at a time. In our case, we move recordings to object storage after 1 week, so each time we check for a file it passes api calls which take over 1 second each to return a result. This causes this page to not load at all for us in many cases.

Regardless, this current method is unnecessarily I/O intensive and really page load is probably not the time to be checking for each file one by one.

So this PR is the simplest solution - remove the check entirely. I would contend that the administrator should remove the record_path from the database if the file was removed so this should be acceptable.

This solves this particular issue for us, but would need feedback from others if not checking for files makes sense.

### DIFF
--- a/app/xml_cdr/xml_cdr.php
+++ b/app/xml_cdr/xml_cdr.php
@@ -762,7 +762,7 @@
 					}
 				//recording
 					if (permission_exists('xml_cdr_recording') && (permission_exists('xml_cdr_recording_play') || permission_exists('xml_cdr_recording_download'))) {
-						if ($record_path != '' && file_exists($record_path.'/'.$record_name)) {
+						if ($record_path != '') {
 							$content .= "	<td class='middle button center no-link no-wrap'>";
 							if (permission_exists('xml_cdr_recording_play')) {
 								$content .= 	"<audio id='recording_audio_".escape($row['xml_cdr_uuid'])."' style='display: none;' preload='none' ontimeupdate=\"update_progress('".escape($row['xml_cdr_uuid'])."')\" onended=\"recording_reset('".escape($row['xml_cdr_uuid'])."');\" src=\"download.php?id=".escape($row['xml_cdr_uuid'])."&t=record\" type='".escape($record_type)."'></audio>";


markjcrane:
I'm interested to know what other people think about removing the file exits check.

winsock:
I think it's a reasonable change since the file is still validated in [resources/classes/xml_cdr.php#L1154](https://github.com/fusionpbx/fusionpbx/blob/master/app/xml_cdr/resources/classes/xml_cdr.php#L1154)

The file exists check would fail too if recordings were stored in base64 in the database, but it looks like xml_cdr doesn't support the base64 storage to begin with/was never completed.

tony1661:
Forgive me if this question seems dumb, but wouldn't removing the file check mean that if, for whatever reason, the file doesn't exist, you will still be shown the play button?

winsock:
@tony1661 It shouldn't unless the database is incorrect. The check to ensure the recording path isn't empty is still there. And the file exists check is done at CDR import too.

https://github.com/fusionpbx/fusionpbx/blob/master/app/xml_cdr/v_xml_cdr_import.php#L492

https://github.com/emaktech/fusionpbx/blob/a0ab52d36901e7c0c4976f43dcb69b73e6fcd1b5/app/xml_cdr/xml_cdr.php#L765

winsock:
We could also have the download function in [resources/classes/xml_cdr.php#L1154](https://github.com/fusionpbx/fusionpbx/blob/master/app/xml_cdr/resources/classes/xml_cdr.php#L1154) update the database and clear the recording path if the file is no longer found if we are concerned about the database getting out of sync with the filesystem.

emaktech:
I think @tony1661: winsock answered the same as I would have here - the path is already in the db so everything still functions as it should.

@winsock I had thought about something similar, but it occurs to me that if for whatever reason the filesystem becomes temporarily unavailable, it could cause the path to be deleted without the administrator wanting it to happen which could be problematic.

winsock:
That is very true and unwanted. If it's wanted still the exists check could be hidden(or disabled) behind a setting to allow it to be user selectable.

markjcrane:
Thanks everyone for responding based on feedback will accept this pull request. 
@winsock It maybe a good idea to have a default setting for an option.

PR #5793, opened 2021-03-06
Adding Call Center Base Score Time in Seconds

tony1661:
This pull request is to add a field in Call Centers to manually specify the Time Score when entering the Call Center. I was requested by a customer that they wanted one Call Center to be prioritized over another.

I was able to do this by editing the XML in the dialplan however I did not like that solution since it would be hard to explain to all techs in our company.

Hopefully this pull request is ok with you guys. Let me know if you have any questions.


markjcrane:
Looks great accepting your pull request.

PR #5802, opened 2021-03-11
Small error in Fax Retry.lua

netpro25:
Zero should not be wrapped in double quotes as it's causing an SQL error.

PGRES_FATAL_ERROR
2021-03-11 10:30:47.738673 [ERR] switch_pgsql.c:680 Error executing query:
ERROR:  invalid input syntax for type numeric: "'0'"
LINE 1: ...p/c8505148-1b02-4f6f-9157-b5cf64ebcc4d.tif','','','''0''',''...
                                                             ^

### DIFF
--- a/app/scripts/resources/scripts/app/fax/resources/scripts/queue/retry.lua
+++ b/app/scripts/resources/scripts/app/fax/resources/scripts/queue/retry.lua
@@ -223,8 +223,8 @@
 			fax_file                       = fax_file or dbh.NULL;
 			fax_ecm_used                   = fax_ecm_used or dbh.NULL;
 			fax_local_station_id           = fax_local_station_id or dbh.NULL;
-			fax_document_transferred_pages = fax_document_transferred_pages or "'0'";
-			fax_document_total_pages       = fax_document_total_pages or "'0'";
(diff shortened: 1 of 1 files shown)


netpro25:
So I have realized that fixing this issues changes how fax logs are displayed. It used to be that every outbound fax only a single log no matter how man retries were exhausted. With this change every attempt is logged under the fax logs for correlating fax server. I'm not sure which behavior is more desirable but I would presume that the original coder wanted all faxes to be logged and did not add these double quotes on purpose to cause an SQL error.

demonspork:
That sounds like it would have been the original intention, only blocked by this SQL error. Every SQL error is something that isn't working as intended.

netpro25:
This seems like an easy fix. To just toss in an if statement to fix it so it only logs when fax_document_total_pages and fax_document_transferred_pages is not null (or empty).

netpro25:
Here's a fix that does not create the extra log entries or yield an SQL error:

`
fax_document_transferred_pages = fax_document_transferred_pages or '-1';
fax_document_total_pages       = fax_document_total_pages or '-1';

if (fax_document_total_pages ~= -1) then
	dbh:query(sql, params);
`

Let me know which way you guys wanna go with this.

PR #5803, opened 2021-03-11
Update provision write() to match render()

winsock:
# Context
This will ensure that anyone using TFTP still can use array template variables. I am one of those unlucky ones that have to support a few devices that do not have HTTP provisioning.

# Overview
- Apply changes from 906b86d3dc4c00ad53e2b93a52f854ab5b5d9b46
- Apply changes from 0b91af9f424cae593625dd133ca61d1a77e72b02


markjcrane:
A good improvement thanks for sharing it.

PR #5807, opened 2021-03-14
Support capture_server variable in config.lua

demonspork:
Enable the ability to add a capture_server variable to config.lua. This allows us to set the capture server without having to edit the sofia.conf.lua script.


daniel-lucio:
I remember I touched this subject years ago, my proposal was rejected, it was more or less like yours.

Note: instead of using a global variable that lives in the config.lua file, better if you use the sip_profile app and add the tag only if the variable is present. This way you could have granular control of what sip profiles you want to send to a specific capture-server. Since mod_sofia is only loaded at boot-time, having it in config.lua is kind of useless.

Send me a text if you want I do the PR.

demonspork:
I was unaware that capture-server could be set on a per profile context. This approach is the bare minimum approach to supporting a capture server and doesn't change anything in default configuration or introduce any new behaviors except it allows you to protect your capture server settings from updates.

demonspork:
This seems to have been implemented with a new sofia global config app instead of this method. Closing.

PR #5818, opened 2021-03-18
Hide Global Dialplans & Add "Show Global" Button

c0reM:
This code hides global context dialplans from the dialplan manager by default.

The goal is to prevent administrators from accidentally editing global dialplans in day to day operations:

- Prevents accidentally modifying a dialplan in global context and breaking entire system
- Safer when using the select all / checkboxes
- Faster to find the dialplans unique to the current domain/tenant by reducing clutter of global dialplans when not needed

There are also a few changes to the buttons:

- Adds a "Show Global" button to show both local + global dialplans
- Adds a "Show Local" button in global view  to return to local dialplans only
- Adds both buttons when in the "Show All" view to allow users to return to another view


greenbea:
global context can also have the value of `global` not only `${domain_name}'`  as you can see https://github.com/fusionpbx/fusionpbx/blob/master/app/scripts/resources/scripts/app/xml_handler/resources/scripts/dialplan/dialplan.lua#L163
you'll have to add something like `$sql .= " and dialplan_context not in ('\${domain_name}', 'global')"`

markjcrane:
Waiting some time to decide whether to like or dislike this change. My initial reaction was a preference for adding a permission to hide or show global rather than additional buttons on the list. Will think about it more.

emaktech:
@greenbea Thanks for the tip, I've modified the code as suggested and also added a comment to follow existing code style.

@markjcrane Certainly this is a change people will notice so it's important to think about.

I thought about a permission, but it's not appropriate for how we commonly use dialplan manager.

- When working on customizing a tenant's settings: want to see the local view
- When troubleshooting a call flow: want to see the global view

A permission could definitely be useful too, but it's not that we don't want to see global dialplans ever. It's really just about preventing human error in day to day. Unsurprisingly, there are stories to this :)

markjcrane:
I've been thinking the next logical step is a select box that shows each unique context. Default the view to the domain context with an option for all, global, and domain. People can actually have more context then just the domain and global context. I also think a permission for dialplan global is a good idea so they can be hidden if desired.

daniel-lucio:
Hiding global dial plans will just make harder the debug. 

If you want to make evident the presence of a global dial plan, why not just put a different background in that row (grey?)

demonspork:
I've always been in favor of having all pages with local and global settings/dialplans/features Having a "Default" and "Modified" visual identifier. So instead of having an independent dialplan for do-not-disturb present for each domain, simply have it be in the list and indicate it is unmodified. Once edited, it has a column or coloring indicating it is different than the system default and disabling it only disables it for that specific domain unless  you go into it and select a "Disable globally" option.

This would also be useful for "Domain Settings" that instead of having default settings and domain settings separate, they should really be one list and you can filter by "customized" settings that are "overridden" by a specific domain. This would also make the Domain Settings, Device Settings, Profile Settings completely redundant, the default settings would simply need a domain_uuid, device_uuid, profile_uuid column and when a setting is "copied" to a domain, the button just reads "Modify for Domain" or something like that, even though it copies it in the background and applies a domain_uuid to it.

markjcrane:
@demonspork I like the domain settings separate so you can easily see the specific changes rather than having to look through over a thousand settings.

I do have the default settings now bold if modified or custom and indicate the default value.

PR #5843, opened 2021-03-22
added more yealink DST vars

chansizzle:


chansizzle:
migrating to daylight_savings_start_... and daylight_savings_stop_... concatenations

PR #5845, opened 2021-03-23
Dynamic Destination Edit Button

winsock:
# Overview
This allows the new destination edit button to change and update dynamically as different options are selected.

This sets the UUID as the ID in the options list. The we use the javascript onclick listener to get the id(uuid) of the selected option and open the link to the edit page. This also adds voicemail_uuid to the destinations array to enable that to work with this feature.

# Potential Improvements
- This currently always shows the button even if it will not link to anything. The button should be hidden if nothing can be linked to/edited.


winsock:
@demonspork This is the dynamic changing support addition to your recent change.

PR #5847, opened 2021-03-24
Template Alias Support

winsock:
# Context
There are a bunch of templates that are copies and are exactly the same. This MR adds support for a `.alias` file that would allow templates to instead use the files of a different template. This can allow us to clean up some of the duplicated templates while retaining the user friendly naming and backwards compatibility.

# Overview
- Add a new resolve_template function to take care of the template path resolution in a single spot
- Modify render() to use resolve_template
- Modify write() to resolve template
- Add support in write() for custom domain_name template directories
- Modify device_edit.php to use the resolve_template function

# Example .alias file
Inside of cisco/7841/.alias:
```
cisco/79x1
```

### DIFF
--- a/app/devices/device_edit.php
+++ b/app/devices/device_edit.php
@@ -826,7 +826,7 @@
 				$prov = new provision;
 				$prov->domain_uuid = $domain_uuid;
 				$template_dir = $prov->template_dir;
-				$files = glob($template_dir.'/'.$device_template.'/*');
+				$files = glob($prov->resolve_template($template_dir, $device_template).'/*');
 			//add file buttons and the file list
 				echo button::create(['type'=>'button','id'=>'button_files','label'=>$text['button-files'],'icon'=>$_SESSION['theme']['button_icon_download'],'onclick'=>'show_files()']);
 				echo 		"<select class='formfld' style='display: none; width: auto;' name='target_file' id='target_file' onchange='download(this.value)'>\n";
--- a/app/provision/resources/classes/provision.php
+++ b/app/provision/resources/classes/provision.php
@@ -30,7 +30,6 @@ include "root.php";
 	class provision {
 
 		public $domain_uuid;
-		public $domain_name;
 		public $template_dir;
 		public $mac;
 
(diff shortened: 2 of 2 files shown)


emaktech:
This is a very good idea to help with managing multiple templates! I can provide a bit of information on how we manage this in case it's helpful.

To avoid needing to do any code changes in FusionPBX, we created symlinks to our main device template files at the filesystem level. This avoids needing to add any logic into FusionPBX and is transparent.

This said, there are usually a few differences between device templates, such as firmware URLs, for example. So we also modified our templates to check to see what path the phone looked in and use that as a variable to decide which specific variables we want to spit out from the main template.

Your way looks like it could be really nice too, just mentioning what we do in case it gets some ideas going :)

winsock:
@emaktech That was my initial gut reaction too, but I wasn't thrilled about commiting symlinks into Git. It can be error prone if you don't create it with the right relative path or if you accidentally create it with an absolute one. The other issue I thought of (but I don't know if this really could happen, maybe Windows) is what would happen if the filesystem didn't support symlinks.

daniel-lucio:
not really conviced about using .alias,

maybe a new default setting _SESSION['provision']['template_alias_$brand_$model] = $originalbrand/$originalmodel
_SESSION['provision']['template_alias_$brand] = $originalbrand

this way you can even expand to brands and you dont touch the FS at all

winsock:
@daniel-lucio I really like that idea! My only concern is that (at least to me) the default settings page is getting a bit unwieldy and if in subsequent merge requests we start adding in alias to de-duplicate the templates that are(or should) be the same we could end up with somewhere between 10-50 new default settings(maybe higher) I think.

My immediate first MR after getting alias support of some type was to create a `grp261xp` template and alias all of the existing grp2612-2616 templates to it. I'm sure the same could be done to some of the GXP templates too. So that alone may be 10 new settings.

demonspork:
Maybe another similar approach, instead of having templates with aliases, have a table of phone models, then a template relationship that is used for phones of that model. 

So the "Template" selector would be purely a list of phone models, then the on-disk templates would just have multiple models associated with them. 

This would also make it pretty easy to supported "Versioned" templates by having the table be v_device_models with device_model_uuid, device_model_vendor, device_model_template, and device_model_version. We can either have the version be the actual firmware version required for supported phones or just a version tracking for the model templates themselves, like V1.7 of the Grandstream GXP template for Fusionpbx.

This could also just be an update/repurpose of the device vendors table where vendors would have multiple entries, one for each phone model.

winsock:
@demonspork I think that is the ideal solution where the management of this is under the `Device Vendors` button completely backed by a database table that is populated by individual phone vendor applications by default. I think if we do that solution it makes since to do the movement of the templates to vendor applications at the same time plus any other planned improvements or changes since it would be a bigger re-work of the system than the other two proposed solutions.

Another question would be then do we want this or @daniel-lucio's solution as a stop-gap to start cleaning up templates for the ideal solution? That way when we move everything into that new system we already have just the templates needed beforehand.

markjcrane:
Seems like there should be a better way than all of the suggestions. The file alias idea would work... but still feels likely there is a better way.

winsock:
I am in agreement Mark. I personally think the ultimate solution(but the one that would take the most development time) would be a management UI/database with all templates in their own vendors applications. By doing that it would open up the possibility of storing the templates eventually in the database too in the future if that is desired (my idea on that would be have it operate a bit like the dialplan XML).

I've been slammed at my dayjob a bit with a large project but it should be done this week. If you have an idea you are comfortable moving forward with I am more than willing to spend the time developing it. I just want to make sure the path I take is one that would be wanted if its a larger change.

PR #5855, opened 2021-04-01
Fix CLI editors with faulty syntax highlining.

daniel-lucio:
When editing in the CLI with any editor that does syntax highlight, the * breaks it. Adding on the following line /* **/ restores the right highlighting


markjcrane:
The app/ecit tool in FusionPBX doesn't have any problem with the syntax. However you mentioned CLI editor. Anyone else have comments or opinions on this pull request?

markjcrane:
Seems more like a bandied for a program with broken syntax highlighting.

markjcrane:
I'll accept the pull request but will add a description to the code to say why it has been added to the code.

PR #5856, opened 2021-04-02
[4.4] important variable that needs to be present always

daniel-lucio:


markjcrane:
More accurately this variable is important to people that are using accountcode.

PR #5857, opened 2021-04-02
[4.5] Important variable that needs to be present

daniel-lucio:


markjcrane:
Original pull request commit message needs to be more descriptive.
- [4.5] Important variable that needs to be present.

greenbea:
In my opinion, if the whole domain uses the same accountcode it should be placed in the domain-variables dialplan as loopback (that's is used in ring group and follow me) won't have it set otherwise

daniel-lucio:
@greenbea the thing is that 1 domain = 1 customer is only one possible scenario. Think on the scenarios where 1 extension = 1 customer, or N extensions = 1 customer within the same domain. (residential service, wholesale, internal accounting in medium companies, etc.)

Accountcode, as a variable itself it has no technical meaning, but managment. The loopback variables have a technical meaning, this is why the place where they are set is different.

demonspork:
The idea would be that if the extensions have an account code set, it supersedes the domain default but otherwise it uses the domain default variable. The default state of an account code for an extension would be empty with this setup

daniel-lucio:
@demonspork you can do that yourself by adding the acocuntcode variable in the "variables" dialplan.

markjcrane:
I agree with other comments here from @greenbea and @demonsport. I had accepted the pull request but I just changed it to be disabled by default. This allows the administrator the power to choose what they want according to their needs.

Next time use a better description that actually describes the changes.

PR #5859, opened 2021-04-05

emaktech:
The announce sound previously required users to manually type in the full path of the announce sound. It has now been modified to be a hybrid dropdown which shows available recordings on the domain and also allows for the old text entry method (type in the path).


markjcrane:
We really shouldn't use if_group("superadmin") this approach to permissions aren't nearly as flexible as using permissions. So we have been slowly replacing the if_group references in most circumstances.

markjcrane:
Please use a permission instead of the group and then I will accept your changes.

emaktech:
Closing PR and will submit a new one that works with updated version of this file.

PR #5864, opened 2021-04-06
Dialplan Fix for Feature Key Sync Race

demonspork:
This is a dialplan approach to fix the feature key sync loop/race with yealink phones. This approach may also need to be extended to the Call Forward dialplans. It limits the number of simultaneous calls for a particular user to change their DND/Forwarding status to a single call at a time and returns user busy. The yealink phones will re-try their background feature key sync phone call several times and the call is very short it doesn't introduce any risk of the phones being "out of sync" with the server's DND status.

I am also reverting the lua approach because it was not reliable and can still easily allow the loop to occur.


markjcrane:
This looks like a better approach to the problem. Accepted your changes.

PR #5865, opened 2021-04-07
Fixed Queues so they work and fixed the commented examples

tony1661:
In troubleshooting a known FreeSwitch bug with mod_callcenter, I started to try out mod_fifo as an alternative. I realized that the dialplan that was being created by Fusion was not working. Creating a queue is very simple so I knew it wasn't a user error on my part so I dug in deeper. There were a few issues in the dialplan generation that I needed to fix to get queues working again.

1. Add a "break on-true" to each condition destination_number (it was changed in two places)
2. populate the "action" for callers entering the queue via the $agent_queue_extension_number
3. remove the appended "_agent" which was added to the $queue_name which was causing the queue for entering and retrieving calls different.
4. remove the inline="true" for the action that executes the lua script

I also fixed the example comments to be accurate to how the queues should be written in the dialplan.

### DIFF
--- a/app/fifo/fifo_add.php
+++ b/app/fifo/fifo_add.php
@@ -167,9 +167,7 @@
 				$dialplan_detail_order = $dialplan_detail_order + 10;
 
 				//action fifo
-				//if (strlen($pin_number) > 0) { $pin_number = "+".$pin_number; }
-				//if (strlen($flags) > 0) { $flags = "+{".$flags."}"; }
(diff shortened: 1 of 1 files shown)


tony1661:
It should also be noted that there seems to be a bug that was existing before my proposed changes that when the Queue is first created and saved, FusionPBX doesnt actually add it to the dialplan. FOr some reason you need to open the queue back up and then click save again and the dialplan will be live.

As I mentioned this was found to be the case even before my PR.

PR #5867, opened 2021-04-08
[4.5] AS support to the [cdr][field] default setting

daniel-lucio:
if you put a default setting like

_SESSION[cdr][field][array] = xml_cdr_uuid as CDR

it allows you to present a column with an alternative name. 

The best use case is letting users see their CALL UUID (when reporting a call) without letting them view the call detail. Very handy to do support.

### DIFF
--- a/app/xml_cdr/resources/classes/xml_cdr.php
+++ b/app/xml_cdr/resources/classes/xml_cdr.php
@@ -108,6 +108,7 @@ if (!class_exists('xml_cdr')) {
 		 * cdr fields in the database schema
 		 */
 		public function fields() {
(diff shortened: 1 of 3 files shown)


markjcrane:
If you want it committed provide an example use case that shows why it has value.

daniel-lucio:
> The best use case is letting users see their CALL UUID (when reporting a call) without letting them view the call detail. Very handy to do support.

Screenshot_20210407_230049

When someone wants to report a call, it is more accurate to specify the UUID rather than date/caller/callee. Remember the date (in the CDR) is displayed in the local timezone which is very easy to mess up.

I also will ask you to start asking everybody for a use case.

PR #5922, opened 2021-04-28
Update valet parking dialplan to allow it to work with the auto parking dialplan

winsock:
# Context
This is a continuation of the work reverted in #5611 and effectively reverts #5918

**This PR is still considered a draft and untested currently**

I plan on testing this out this week to ensure this kind of hook/forwarding works as I expect it to but I wanted to get it out there for comments and concerns with this kind of approach.

# Overview
* Parking lots are no longer directly prefixed with *. But for backwards compatibility can be called that way still.
* A lua event hook is added to forward `park+59xx` events to `park+*59xx` for BLFs still listening to that 
* Parking lot name changed from park to 5900 to match #5612 for whole lot BLF monitoring
  * https://freeswitch.org/confluence/display/FREESWITCH/mod_valet_parking#mod_valet_parking-UsingPresence
* Fix regex from matching 5900 to allow this dialplan to work with the valet_park_auto
* Enable valet_park_auto by default now that it can be safely and sanely used in combination with manual parking.
* Remove 475_valet_park_out.xml as it is not needed since the normal valet_park 470 order diaplan can now retrieve calls in the parking lots.

### DIFF
--- a/app/dialplans/resources/switch/conf/dialplan/470_valet_park.xml
+++ b/app/dialplans/resources/switch/conf/dialplan/470_valet_park.xml
@@ -1,18 +1,17 @@
 <context name="{v_context}">
-	<extension name="valet_park" number="park+*5901-*5999" continue="false" app_uuid="3cc8363d-5ce3-48aa-8ac1-143cf297c4f7" enabled="true" order="470">
-		<condition field="destination_number" expression="^(park\+)?\*(59[0-9][0-9])$" break="never"/>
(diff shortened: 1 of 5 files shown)


winsock:
@markjcrane This is what I was thinking for allowing the parking change as discussed yesterday. The only thing I'm not sure on how to do is force the lua.conf.xml change on upgrade so the hook will run for everyone. Maybe there is another place where I can start the hook/lua script that I am not aware of?

PR #5923, opened 2021-04-28
Yealink add call forwarding dial code variables

greenbea:


demonspork:
This has been on my to-do list for a while. If these dial code variables are added to all the templates, the feature key syncing works for call forwarding. It only currently works for DND.

greenbea:
it's not feature-sync it just dials the actual code

demonspork:
It is also needed for the feature sync to work properly for forwarding. 2 birds with one stone.

PR #5924, opened 2021-04-29
[voicemail] allow setting greeting_number_id to default

greenbea:
This adds the ability to override any set greeting_id to use the default greeting


greenbea:
`<action application="set" data="voicemail_greeting_number=default"/>`

PR #5931, opened 2021-05-06
Allow the use of TAB as a delimiter

frytimo:
Allows the use of TAB character for a delimiter while importing.


greenbea:
How will you differentiate between values that have a space in them and the delimiter?

frytimo:
The delimiter is an actual TAB character ASCII code 9 and the space will be ASCII code 32.

greenbea:
Oh I didn't realize you said tab I read space 👍

PR #5937, opened 2021-05-11
[voicemails] Option to have voicemail emailed upon request only

greenbea:
Voicemail to email automatically sends an email if the `voicemail_mail_to` field has an email in it. There are some cases where people only want to get their voicemails emailed only when they request it from option 9 of the `listen_to_recording`

This commit adds an option to set voicemail_auto_mail to false and won't automatically get voicemails to email


demonspork:
I had actually already started on this myself. The next step is to get email queue to support this by doing transcribe but not sending.

PR #5940, opened 2021-05-12
Grandstream Fix Logic for outbound proxy and secondary servers

demonspork:
There was an illogical if/else statement that the "else" condition would always set a null value, so it didn't really matter. Re-wrote it to properly filter if the transport is set to dns srv and not affix the SIP Port to the server address.

### DIFF
--- a/resources/templates/provision/grandstream/dp750/{$mac}.xml
+++ b/resources/templates/provision/grandstream/dp750/{$mac}.xml
@@ -27,12 +27,16 @@
 
 <!-- Primary SIP Server -->
 <!-- String: serveraddress  -->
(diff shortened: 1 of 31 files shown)


markjcrane:
Currently this commit removes grandstream_ipv_mode possible other settings affected to.

chansizzle:
@demonspork can you look through your commit and adjust to this commit? https://github.com/fusionpbx/fusionpbx/pull/5934

P1415 has been updated from hard coded to be variablized.

demonspork:
That should do it. I also found a few examples of the variable that weren't originally updated to use the grandstream_ipv_mode variable.

chansizzle (approved):
looks good to me. thanks!

PR #5942, opened 2021-05-15
Update call_center_agent_edit.php

borisk1976:
Either this line is unnecessary  or code is wrong


markjcrane:
Looks like you found an extra bit of code used in the past but no longer used.

PR #5944, opened 2021-05-18
Add Permission to set Blank emergency Number

demonspork:


markjcrane:
I was slow to accept this because I wasn't sure on using the word blank. Found a link that compares blank and empty.
- https://ux.stackexchange.com/questions/100151/word-suggestion-empty-vs-blank#100153
  - blank (adjective) - not written or printed on.
  - empty (adjective) - containing nothing; not filled or occupied.

I think empty is a little more accurate from this description and so went with it.

PR #5951, opened 2021-05-20
Fix for ajax redirects when logged out

mhoogveld:
Responses to AJAX calls are now checked for a redirection to the login page and redirect the user to the login page if logged out.
This fixes an issue where fail2ban detects too many login tries when some users leave an "Operator Panel" tab open after logging out in an other tab. The "Operator Panel" tab keeps refreshing and requesting the login page.

### DIFF
--- a/app/basic_operator_panel/index.php
+++ b/app/basic_operator_panel/index.php
@@ -244,11 +244,20 @@ unset($refresh_default);
 	}
 
 	loadXmlHttp.prototype.stateChanged=function () {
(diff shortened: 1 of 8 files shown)


markjcrane:
This caused a bug on destinations PHP class removed the change to fix the problem. 
It caused an error 'Uncaught TypeError: this.xmlHttp is undefined'

This feature doesn't need to check whether its logged out as it is only using to the request to update when the user does something and not refreshing constantly.

- https://github.com/fusionpbx/fusionpbx/commit/4b1ab310b797c669245195b55e7f7d311280d96c

PR #5953, opened 2021-05-21
Yealink Support Searching Remote Phonebook on Dial pad

demonspork:
By default, the dial pad only searches the Local phonebook and recent call history for suggestions while dialing. This adds support for searching the Remote Phonebook if configured and enabled. There is also commented out the definition to search an LDAP phonebook if one has been configured elsewhere.


konradSC:
Nice feature! I didn't know this existed. 

Perhaps a better way to implement this feature would be to set $yealink_super_search = true/false in Default Settings. Then put something like this in the cfg file. 

> {if $yealink_super_search == 'true'}
super_search.url = https://{if isset($http_auth_username)}{$http_auth_username}:{$http_auth_password}@{/if}{$domain_name}/app/provision/?file=super_search.xml
{else}
super_search.url = 
{/if}

demonspork:
I think the reason I put it in this way was so that orgs could easily customize the ldap search settings in super_search.xml without having to modify the file in the actual template, just create a custom one and set the url to point to it.

PR #5955, opened 2021-05-22
Use epoch times for CDR Imports

demonspork:
There has been some discussion of edge cases for CDR Importing time zones/time stamps.

This modification makes the start_stamp, answer_stamp, and end_stamp values use the corresponding _epoch times for import into the v_xml_cdr table.

This prevents scenarios where the DB's default time zone is set differently than the switch's timezone. 

I believe this is the most flexible solution because it doesn't require a new setting, it doesn't require reading the system timezone (which isn't always the same system for postgres and freeswitch) and it prevents all possible timezone offset mistakes when setting the timestamptz columns, regardless if the admin is using a default fusionpbx install or any altered/custom installation this will work for all of those scenarios because it uses an absolute time instead of a formatted string that doesn't define its timezone (start_stamp from freeswitch doesn't tell you what time zone it used).

## Possible Improvements

- This leaves the timestamptz timezone in UTC. We could look up the time zone setting for the domain of the CDR and set the timezone accordingly so that the timestamp column will show the most commonly used view of the date. It won't impact how postgres stores the data, just what timezone postgres displays it in by default. As far as I can tell, all references to the date/time of a CDR elsewhere in fusionpbx are using the various _epoch fields or they are executing a query with a specific timezone using AT TIME ZONE in the SQL, so this won't have any impact on the display except when looking at the records in the DB manually.
- Update the other import script, 'v_xml_cdr_import.php' with these same improvements. 

## Edge Cases/Conflicts
I cannot find a situation or think of one where this will have any impact on anything else. Please speak up if you know of any objections to this change. The reason I made this PR is because I had 3 or 4 scenarios I can readily think of and one I personally encountered when using freeswitch's _stamp fields and I can't think of any problems with using the epoch.

### DIFF
--- a/app/xml_cdr/resources/classes/xml_cdr.php
+++ b/app/xml_cdr/resources/classes/xml_cdr.php
@@ -346,13 +346,15 @@ if (!class_exists('xml_cdr')) {
 						$this->array[$key]['pin_number'] = urldecode($xml->variables->pin_number);
 
 					//time
-						$this->array[$key]['start_epoch'] = urldecode($xml->variables->start_epoch);
-						$start_stamp = urldecode($xml->variables->start_stamp);
-						$this->array[$key]['start_stamp'] = $start_stamp;
-						$this->array[$key]['answer_stamp'] = urldecode($xml->variables->answer_stamp);
-						$this->array[$key]['answer_epoch'] = urldecode($xml->variables->answer_epoch);
-						$this->array[$key]['end_epoch'] = urldecode($xml->variables->end_epoch);
-						$this->array[$key]['end_stamp'] = urldecode($xml->variables->end_stamp);
+						$start_epoch = urldecode($xml->variables->start_epoch);
+						$this->array[$key]['start_epoch'] = $start_epoch;
+						$this->array[$key]['start_stamp'] = date('c', $start_epoch);
+						$answer_epoch = urldecode($xml->variables->answer_epoch);
+						$this->array[$key]['answer_epoch'] = $answer_epoch;
+						$this->array[$key]['answer_stamp'] = date('c', $answer_epoch);
+						$end_epoch = urldecode($xml->variables->end_epoch);
+						$this->array[$key]['end_epoch'] = $end_epoch;
+						$this->array[$key]['end_stamp'] = date('c', $end_epoch);
 						$this->array[$key]['duration'] = urldecode($xml->variables->duration);
 						$this->array[$key]['mduration'] = urldecode($xml->variables->mduration);
 						$this->array[$key]['billsec'] = urldecode($xml->variables->billsec);
@@ -424,6 +426,7 @@ if (!class_exists('xml_cdr')) {
 						$this->array[$key]['pdd_ms'] = urldecode($xml->variables->progress_mediamsec) + urldecode($xml->variables->progressmsec);
 
 					//get break down the date to year, month and day
+						$start_stamp = urldecode($xml->variables->start_stamp);
 						$start_time = strtotime($start_stamp);
 						$start_year = date("Y", $start_time);
 						$start_month = date("M", $start_time);
--- a/app/xml_cdr/resources/classes/xml_cdr.php
+++ b/app/xml_cdr/resources/classes/xml_cdr.php
@@ -638,7 +638,7 @@ if (!class_exists('xml_cdr')) {
 									$array['call_recordings'][$x]['call_recording_name'] = $record_name;
 									$array['call_recordings'][$x]['call_recording_path'] = $record_path;
 									$array['call_recordings'][$x]['call_recording_length'] = $record_length;
-									$array['call_recordings'][$x]['call_recording_date'] = urldecode($xml->variables->start_stamp);
+									$array['call_recordings'][$x]['call_recording_date'] = date('c', $start_epoch);
 									$array['call_recordings'][$x]['call_direction'] = urldecode($xml->variables->call_direction);
 									//$array['call_recordings'][$x]['call_recording_description']= $row['zzz'];
 									//$array['call_recordings'][$x]['call_recording_base64']= $row['zzz'];


app/xml_cdr/resources/classes/xml_cdr.php
```diff
@@ -346,13 +346,15 @@ function ($matches) {
...
-						$this->array[$key]['start_epoch'] = urldecode($xml->variables->start_epoch);
-						$start_stamp = urldecode($xml->variables->start_stamp);
-						$this->array[$key]['start_stamp'] = $start_stamp;
-						$this->array[$key]['answer_stamp'] = urldecode($xml->variables->answer_stamp);
-						$this->array[$key]['answer_epoch'] = urldecode($xml->variables->answer_epoch);
-						$this->array[$key]['end_epoch'] = urldecode($xml->variables->end_epoch);
-						$this->array[$key]['end_stamp'] = urldecode($xml->variables->end_stamp);
+						$start_epoch = urldecode($xml->variables->start_epoch);
+						$this->array[$key]['start_epoch'] = $start_epoch;
+						$this->array[$key]['start_stamp'] = date('c', $start_epoch);
```
winsock, line 351: Might want this to be `date(DateTime::ATOM, $start_epoch)`

In PHP c or DateTime::ISO8601 is not really ISO8601 but ATOM is. They couldn't change it due to backwards compatibility.

app/xml_cdr/resources/classes/xml_cdr.php
```diff
@@ -346,13 +346,15 @@ function ($matches) {
...
+						$answer_epoch = urldecode($xml->variables->answer_epoch);
+						$this->array[$key]['answer_epoch'] = $answer_epoch;
+						$this->array[$key]['answer_stamp'] = date('c', $answer_epoch);
```
winsock, line 354: Same as above, `date(DateTime::ATOM, $answer_epoch)`

app/xml_cdr/resources/classes/xml_cdr.php
```diff
@@ -346,13 +346,15 @@ function ($matches) {
...
+						$end_epoch = urldecode($xml->variables->end_epoch);
+						$this->array[$key]['end_epoch'] = $end_epoch;
+						$this->array[$key]['end_stamp'] = date('c', $end_epoch);
```
winsock, line 357: Same as above, `date(DateTime::ATOM, $answer_epoch)`

app/xml_cdr/resources/classes/xml_cdr.php
```diff
@@ -424,6 +426,7 @@ function ($matches) {
 						$this->array[$key]['pdd_ms'] = urldecode($xml->variables->progress_mediamsec) + urldecode($xml->variables->progressmsec);
 
 					//get break down the date to year, month and day
+						$start_stamp = urldecode($xml->variables->start_stamp);
 						$start_time = strtotime($start_stamp);
 						$start_year = date("Y", $start_time);
 						$start_month = date("M", $start_time);
```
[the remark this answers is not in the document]
winsock: GitHub didn't allow me to highlight down to 433 in my comment but include that line.

winsock (commented):
Looks pretty good! There are a few minor things I would change but otherwise what I had in mind too.

Other changes that may be needed:
- `$array['call_recordings'][$x]['call_recording_date'] = urldecode($xml->variables->start_stamp);` needs to be replaced with the new start_stamp derived from the epoch.
- The same changes need to be applied to `v_xml_cdr_import.php` too I think.

app/xml_cdr/resources/classes/xml_cdr.php
winsock: It seems like PHP did change c to be ATOM in recent PHP versions so this may be okay.

winsock: Yep it seems fixed since PHP5.4 so ignore this :)

demonspork:
I agree with the ATOM change, I'll add a commit for that. 

The start_year etc has to use the generated timestamp from Freeswitch because it is used for file paths for recordings starting at line 573 and I think freeswitch is calculating days/hours etc based on its offset, not UTC.

app/xml_cdr/resources/classes/xml_cdr.php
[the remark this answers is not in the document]
demonspork: The start_year etc has to use the generated timestamp from Freeswitch because it is used for file paths for recordings starting at line 573 and I think freeswitch is calculating days/hours etc based on its offset, not UTC.

app/xml_cdr/resources/classes/xml_cdr.php
demonspork: I agree with the ATOM change, I'll add a commit for these.

app/xml_cdr/resources/classes/xml_cdr.php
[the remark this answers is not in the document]
winsock: Ah that makes sense!

demonspork:
>     * The same changes need to be applied to `v_xml_cdr_import.php` too I think.

I'm not entirely sure why this file exists. I've added my updates to it for other changes, but it just seems to be duplicated code with minor differences in how it is structured because it isn't a class, but does all of the same things.

daniel-lucio:
> > ```
> > * The same changes need to be applied to `v_xml_cdr_import.php` too I think.
> > ```
> 
> I'm not entirely sure why this file exists. I've added my updates to it for other changes, but it just seems to be duplicated code with minor differences in how it is structured because it isn't a class, but does all of the same things.

Historical reasons, v_xml_cdr_import.php was the first import from Fusion 2.x or earlier.

winsock:
> > ```
> > * The same changes need to be applied to `v_xml_cdr_import.php` too I think.
> > ```
> 
> I'm not entirely sure why this file exists. I've added my updates to it for other changes, but it just seems to be duplicated code with minor differences in how it is structured because it isn't a class, but does all of the same things.

It looks like it's still the script in use today:
https://github.com/fusionpbx/fusionpbx/blob/master/resources/templates/conf/autoload_configs/xml_cdr.conf.xml#L4

demonspork:
Ah. We could probably review for differences (there were a few small ones I remember noticing the other day) and maybe just consolidate it all to use the class. Keep the file in place so people don't have to re-do their freeswitch configs with an update.

winsock:
@markjcrane Can the entire contents of `v_xml_cdr_import.php` be replaced then with:
```php
//check the permission
if (defined('STDIN')) {
	$document_root = str_replace("\\", "/", $_SERVER["PHP_SELF"]);
	preg_match("/^(.*)\/app\/.*$/", $document_root, $matches);
	$document_root = $matches[1];
	set_include_path($document_root);
	$_SERVER["DOCUMENT_ROOT"] = $document_root;
	require_once "resources/require.php";
} else {
	include "root.php";
	require_once "resources/require.php";
	require_once "resources/pdo.php";
}

//increase limits
set_time_limit(3600);
ini_set('memory_limit', '256M');
ini_set("precision", 6);

//import from the FreeSWITCH POST
$cdr = new xml_cdr;
$cdr->post();
```

The post method of xml_cdr doesn't seem to be referenced anywhere else right now and seems setup to handle what this file is doing?

markjcrane:
@winsock It might will need to test and see if it works. I would like to remove the old version haven't done it because it makes the upgrade a little harder and more work.

I would like to deprecate v_xml_cdr_import.php entirely and remove it rather but this will break some people's systems. At the moment would break all new install as well.

winsock:
> @winsock It might will need to test and see if it works. I would like to remove the old version haven't done it because it makes the upgrade a little harder and more work.
> 
> I would like to deprecate v_xml_cdr_import.php entirely and remove it rather but this will break some people's systems. At the moment would break all new install as well.

I can test this out on our office PBX, we always dogfood new features on it before rolling out to customers anyways. It's not a big deal if we run into a few hiccups with storing CDR records since we don't do anything with them anyways.

winsock:
I'm now running the following on our office PBX system for v_xml_cdr_import.php and so far see nothing wrong. I'll see if I notice anything strange over time.
> ```
> //check the permission
> if (defined('STDIN')) {
> 	$document_root = str_replace("\\", "/", $_SERVER["PHP_SELF"]);
> 	preg_match("/^(.*)\/app\/.*$/", $document_root, $matches);
> 	$document_root = $matches[1];
> 	set_include_path($document_root);
> 	$_SERVER["DOCUMENT_ROOT"] = $document_root;
> 	require_once "resources/require.php";
> } else {
> 	include "root.php";
> 	require_once "resources/require.php";
> 	require_once "resources/pdo.php";
> }
> 
> //increase limits
> set_time_limit(3600);
> ini_set('memory_limit', '256M');
> ini_set("precision", 6);
> 
> //import from the FreeSWITCH POST
> $cdr = new xml_cdr;
> $cdr->post();
> ```

PR #5956, opened 2021-05-25
Replace duplicated v_xml_cdr_import.php with a call to the post method of xml_cdr

winsock:
# Context
While we were discussing and proposing changes for fixing some potential timezone issues with importing CDRs from FreeSWITCH on #5955 we noticed that code appears to be duplicated between the xml_cdr class and v_xml_cdr_import.php for importing CDRs from FreeSWITCH

# Overview
- Replace the duplicated code in v_xml_cdr_import.php with xml_cdr->post();


winsock:
@markjcrane So far no regressions in the imported CDR records that I can notice running this on our companies office PBX system. CDRs seem to be importing properly. Is there anything in particular I should look for or validate to ensure this is 100% working?

demonspork:
https://github.com/fusionpbx/fusionpbx/blob/master/app/xml_cdr/xml_cdr_import.php already has everything to handle the file import correctly, including some extra stuff with memory limits. We might just add support to POST to that file and replace v_xml_cdr_import.php with require_once xml_cdr_import.php and add the post support to that file.

winsock:
> https://github.com/fusionpbx/fusionpbx/blob/master/app/xml_cdr/xml_cdr_import.php already has everything to handle the file import correctly, including some extra stuff with memory limits. We might just add support to POST to that file and replace v_xml_cdr_import.php with require_once xml_cdr_import.php and add the post support to that file.

That is true, this change does make this file basically that file but with an extra call to `post()`
I see no reason not to do that, let me do that now.

PR #5965, opened 2021-06-02
[4.5] Make fool probe the CDR configuration (default setting)

daniel-lucio:
the SQL query already has some fixed fields, if someone puts in the default settings those variables, you may end having a query like "select domain_name, .... domain_name from v_xml_cdr" which will break Pgsql and Mysql. This patch reviews before adding.


daniel-lucio:
This also happens when inserting, fields mentioned twice

PR #5979, opened 2021-06-23
Update XML CDR Extension Summary Queries

emaktech:
The queries in the Extension Summary have not been updated for a long time. Many changes have been made to the CDR import script that has changed how calls are counted. As a result, the data population in the extension summary has been wildly inaccurate.

The queries have now been copied from the $sql query output of the xml_cdr page. As such, they are updated to match what a user would see if they were to use basic filters on xml_cdr.

Additionally, the following logic now applies:

* Inbound Calls = Answered + Missed
* CC Missed  = Call Center Agent Summary Missed Calls

We have tested these queries quite extensively for a few months and compared to many different call flow configurations of live data and everything now continues to match quite well.


emaktech:
Any comments on this PR?

We now have a number of customers relying on these improvements and would really like to get this merged if possible.

emaktech:
Closing this PR because and will re-submit because it needs to be rebased.

PR #5983, opened 2021-06-27
Comments out unneeded socket connection

gruz:
$fp variable is never declared here, so it just throws an annoying warning on debug


markjcrane:
Instead of commenting it out I made sure the $fp handle was created.
- https://github.com/fusionpbx/fusionpbx/commit/f9f730024cc675153193504b9919fd724e7a7249

Credit to you for finding the problem.

PR #5990, opened 2021-07-07
Yealink add call forward code variables

greenbea:


markjcrane:
Following the names used in Yealink it seems like these would be a better variable name.

forward.always.on_code = {$yealink_forward_always_on_code}
forward.always.off_code = {$yealink_forward_always_off_code}

greenbea:
done

markjcrane:
Thanks!

PR #5993, opened 2021-07-23
Fix dashboard widget for agent status management

winsock:
# Context
The dashboard widget for call center agent self management was not working. It seems like it has been broken for some time and certainly stopped working when we switched the call center naming to <ext>@<domain> from <uuid>.

# Overview
- Fix setting agent_status in the `call_center_agents` table
- Actually set the agent status in FreeSWITCH
- Fix the tier creation and deletion to match the new `<ext>@<domain>` call center naming format
- Fix an issue where $agent array access wouldn't work due to having a container array.
- Add support for On-Break

# Possible Improvements
- This widget assumes each user is only associated to one agent. This could be restructured a bit to support multiple agents associated with a single user account if desired.

### DIFF
--- a/app/call_centers/call_center_agent_dashboard.php
+++ b/app/call_centers/call_center_agent_dashboard.php
@@ -52,42 +52,44 @@
 
 //get the http post values and set them as php variables
 	if (count($_POST) > 0) {
+	//add the user_edit permission
+		$p = new permissions;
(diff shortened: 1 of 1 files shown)


markjcrane:
Nice job! I'm accepting your pull request.

PR #6001, opened 2021-08-25
Support for Disabling Non-configured ports on GXW42XX gateways

demonspork:
We don't want dial tone on unconfigured ports. This makes it so that the device will disable ports that have no user id configured so they don't provide dial tone. The behavior can be reversed by setting grandstream_enable_fxs to 1 in the settings.


markjcrane:
It would be more consistent with the rest of FusionPBX if enable was at the end of the variable name. Change grandstream_enable_fxs to grandstream_fxs_enable. It would be more consistent with the project and then I'll accept the changes.

PR #6020, opened 2021-08-30
Simplified account array and cosmetic changes

cranben:


cranben (commented):
added the space back in

PR #6055, opened 2021-09-21
Sendmail and GXP17xx random_port support

cdp1337:


markjcrane:
The code has diverted too much to resolve the file conflicts.

- The grandstream_random_port has been added to the app_config.php file and the gxp17xx template.
- The smtp option for sendmail  was added at some point in the past.

PR #6062, opened 2021-10-11
[bind-digit-action]Fix wrong audio play on transfer

inoutglobal:
When transfering audio by using assisted *4 or *1 direct code feature, hold on audio and second tone are both playing incorrectly.


yois615:
@inoutglobal I just submitted a PR with a different fix that was merged.  Can you try with the fix in #6155 ?

yois615:
This PR can be closed without merge

PR #6063, opened 2021-10-15
Update ivr_menu_edit.php

udy786:
Because of this not able to put any phone number like 19876543210. It was allowing to type anything in Destination and submitted without any error but it was saving only menu-exec-app:


greenbea:
Simplest for your case is ```app/destinations/destinations.php?type=outbound```

For actionss other than `transfer` you can create a custom dialplan with that action and select destination to true on the dialplan, you'll get in the dropdown lst

For paid members there's a feature special for creating dptools.

markjcrane:
@greenbea shared better and more secure ways to handle this. Closing the pull request.

PR #6071, opened 2021-10-20
Updated wording for "Outbound Proxy"

tony1661:
The wording is updated so that the column header will also show the Proxy number.


tony1661:
@markjcrane Hi Mark, just wanted to bring this relatively small change back up since it may have gotten lost in the bunch of PRs you have to sort through.

This PR just changes the wording for the two "Outbound Proxy" servers to have a 1 or 2 at the end of it.

markjcrane:
I kind of prefer outbound proxy primary and secondary

PR #6072, opened 2021-10-21
Update switch.php

ednt:
The original outbound_route_to_bridge did not work correct with our legal and working outbound routes.

Since I don't know why a domain can not have a uuid I ignored this. Also the hostname is ignored.
If you need it, add it again please.

It was necesarry again after an update, because we were not able to send a fax.

I already did something similar for an older version, but it was ignored.
Here is the second try.

With this modified SQL, the code afterwards can be optimized, because there are always 2 records.
The first with the rule and the second with the bridge.

Best regards

### DIFF
--- a/resources/switch.php
+++ b/resources/switch.php
@@ -432,27 +432,20 @@ function outbound_route_to_bridge($domain_uuid, $destination_number, array $chan
 		$hostname = 'unknown';
 	}
 
-	$sql = "select d.dialplan_uuid, ";
-	$sql .= "d.dialplan_name, "; 
-	$sql .= "dd.dialplan_detail_uuid, ";
-	$sql .= "dd.dialplan_detail_tag, ";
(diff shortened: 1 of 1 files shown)


markjcrane:
The reason the delay in being accepted is the following statement. Hostname and domain_uuid.
```
Since I don't know why a domain can not have a uuid I ignored this. Also the hostname is ignored.
```

markjcrane:
Previous commit was ignored because the reasons for the changes were not explained. Looking at this a little more it appears you have an outbound route that is using the dialplan_detail_group and this is at least one thing you were trying to solve. This is not used by most people for outbound routes but it is a legitimate for dialplans and outbound routes. However this change will fix that issue and break 2 possibly 3 other things. One of them that would likely break many systems. 2 issues are easy to solve the third I need to do more testing to see if it is a concern.

PR #6095, opened 2021-11-05
Use effective_caller_id_name for origination_callee_id_name

yois615:
The current dialplan has the destination number used as the callee ID
name for display updates.  If the call is local, we should look for
effective_caller_id_name to make the information more complete.


yois615:
Close

PR #6096, opened 2021-11-05
Use effective_caller_id_name for origination_callee_id_name

yois615:
Currently, we use the destination number for
origination_callee_id_name.  If available, we should
use effective_caller_id_name so that display name
is available to the caller.


markjcrane:
It would probably be better to put this in the user_exists dialplan.

yois615:
My thought to put it here is so that we override the setting already being set in this file (see lines above this commit).

PR #6097, opened 2021-11-05
Add new Polycom provisioning default settings for polling

twilley:
Add new Polycom provisioning default settings, and update provisioning templates to support for additional polling options for supported versions (4.x, 5.x, 6.x)


markjcrane:
What is the purpose of the Polycom polling?

twilley:
The `polycom_provision_polling_*` settings tell the phones how to schedule automatic checking against the provisioning server for configuration changes. If these options are not used, I believe that the phones will only check for updated configurations at boot-time or if manually triggered (via on-device menu or device web-interface, for example).

More official description/documentation is available at https://community.polycom.com/t5/VoIP-SIP-Phones/FAQ-Automatically-check-for-configuration-changes-or-new/td-p/68444

Before this pull request, the following settings were already available, so the feature was partially implemented:
- `polycom_provision_polling_enabled` toggle the functionality on the phone
- `polycom_provision_polling_time` set a time of day when the phone will check-in. Without my additions, this option does not apply unless the phone has previously been configured to use the abs (absolute) polling_mode somehow.

The settings this pull request introduce are:
- `polycom_provision_polling_mode` sets the mode (strategy) the phone will use to schedule check-ins. The phone will default to `rel` mode, which ignores polling_time, and will use polling_period. I chose to use the default setting of `abs` to take advantage of the pre-existing default settings I already mentioned.
- `polycom_provision_polling_period` sets the interval that the phone will wait between check-ins. The phone will default to 1 hour if this setting isn't present (or already configured).
- `polycom_provision_polling_timerandomend` when using the `random` polling_mode, this, in conjunction with the polling_time setting (and sometimes the polling_period) will determine the time window for checking in. The phone will randomly choose a time within this window when it will check in with the provisioning server.

markjcrane:
Okay thanks for the description. In my opinion polling should be disabled by default. No need to update phones unless the provisioning is actually changed. However I don't mind it being an option for people that want it.

twilley:
I agree. 

I believe that a fresh install won't have provisioning enabled, plus the polycom_provision_polling_enabled setting is off by default.

Both have to be enabled by the PBX administrator before Polycom phones will even start to use the other polycom_provision_polling_* settings. And on any system where provisioning + polycom_provision_polling_enabled are both on, the new polycom_provision_polling_* settings will need to be turned on before they are applied.

Here is the configuration the phone will see if provisioning is enabled but the polling settings are default:
```
prov.polling.enabled="0"
prov.polling.mode=""
prov.polling.period=""
prov.polling.time="03:00"
prov.polling.timeRandomEnd=""
```

Here is what an unpatched configuration would look like in the same scenario:
```
prov.polling.enabled="0"
prov.polling.time="03:00"
```

twilley:
And here is the same configuration the phone sees on my test instance if I have configured each of the polling options:
```
prov.polling.enabled="1"
prov.polling.mode="rel"
prov.polling.period="3600"
prov.polling.time="23:20"
prov.polling.timeRandomEnd="03:00"
```

time and timeRandomEnd are ignored by the phone in this scenario, and the phone will check with the server for configuration changes every 3600 seconds starting at boot time.

markjcrane:
Thanks for the explanation I'm accepting this because its offers more functionality and its off by default.

PR #6104, opened 2021-11-09
Add grandstream_dnssrv_transport setting to GRP and GXP21xx templates

jonathanblack1000:
Add grandstream_dnssrv_transport setting.  This will enable selection of transport (udp, tcp, tls) when using dns/srv.  Added on GRP  and GXP 21xx templates. Before this, if you use dns/srv, it defaults to tcp.


markjcrane:
Please update the pull request.

jonathanblack1000:
Should be updated now.  Shows no conflicts.

PR #6130, opened 2021-11-19
[4.4] Fixes many SQL NOW() queries

daniel-lucio:
You have NOW()  AT TIME ZONE XXX which is not handled correctly. Happily, this only breaks Postgresql. MySQL/MariaDB users are safe.

### DIFF
--- a/resources/classes/database.php
+++ b/resources/classes/database.php
@@ -1489,8 +1489,8 @@ include "root.php";
 													if (strlen($array_value) == 0) {
 														$sql .= "null, ";
 													}
-													elseif ($array_value === "now()") {
-														$sql .= "now(), ";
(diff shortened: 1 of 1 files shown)


markjcrane:
Looks like this adds a vulnerability and fixes nothing int the code.

daniel-lucio:
Looks like you dont justify your statements.

markjcrane:
You mean like you didn't bother to justify this commit until I rejected it. The substr validation would cause a security risk. The code currently checks for valid information in this case an exact match now() and this commit changes it to a partial match with an open ended value after the now() it does introduced a vulnerability that I'm not willing to accept.

daniel-lucio:
What part of fixes "NOW() AT TIME ZONE" queries is not clear?

PR #6131, opened 2021-11-19
[4.5] Fixes many SQL NOW() queries

daniel-lucio:
Same as 4.4

### DIFF
--- a/resources/classes/database.php
+++ b/resources/classes/database.php
@@ -1787,8 +1787,8 @@ include "root.php";
 														if (strlen($array_value) == 0) {
 															$sql .= "null, ";
 														}
-														elseif ($array_value === "now()") {
-															$sql .= "now(), ";
+														elseif (strtolower(substr($array_value, 0, 5)) === "now()") {
+															$sql .= "$array_value, ";
 														}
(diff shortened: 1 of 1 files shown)


markjcrane:
I think there is a good chance your substring creates a security risk. As the now() used is exact and yours is open ended.

markjcrane:
There is nothing wrong with the current now().

daniel-lucio:
Well, if we are saying what we think: I think Fusion is full of bugs and you are too stubborn to accept a different vision.


Read your own code, you are not using "now()", there are someplace where you are using "now() AT TIME ZONE ....", your code tries anything that is !== "now()" as a string instead like a function or a proper SQL statement.

When you get a complaining with the FAX queue (and any other place using a modifier with NOW(), remember this PR.

markjcrane:
This one would introduce a vulnerability. My code is look for an exact match to now() and yours would validate the beginning of the string  now()  and then allow anything after that. Some months ago I accepted a pull request for the log viewer that used a substr to validate the beginning of the string only and it introduced a vulnerability. So this change would create a vulnerability. Your change to be safe would need to validate the entire now() string.

PR #6146, opened 2021-11-23
use application tag to enable firmware updates

twilley:
I've been struggling to get FusionPBX to host firmware upgrades for my polycom phones. I've found a solution that works for me and would like to merge my changes.

This adds four default settings, and some logic to the polycom 4.x, 5.x, and 6.x provisioning file templates to handle the settings.

The new default settings are:
- polycom_provision_app_include - a boolean to enable/disable the new code in the provisioning template. defaults to **disabled**.
- polycom_provision_app_file_path_4x - a text string representing a portion of a URL/directory path; only used in the 4.x provisioning template.
- polycom_provision_app_file_path_5x - a text string representing a portion of a URL/directory path; only used in the 5.x provisioning template.
- polycom_provision_app_file_path_6x - a text string representing a portion of a URL/directory path; only used in the 6.x provisioning template.

An Nginx rewrite rule is required to allow the phones to download the firmware. I may open a pull request on the fusionpbx-install.sh repository to get the rewrite rule included in new installations if there is interest. The rewrite rule I have had success with so far (this could probably be improved) is:

`rewrite "^.*/provision/polycom/firmware/(.+)$" /app/polycom/resources/firmware/$1;`

Firmware files should be obtained from Polycom and placed in /var/www/fusionpbx/app/polycom/resources/firmware/x.y.z/ where x.y.z matches the version number (by convention).

For example, If I want to use UC version 6.4.1 on my 6.x provisioned phones, I would unzip the firmware files into `/var/www/fusionpbx/app/polycom/resources/firmware/6.4.1`. I would then make sure that the polycom_provision_app_file_path_6x setting has a value of `6.4.1/`

Example scenario: when a VVX250 phone is provisioned, the phone will do an HTTP request for /app/provision/polycom/firmware/6.4.1/3111-48820-001.sip.ld, and the rewrite will pass the request to the local file stored at /var/www/fusionpbx/app/polycom/resources/firmware/6.4.1/3111-48820-001.sip.ld. The phone will download and install the firmware, rebooting in the process.

### DIFF
--- a/app/polycom/app_config.php
+++ b/app/polycom/app_config.php
@@ -550,5 +550,37 @@
 		$apps[$x]['default_settings'][$y]['default_setting_value'] = "04:00";
 		$apps[$x]['default_settings'][$y]['default_setting_enabled'] = "false";
 		$apps[$x]['default_settings'][$y]['default_setting_description'] = "Set the time to stop polling when using random mode";
+		$y++;
+		$apps[$x]['default_settings'][$y]['default_setting_uuid'] = "fb4fe110-bb83-4b64-a39b-c879ebcc8af7";
+		$apps[$x]['default_settings'][$y]['default_setting_category'] = "provision";
+		$apps[$x]['default_settings'][$y]['default_setting_subcategory'] = "polycom_provision_app_file_path_4x";
+		$apps[$x]['default_settings'][$y]['default_setting_name'] = "text";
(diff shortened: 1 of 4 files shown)


twilley:
It looks like the current firmware url method works fine. I had a configuration precedence issue due to how I was bootstrapping my phones that was causing my phones to ignore the configuration provided by the server.

PR #6148, opened 2021-11-23
enable customization of emergency numbers on polycom phones

twilley:
This adds three default settings, and code to handle them in the Polycom 5.x and 6.x provisioning templates.

The purpose is to allow for the Polycom phones to detect "non-standard" emergency numbers when dialed and treat them the same as when standard emergency numbers are dialed. For example, when 911 is dialed, e911 information is sent with the call. But providers can have e911 "check" numbers, such as 933, that will tell the caller their e911 info when dialed. This is useful as a method to test whether the e911 information is correct/working without actually dialing 911 and having to warn the local PSAP, etc.

Polycom phones don't send e911 info for normal calls, so this allows the Polycom phone to "learn" what numbers we want to consider as emergency numbers.

The settings I have added are:
polycom_emergency_customized - a boolean; disabled by default. toggles the functionality on/off
polycom_emergency_server - text; can/should be filled in with the hostname/ip of the server that will handle emergency calls. I fill this with the FQDN of my PBX.
polycom_emergency_port - numeric; default value 5060. The port of the SIP listener on the server.
polycom_emergency_number - an array; default values are 911 and 933. A list of numbers to be treated as emergency numbers by the Polycom phone.

Additional notes: adding the emergency numbers to your digitmap may be wise.


markjcrane:
Looks like a great improvement thanks for sharing!

PR #6155, opened 2021-11-24
Fix permissions on inband feature codes

yois615:
Using ${sip_authorized} is not effective since the inbound
call often has authorization to call into the PBX, either via
ACL or registration.  Instead, use ${from_user_exists} to test
if this user exists on the system.

We also fix a condition that the local user types *1 and the
remote user receives the dial tone.

### DIFF
--- a/app/dialplans/resources/switch/conf/dialplan/085_bind_digit_action.xml
+++ b/app/dialplans/resources/switch/conf/dialplan/085_bind_digit_action.xml
@@ -1,17 +1,19 @@
 
 <context name="{v_context}">
 	<extension name="bind_digit_action" number="" continue="true" app_uuid="7e007a4d-886b-4cdf-81f4-da9eb4689f1c" enabled="false" order="85">
-		<condition field="${sip_authorized}" expression="true" break="never">
-			<action application="set" data="bind_target=both" inline="true"/>
-			<!-- set to peer to prevent manipulate of call by caller -->
+		<condition field="${from_user_exists}" expression="true" break="never">
+			<action application="set" data="bind_target=self" inline="true"/>
+			<action application="set" data="bind_action_target=self" inline="true"/>
+			<!-- The caller here is remote so only assign feature codes to internal recipient -->
 			<anti-action application="set" data="bind_target=peer" inline="true"/>
+			<anti-action application="set" data="bind_action_target=peer" inline="true"/>
 		</condition>
 		<condition>
 			<!-- <action application="export" data="conf_xfer_number=xfer-${create_uuid foo}-${domain_name}"/>-->
-			<action application="bind_digit_action" data="local,*1,exec:execute_extension,dx XML ${context},${bind_target}"/>
-			<action application="bind_digit_action" data="local,*3,exec:execute_extension,cf XML ${context},${bind_target}"/>
-			<action application="bind_digit_action" data="local,*4,exec:execute_extension,att_xfer XML ${context},${bind_target}"/>
-			<!-- <action application="bind_digit_action" data="local,*0,exec:execute_extension,conf_xfer_from_dialplan XML conf-xfer@${domain_name},${bind_target}"/> -->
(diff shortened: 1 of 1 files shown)


inoutglobal:
Hi there,

Tried to use bind_digit_action with this merged fix  without success. If Extension A calls Extension B, then B are not able to place transfer with *1 and *4 code.
bind_digit_action

yois615:
Hmm, that makes sense looking at it.
We can probably set a condition that if from and to users exist then bind both with action on peer

yois615:
@inoutglobal Try this #6157

inoutglobal:
Hi, i started a lab to check this new fix approach. It seems to be good now! I think you nailed it. Both sides are able to transfer now, no reversed audio, it transfers ok both with direct and assisted transfer code  (*1 / *4).

But there are two things:

1- This is the working XML dialplan for my bind digit action LAB. Im not familiar with github stuffs, but it seems to be different from yours.

<extension name="bind_digit_action" continue="true" uuid="ab511d17-dccc-4ba8-9db1-787b12e193a5">
	<condition field="${from_user_exists}" expression="true" break="never">
		<action application="set" data="bind_target=self" inline="true"/>
		<action application="set" data="bind_action_target=self" inline="true"/>
		<action application="set" data="bridge_pre_execute_bleg_app=lua"/>
		<action application="set" data="bridge_pre_execute_bleg_data=bind_digit_bleg.lua ${context}"/>
		<anti-action application="set" data="bind_target=peer" inline="true"/>
		<anti-action application="set" data="bind_action_target=peer" inline="true"/>
	</condition>
	<condition field="" expression="">
		<action application="bind_digit_action" data="local,*1,exec:execute_extension,dx XML ${context},${bind_target},${bind_action_target}"/>
		<action application="bind_digit_action" data="local,*3,exec:execute_extension,cf XML ${context},${bind_target},${bind_action_target}"/>
		<action application="bind_digit_action" data="local,*4,exec:execute_extension,att_xfer XML ${context},${bind_target},${bind_action_target}"/>
		<action application="digit_action_set_realm" data="local"/>
	</condition>
</extension 

2- Is it possible to lower the interdigit-timeout so it can transfer faster? One workaround is dialing # at the end.

inoutglobal:
`<extension name="bind_digit_action" continue="true" uuid="ab511d17-dccc-4ba8-9db1-787b12e193a5">
	<condition field="${from_user_exists}" expression="true" break="never">
		<action application="set" data="bind_target=self" inline="true"/>
		<action application="set" data="bind_action_target=self" inline="true"/>
		<action application="set" data="bridge_pre_execute_bleg_app=lua"/>
		<action application="set" data="bridge_pre_execute_bleg_data=bind_digit_bleg.lua ${context}"/>
		<anti-action application="set" data="bind_target=peer" inline="true"/>
		<anti-action application="set" data="bind_action_target=peer" inline="true"/>
	</condition>
	<condition field="" expression="">
		<action application="bind_digit_action" data="local,*1,exec:execute_extension,dx XML ${context},${bind_target},${bind_action_target}"/>
		<action application="bind_digit_action" data="local,*3,exec:execute_extension,cf XML ${context},${bind_target},${bind_action_target}"/>
		<action application="bind_digit_action" data="local,*4,exec:execute_extension,att_xfer XML ${context},${bind_target},${bind_action_target}"/>
		<action application="digit_action_set_realm" data="local"/>
	</condition>
</extension`

yois615:
The problem with your way is with an outbound call to an external number... The callee can initiate a transfer

inoutglobal:
I got it! Yep, it is a problem. Does it can act only locally, i mean, between extensions?

yois615:
Yes, that is what I fixed in #6157 

If what I did there works for you marked it as approved so it gets merged

inoutglobal:
@yois615, sorry i messed all the things. Now it is working with your fixed code. Yes, it is approved for internal calls. I will try to place some external incoming/outgoing calls to test it. Thanks a lot!

image

PR #6157, opened 2021-11-26
bind_digit_action for B-leg during internal call

yois615:
The existing code only binds properly when one leg
is not internal.  Here we use LUA to bind the bleg
for internal calls.


inoutglobal (approved):
Yep, as i told you, I was able to test internal calls and now direct/assisted transfers are running good with bind digit action, no more reversed audios.

PR #6163, opened 2021-11-30
Retain recording display name and reset digit timeout

yois615:
This commit fixes 2 issues:
1. The recording name was not preserved upon
rerecording

2. When selecting rerecord, the digit timeout would retain
the value of 100, preventing the user from selecting the recording
id


yois615:
Anyone looking at this?

PR #6167, opened 2021-12-02
[xml handler] Fix error channel

greenbea:
using !USER_BUSY gives an error `Could not locate channel type !USER_BUSY`


markjcrane:
Thanks for providing the reason.

PR #6168, opened 2021-12-02
[feature_event] DND don't update follow me

greenbea:
* Don't update follow me and forward all
* Cleaned up some code


greenbea:
I'll put it together with the previous pull request

PR #6176, opened 2021-12-07
Better detection of trailing '/' in path

frytimo:
Currently the substr method may return an invalid path if the $recording_filename starts with a '/'. This will correct it by looking at the string in reverse.

### DIFF
--- a/app/recordings/recordings.php
+++ b/app/recordings/recordings.php
@@ -66,7 +66,7 @@
 				}
 
 			// build full path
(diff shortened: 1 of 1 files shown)


markjcrane:
From this link.
- https://www.php.net/substr
- $rest = substr("abcdef", -1);    // returns "f"
- That means your code would pick up the trailing / 
Purpose of the code is trying to determine if the forward slash is at the beginning of the string to see if its a relative path. If it has a forward slash on the first character of the string then its an absolute path.

PR #6178, opened 2021-12-09
Update app_config.php

rabbidiesel:
Add option to enable or disable remote phonebook lookup.


markjcrane:
A different variable was used.

PR #6198, opened 2021-12-16
Update destination_edit.php

KonradWebArray:
Changed SQL query that searches for duplicates to include destination_area_code.
Removed the recursive check for destination_number since prefix + area code + number are concatenated.

### DIFF
--- a/app/destinations/destination_edit.php
+++ b/app/destinations/destination_edit.php
@@ -167,7 +167,7 @@
 				$sql = "select count(*) from v_destinations ";
 				$sql .= "where (destination_prefix || destination_area_code || destination_number = :destination_number) ";
 				$sql .= "and destination_type = 'inbound' ";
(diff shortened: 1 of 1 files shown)


markjcrane:
Please signup as a contributor here.

KonradWebArray:
> Please signup as a contributor here.
> 
> * https://github.com/fusionpbx/open-source

Reviewed CLA and sent email. Thanks.

markjcrane:
I believe there is an issue with this code. For the current version and the with your changes.  I believe the query will fail because of the concatenation using || if the value of the field(s) are null the query fails.

The PostgreSQL function concat solves this problem.
https://www.postgresqltutorial.com/postgresql-concat-function/

markjcrane:
Email sent but also need to send the file as a pull request to this directory
- https://github.com/fusionpbx/open-source/tree/master/contributors

PR #6209, opened 2021-12-22
Time_zone feature added.

sergenkabakci:


markjcrane:
I'm not sure the purpose of this commit? Probably not aware that the time zone is being handled in the SQL query. To use it you need to set the time_zone variable in default settings.

PR #6227, opened 2022-01-05
[Yealink] add ringing_timeout variable

greenbea:


markjcrane:
Looks good thanks for the improvement.

PR #6229, opened 2022-01-06
[provision] don't show file path on not found

greenbea:


greenbea:
Besides revealing the path issue, the server should not return a 200 on a not found

PR #6230, opened 2022-01-07
[user_record] move back to record_session

greenbea:
use RECORD_ANSWER_REQ variable to start recording after answer


markjcrane:
Would be good to explain why as I've seen with going back and forth between record commands that we usually gain something and loose something. This the commit nor comments are stating the benefits to the change and they are needed.

greenbea:
1. `record_session` is cleaner
2. `api_on_answer` can only be used once per session. If it's being used in user_record dialplan, we lose the opportunity to use it elsewhere when needed.

I wish I could answer why there were changes back and forth on this.

emaktech:
I was about to make this exact same PR because many domains that were created after this was changed to api_on_answer have not been recording calls in a variety of conditions. Been manually updating dialplans on the affected domains back to the old record_session variable.

Can definitely vouch for this being at minimum more reliable as it stands right now.

PR #6232, opened 2022-01-07
[email validation] improve validation to conform the RFC

greenbea:
The current validation regex didn't conform with RFC 5322, but PHP `filter_var` does.
More information can be found at https://en.wikipedia.org/wiki/Email_address#Local_part

I also removed the unused function `email_validate`

### DIFF
--- a/resources/functions.php
+++ b/resources/functions.php
@@ -1064,13 +1064,7 @@ function number_pad($number,$n) {
 // validate email address syntax
 	if(!function_exists('valid_email')) {
 		function valid_email($email) {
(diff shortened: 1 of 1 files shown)


markjcrane:
My concern on this is to make sure it works on all operating systems FusionPBX supports. If I remember right there was an issue so I will need to test that to confirm whether it works or not on Raspberry Pi and on FreeBSD.

markjcrane:
Testing showed this works fine on the platforms I was concerned about. Accepting the changes.

PR #6233, opened 2022-01-10
Removing {$row.display_name} from the line button

tony1661:
The reason I think this should be removed is because setting the label on this model of phone removes the ability to see status of the line button.

For example (without the label), if I have a call on hold on my first line button, it will show the caller's number and the state of that call. For example: "4165551234 holding"

When you put the label, it removes that great feature. When the label is left blank, it defaults to "Line".

## Here is a pic of the current way things work and what the user sees with a call from extension 101 on hold
<details>
  <summary>Click to view image</summary>

</details>

## Here is a pic of the way I am proposing be default
<details>
  <summary>Click to view image</summary>

 </details>


markjcrane:
Sounds like a  good reason thanks for the unusually good explanation.

PR #6239, opened 2022-01-11
Update xml_cdr_statistics.php

vivoras:


markjcrane:
Don't see why there needs to be an extra line feed.

markjcrane:
I see why this is needed because of the extra double quote. Thanks for the change.

PR #6240, opened 2022-01-12
Always authorize vm when voicemail authorized=true

emaktech:
To major changes have been made:

* Removed redundant check for the existence of the voicemail_authorized variable since in this logic it is not needed
* Removed additional check to see if voicemail_id or user_name matches the vm box

Purpose being that it prevents use cases such as setting voicemail_authorized=true to disable password prompt with *95[ext] dialplan. In other words, it makes it possible to disable voicemail password check through dialplans under any condition, which I suspect is the intended purpose of the voicemail_authorized variable in the first place.

### DIFF
--- a/app/scripts/resources/scripts/app/voicemail/index.lua
+++ b/app/scripts/resources/scripts/app/voicemail/index.lua
@@ -383,16 +383,8 @@
 		if (session ~= nil and session:ready()) then
 			--check the voicemail password
 				if (voicemail_id) then
-					if (voicemail_authorized) then
-						if (voicemail_authorized == "true") then
-							if (voicemail_id == user_name or voicemail_id == sip_number_alias) then
-								--skip the password check
-							else
-								check_password(voicemail_id, password_tries);
-							end
-						else
-							check_password(voicemail_id, password_tries);
-						end
+					if (voicemail_authorized == "true") then
+						--skip the password check
 					else
 						check_password(voicemail_id, password_tries);
 					end


markjcrane:
It appears your changes would allow anyone in the same company to listen to anyone's voicemail without a password. For example dial *98 and then someone then enter voicemail id to listen to someone's voicemail. Also concerned that if someone opens up remote voicemail from calling into voicemail from the outside that this may open up ability to listen to anyone's voicemail from an external number. Currently there is a pin number that would be required in both of these cases.

emaktech:
> It appears your changes would allow anyone in the same company to listen to anyone's voicemail without a password. For example dial *98 and then someone then enter voicemail id to listen to someone's voicemail. 

In our case, this is the intended purpose. We sometimes configure a "shared voicemail" box for users where all extensions are essentially sharing the same voicemail box. The VM key is remapped using *95[ext] in our case and the users do not want to require a PIN.

Thus, this change makes it possible to set voicemail_authorized=true in the dialplan for this domain and removes the PIN requirement.

Essentially the purpose is to allow a user to choose to disable the PIN with this variable through a dialplan. If no dialplan is created to do this, a PIN will still be required as it always was.

yois615 (changes requested):
I understand your usecase but I don't see why this should be merged into the mainline project.  It puts a built-in security flaw that can be accidentally enabled without an admin fully understanding the consequences.

markjcrane:
Closing the pull request because I think it is a security risk.

PR #6248, opened 2022-01-20
Update {$mac}.cfg

garryavx:
I added a line to set SRTP encryption. the default setting yealink_srtp_encryption was not setting properly to my T56a devices. I downloaded my config files and did not find an entry for srtp at all.

compared with the t54w template and it did have a srtp setting. So I pretty much stole the line from the T54w template and dumped it in here.

this is dirty and tbh the t54w template is vastly superior to this one and this probably needs a rewrite to bring it in line.

tested on my T56a and it at least gets the job done.

### DIFF
--- a/resources/templates/provision/yealink/t56a/{$mac}.cfg
+++ b/resources/templates/provision/yealink/t56a/{$mac}.cfg
@@ -102,6 +102,10 @@ account.1.register_mac =
 account.1.register_line = 
 account.1.reg_fail_retry_interval = 
 
+#set encryption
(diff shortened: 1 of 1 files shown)


garryavx:
looking around i see a t5x template that looks much closer to the t54w one and includes the srtp value already...I re provisioned my t56a using it and it works great.

I think moving forward I am going to use this for all T56a, might try it out on the other T5 series phones as well to see if it is a viable universal replacement for the other templates.

linkandzelda91:
Mmmm this is probably doable by just using the t54w templqt as a reference I'll try it on Monday.

PR #6257, opened 2022-01-23
Dual Registration supported non published firmware files for EoL phones. 

kapetrosyan:
These are Dual Registration supported non-published by Yealink firmware files for EoL phones. (T41P, T42G, T27P, T29G, T46G, T48G)

P.S. This is my first contribution, I'm sorry if I made something wrong. This is not a functionality-related change that's why I decided to do this as my first contribution.


markjcrane:
We appreciate you sharing these files.

PR #6284, opened 2022-02-03
Yealink template consistency

tony1661:
Changes to make the templates a bit more consistent while also fixing some typos and creating a new variable for the NAT Update setting.

By default, the account.1.nat.udp_update_enable is set to "3" on most templates which means to have the phone periodically send NOTIFY messages. This is not the case though on additional accounts (e.g. account 2). This makes for inconsistent configs that sometimes cause issues with phones ringing.

This fixes the issue while also giving the ability to use a variable to set it.


markjcrane:
Need to add the setting to app/yealink/app_config.php for the default setting yealink_udp_update_enable

markjcrane:
I've added the default settings for yealink_udp_update_enable
- https://github.com/fusionpbx/fusionpbx/commit/47aa8e68af9aa52152e56d80cd7989d1f6f8f020

PR #6319, opened 2022-02-25
Removing invalid lines in Yealink t31g template, appears to be a dodgy concatenation

harryfoster:
Looks like a combined boot file and a config, minor template issue. Stops the Yealink t31g from provisioning.


markjcrane:
These do look out of place.

PR #6322, opened 2022-03-08
[4.5] Fixes the blank password issue

daniel-lucio:
When an old server has been upgrading fusion, it gets to the combination of having PHP5 + Fusion 4.5  And yes, there are a lot of these combos out there.

random_int function is not properly implemented.

I have been reported that creating an extension gets a blank extension, after researching I found the root cause.

PR #4460 tried to fix it properly, but you rejected it. This is an ugly patch following your style.

### DIFF
--- a/resources/functions.php
+++ b/resources/functions.php
@@ -2181,8 +2181,8 @@ function number_pad($number,$n) {
 
 //convert bytes to readable human format
 	if (!function_exists('random_int')) {
-		function random_int() {
-			return rand ();
+		function random_int($min, $max) {
+			return rand ($min, $max);
 		}
 	}
 


markjcrane:
Accepting the patch but what really needs to happen is those that still using PHP5 its time to upgrade. Really how long are you going to wait until upgrading? Servers should be on a minimal version of PHP 7.1 or higher though 7.0 would likely work.

PHP 7.0 released
- Released: 03 Dec 2015

PHP 7.1 
- Released: 01 Dec 2016

PHP Release History
- https://www.php.net/releases/index.php

It is not unreasonable to require people at very least be on PHP 7 or higher after so many years of its release. It is not safe to be so far behind on versions that are no longer maintained for security. If the system is running PHP 5 then likely operating system and host of other software versions are also behind.

daniel-lucio:
I do agree that PHP5 shouldn't be in production anymore, but I am not the sysadmin of those servers and can't judge why they won't upgrade. I just fix what I am asked to.

markjcrane:
I looked at the php documentation matched parameters mentioned. But it I’m certain they are not being safe by using such an old version. In my mind it’s lazy and irresponsible as it puts peoples servers at risk.

PR #6323, opened 2022-03-08
Fix Missed Call email when Follow Me is On

daniel-lucio:
When a user has the Email Alerts for Missed Calls set, it works
As soon as there is a Follow Me configured, the email stops arriving

This patch restores the desired behaviour


amanadlakha1 (approved):

markjcrane:
Sorry it took me a while to get around to accept this... It looks like a good improvement. Good job.

PR #6324, opened 2022-03-09
[extension settings] fix cache not deleted on toggle

greenbea:


markjcrane:
Looks like a good chance thanks for sharing.

PR #6330, opened 2022-03-20
[callcenter] Add setting DSN from config.lua file

greenbea:


markjcrane:
No longer have config.lua using config.conf

PR #6332, opened 2022-03-23
Delete duplicate enabled="true" statements

emaktech:
Most recent PR on this file duplicated enabled statements causing the dialplan to be invalid and not be generated when running App Defaults.


greenbea:
What's the point of having the mkdir in the dialplan?

markjcrane:
Last I checked FreeSWITCH didn't check to see if the directory existed and add it if its missing. So with adding a file to a different directory for the month and day this means we have to constantly add a new directory. Would be best if FreeSWITCH did that last I checked it didn't do this.

PR #6334, opened 2022-03-28
Add grandstream_idle_screen_server_path variable

MarvinSk:


markjcrane:
Can you update your pull request so that this only looks like the one change that it is rather than looking then all the Merge branch info?

PR #6339, opened 2022-04-08
Fix voicemail transcriptions not being pushed

junction1153:
When creating a new extension, under the Voicemail tab, Transcribe is empty. This was causing users to not receive VM transcriptions. Adding the above code will push transcribe = true or false (depending on the value in default settings) after creating a new extension.

### DIFF
--- a/app/extensions/extension_edit.php
+++ b/app/extensions/extension_edit.php
@@ -679,11 +679,10 @@
 											if (permission_exists('voicemail_local_after_email')) {
 												$array["voicemails"][$i]["voicemail_local_after_email"] = $voicemail_local_after_email;
 											}
-											$array["voicemails"][$i]["voicemail_enabled"] = $voicemail_enabled;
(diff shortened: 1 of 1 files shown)


app/extensions/extension_edit.php
```diff
@@ -886,6 +888,8 @@
 				$voicemail_file = $row["voicemail_file"];
 				$voicemail_local_after_email = $row["voicemail_local_after_email"];
 				$voicemail_enabled = $row["voicemail_enabled"];
+				$voicemail_transcription_enabled = $row["voicemail_enabled"];
```
junction1153, line 891: one last modification, please change $voicemail_transcription_enabled = $row["voicemail_enabled"]; to: $voicemail_transcription_enabled = $row["voicemail_transcription_enabled"];

junction1153 (commented):
line 891 one last modification, please change $voicemail_transcription_enabled = $row["voicemail_enabled"]; to: $voicemail_transcription_enabled = $row["voicemail_transcription_enabled"];

PR #6340, opened 2022-04-08
fix issue with device vendors

junction1153:
If mac address and template already exists in Devices page and you attempt to associate the mac address with the extension from the extensions page, the "vendor" is removed from the device, thus causing provisioning not to look at the device profiles settings of the mac address


markjcrane:
Probably needs both what happens if the device vendor is still empty after using the mac address because there is a new mac address vendor ID. I'll accept the pull request and then modify it to account for this possibility.

PR #6343, opened 2022-04-12
Add DND Variable to Yealink T54W Template

ggflow:
Resolve current issue with T54W template DND using the call forward all feature code. My other pull request adding the DND variable option will need to be accepted prior to this one.


markjcrane:
Thanks for the improvements accepting both pull requests.

PR #6355, opened 2022-04-29
Update 480_operator.xml

markjcrane:


markjcrane:
**How to update**
- To use it update your source code.
- Then search for all dialplans called operator and then delete them.
- After that Upgrade -> App Defaults.

PR #6364, opened 2022-05-05
Correctly align fields when using larger fonts

ggflow:
When using larger font sizes some of the menu items get out of alignment's this update solves that issue.


ggflow:
Will submit cleaner request

PR #6371, opened 2022-05-11
Added variables to grandstream configs for idle mute fuction

piajesse:
I have changed them all to the default of "0" and then added a
variable called "grandstream_idle_mute_function" that overwrites
the default option of 0

There seems to be older configs where 0 = no instead of dnd
so I've added grandstream_idle_mute_function_old for these configs


markjcrane:
Thanks

PR #6384, opened 2022-05-13
Update polycom configs each separate

rabbidiesel:


markjcrane:
Best work flow is create a new branch make a pull request from it when its accepted delete the branch. This way old changes don't show up in the pull request and have a clean pull request with only the changes each time.

Alternative is a new git clone add changes send pull request. Please resend and I will accept it if the pull request is cleaned up.

PR #6391, opened 2022-05-19
Snom templates - added per model wallpaper settings

tony1661:
Currently there is only one wallpaper variable used for all Snom phones. That works fine if you set a wallpaper on a per device level.

It doesn't work if you want to set the same wallpaper across the domain since some users may have different models of phone which might use a different resolution.

This PR hopes to fix that by maintaining the existing variable that users might be using as well as adding a few new ones for model specific wallpapers.

If a domain has just one model of phone they can keep using the old variable.


markjcrane:
I appreciate you being considerate of others with the backwards compatibility.

PR #6394, opened 2022-05-24
Add iphone style toggle to enable/disable ring groups destinations

k-flagman:


greenbea:
this is a useful feature, but this needs to be handled too in the ring group index.lua script to ignore disabled destinations

PR #6399, opened 2022-05-27
added dialplans to Snom templates plus some address book fixes

tony1661:
This PR adds functionality to the Snom templates to have multiple dialplans so that the phone will know when to automatically send the dialed digits to PBX.

For this to work, you can create a _Domain Setting_ variable for each dialplan entry. I did not create a _Default Setting_ entry since dialplans may vary greatly between countries and domains.

Below is an example of the _Generated XML_ the provisioning system will generate as well as an example of what _Domain Settings_ you would need to set.

### Generated XML
```
<dialplan e="2">
 <template match="2..." timeout="4" user="Phone"/>
 <template match=".........." timeout="2" user="Phone"/>
 <template match="1.........." timeout="2" user="Phone"/>
 <template match="9911" timeout="0" user="Phone" Rewrite="911"/>
</dialplan>
```

#### First Dialplan Entry

 - **Category:** provision
 - **Subcategory:** snom_dialplan
 - **Type:** array
 - **Value:** \<template match="2..." timeout="4" user="Phone"\/\>
 - **Order:** 000
 - **Enabled:** True

#### Second Dialplan Entry
 - **Category:** provision
 - **Subcategory:** snom_dialplan
 - **Type:** array
 - **Value:**  \<template match=".........." timeout="2" user="Phone"\/\>
 - **Order:** 001
 - **Enabled:** True


tony1661:
Let me know if I can explain anything about this PR

markjcrane:
My main questions is what is this? Is it a default setting that is currently not in the snom/app_config.php for controlling the numbers that can be called? If show an outputted line as an example.

    {foreach $snom_dialplan as $row}
      {$row}
    {/foreach}

tony1661:
hi @markjcrane this PR is to add dialplans to the snom templates so that after I do not need to click the ✅ button for the phone to dial out/send digits.

$snom_dialplan is not a default variable (it can be though). I didn't add it since currently no one uses this functionality and I didnt want to break anything.

$snom_dialplan is a variable (type:array).

What gets generated looks something like what is written below. The **bold** is what is hard-coded into the template. Each $snom_dialplan variable would contain what is not in bold.

```
**<dialplan e="2">**
 <template match="2..." timeout="4" user="Phone"/>
 <template match=".........." timeout="2" user="Phone"/>
 <template match="1.........." timeout="2" user="Phone"/>
 <template match="9911" timeout="0" user="Phone" Rewrite="911"/>
**</dialplan>**
```

As an FYI, I do plan to document this if accepted.

tony1661:
would it be possible to maybe make the values csv rather than the raw xml? @markjcrane

tony1661:
@markjcrane any idea how we can fix this?

PR #6405, opened 2022-06-06
feat(mod_callcenter): change agent state

ccppprogrammer:
If we change the agent's call status to Available/Available (On Demand), we must also change the agent's status to Waiting, otherwise the agent will not accept calls. In all other cases, the agent state must be Idle.


markjcrane:
Sign up here as a contributor and your pull request will be accepted.

markjcrane:
If you sign up as a contributor I will accept your pull request.

PR #6448, opened 2022-08-02
add option to save voicemail message as new

nktech1135:
This is my attempt at adding an option to the voicemail app that lets users save messages as new. I only set it up for english using the sounds and voicemail language files. I'm not sure how to translate it.
This is working for me on a test system.


markjcrane:
0 is used for the operator by default. Looks like 6 isn't taken. Button 6 has MNO on it which N could stand for NEW as a way to remember the button.

nktech1135:
6 is used in streamfile.lua for fast forward, and i'd like to keep that if possible. I am thinking of rewriting the message forward script to combine forward to ext and forward to email under option 8. This would potentiallhy free up option 9 to be reused.
I assume we need the forward to operator on option 0 for someone?

markjcrane:
4 and 6 do need to be preserved as they are used for rewind and faster forward. 5 is pause. These numbers can be used only when the voicemail recording is playing.

nktech1135:
So, i'm curious what number you would suggest? Would you support a rewrite of the forward options to free up 9 for this PR? or is there another number you'd rather see used?

nktech1135:
closing as invalid. I'll recreate.

PR #6458, opened 2022-08-18
Adding the Yealink Tone Country setting to all templates, and including two more minor revisions of handsets that have been released.

harryfoster:
As the title says, the following is added to each template:

voice.tone.country = {$yealink_voice_tone_country}

And the CP925 and T42U were released.


markjcrane:
Thanks for sharing your improvements to Yealink provisioning templates.

PR #6461, opened 2022-08-22
Fix fax_queue to Email PDF Instead of TIF

emaktech:
Variables `$fax_file_tif` and `$fax_file_pdf` are missing a period before the extension so file_exists is always evaluating to false. Thus, TIF was always sent. Adds the period back into the path so the check works correctly.

NOTE: `$fax_file_tif` is completely unused and can be removed from this code for cleanup.


markjcrane:
Good job finding this error. Changed the double quotes to single quotes around the missing period.

markjcrane:
Thanks for contributing the bug fix.

PR #6464, opened 2022-08-24
Fix PHP warning for PDO::setAttribute() requires 2 parameters

frytimo:
Database setAttribute needs 2 parameters.


markjcrane:
Thanks for sharing the improvement.

PR #6465, opened 2022-08-24
Split languages in to multiple files

frytimo:
Split languages in to files named with their language code while still maintaining compatibility.


markjcrane:
The app_languages.php was built for creating an array of the languages phrases used in the web interface. It was never meant to put PHP code beyond just creating the array. Second the phrases in the language files format is important and would be used across the entire system. So I would really need to like the change.

PR #6466, opened 2022-08-26
Language split 2

frytimo:
Another version of splitting files but requires changing the .


markjcrane:
A few comments on the changes in this pull request.

Changing from $text to $strings and bridge_name

From
`	echo "	".$text['label-bridge_name']."\n";`

To
`echo "	".$strings['Name']."\n";](url)`

This is quite a large change requiring change to every file that sends language phrases to the web interface and to the console. I think the project to do this is very large. Also if you haven't noticed I don't like getting caught in case sensitivity problems so I use lower case as standard. Don't see the reason to dump bridge_name for only name. If you had to bring together information for multiple features in a single page that seems like it could be a risk for using only the name.

frytimo:
The idea behind this was the English word needed is "Name" so request $strings['Name']. This makes requesting what goes on the page simple. I do not expect this to be pulled in its current form and only be discussed openly about the idea of simplifying the array by requesting the word or sentence needed directly instead of by a key.

PR #6469, opened 2022-08-27
Add Telekonnectors Galaxy 1000 Plus template

JoyceBabu:
This PR adds template for auto provisioning Galaxy 1000 Plus 4 SIP line model from Telekonnectors.

I have re-used the variables from Galaxy 1000 template.

I have also added support for configuring line and memory keys. The function values are same for both Galaxy 1000 and Galaxy 1000 Plus. I will create another PR to add this feature to Galaxy 1000 too.


markjcrane:
Thanks for sharing this!

PR #6473, opened 2022-09-12
Update y000000000068.cfg

zrd740:
Fixed a typo in call pickup variable setup for T41S to make it consistent with other yealink phone templates


markjcrane:
Subtle issue was on the wrong line. Accepting your changes.

PR #6475, opened 2022-09-13
Frytimo patch fix send custom cache event

frytimo:
cache method file does not initialize the variable $fp before requesting the custom event to be sent. This initializes the $fp event socket stream to send the command.


frytimo:
Looks like I did not have the updated source before doing a PR

PR #6476, opened 2022-09-17
add option to save voicemail message as new

nktech1135:
This should be fully functional, we just need to decide which key to assign the function to.


markjcrane:
Looks good there are just the two concerns determining the number to use and improving the phrase upgrade process which right now is manually done.

PR #6478, opened 2022-09-19
added tr069 enable and periodic_inform_enable vars to config template…

nktech1135:
…sadd variable on dp750 template for enabling tr-069 provisioning.
I'd also like to add these to the appconfig.php file in the grandstream app but don't know how you generate settings uids?
Can you give me a rundown on generating uids? and i'll add those before considering this PR complete?


markjcrane:
This pull request has a bug. The closing tag is missing the final > greater than sign. If you update this and fix the bug I will commit the change.
<P4506{$grandstream_tr069_periodic_inform_enable}</P4506

PR #6482, opened 2022-09-30
Update blf_subscribe.lua

levishores:
This PR fixes the WHERE clause of the SQL query used to determine if there are new voicemails that should count towards the mailbox BLF being 'true'. This SQL was mimicked from the voicemail app's main_menu script: voicemail/resources/functions/main_menu.lua

In addition, in case the {scripts_dir}/run subfolder doesn't exist, the command to create it was added.


markjcrane:
Accepting your changes. Thanks.

PR #6484, opened 2022-10-12
remove is_numeric check for voicemail id

frytimo:
Remove is_numeric check for voicemail id when creating the directory.


markjcrane:
Likely to have multiple problems when using a non numeric voicemail ID. Don't expect calling it over *98 to work for example.

PR #6496, opened 2022-11-24
[5.1] Fix Timeconditions (and other apps) with custom values

daniel-lucio:
I found this in the time condition app (but it will apply to any that uses the class destinations).

When the destinations::valid() function is called, it only checks against the static values preventing any valid custom value from being added, as a consequence, it is nulled.

This patch allows you to be more flexible (you may or may not want custom values)

### DIFF
--- a/app/destinations/resources/classes/destinations.php
+++ b/app/destinations/resources/classes/destinations.php
@@ -989,7 +989,7 @@ if (!class_exists('destinations')) {
 		/**
 		* valid destination
 		*/
-		public function valid($destination, $type = 'dialplan') {
+		public function valid($destination, $type = 'dialplan', $strict = false) {
 			//allow an empty destination
 			if ($destination == ':') {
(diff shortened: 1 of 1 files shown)


markjcrane:
The valid function was built to increase security. This commit would neutralize that security improvement. If you want to add custom destinations use the dialplan_tools feature that is here on github.com/fusionpbx

Side note version 5.1 version doesn't exist and hasn't been released yet. There is 5.0.1 there is 5.0 and there is 5.0.5. Version 5.1 has not been released yet.

daniel-lucio:
I know exactly how to do it by myself. I have several newbie customers that are struggling to add custom values by themselves. This is meant to make it easier.

Also, it looks like you just reject any PR I send by instinct, if you had read you would see that there is a $strict parameter, that may be adjustable to allow this flexibility or not.

PR #6497, opened 2022-11-24
[5.0] Fix Timeconditions (and other apps) with custom values #6496

daniel-lucio:
Same as 5.1


markjcrane:
The valid function was built to increase security. This commit would neutralize that security improvement. If you want to add custom destinations use the dialplan_tools feature that is here on github.com/fusionpbx

Side note version 5.1 version doesn't exist and hasn't been released yet. There is 5.0.1 there is 5.0 and there is 5.0.5. Version 5.1 has not been released yet.

PR #6513, opened 2022-12-21
Support updating optional installed apps via git

demonspork:
This adds the ability to detect and update any application in the app/ folder that is a git repository with a remote origin URL. It will list the installed repositories and you can select which apps to update.

I can also be used via `php /var/www/fusionpbx/core/upgrade/upgrade.php apps` from the command line except it updates all at once.

The command line version will work with apps over ssh which is often the case with private repositories/internal code.

image

image


markjcrane:
My hesitation on this is that its doesn't use the git library we have been using in the Upgrade -> source. Instead this depends on pulling git via commands.

demonspork:
Upgrade -> Source doesn't use a git library, it executes the git commands as well.

PR #6514, opened 2022-12-22
Update local_extension fail_on_single_reject

demonspork:
Update local_extension fail_on_single_reject to handle stale registrations and other failures. Previously any failure mode would cause the call to drop to all.
So if a mobile app has a stale registration that times out when called, that leg fails with RECOVERY_ON_TIMER_EXPIRE and all other legs are cancelled and sent to voicemail.

This change handles 3 situations. All others should not fail the entire call to voicemail/forwarding.

1.  CALL_REJECTED - End User Rejects call - this means they hit cancel/end on their mobile app or desk phone. This reason should handle almost all phones. There may be edge cases that a phone doesn't conform and we can add those if discovered
2.  USER_BUSY - endpoint decides user is already on the phone - since we assume that an extension is a single person in most use cases, if an endpoint is configured with no call waiting and responds with a user_busy response, we should stop ringing and send the call to voicemail.
3.  NORMAL_CLEARING - Might not be necessary, but won't hurt anything. Should handle any "Confirm" options if they end up going through the local_extension dialplan, which I'm not sure they do.

This change might also be required in the call_screen dialplan.


markjcrane:
Thanks for sharing good details on your pull request.

PR #6515, opened 2022-12-29
[call_center] Add queue callback functionality

yois615:
* This commit adds the ability to create a queue callback profile and assign it to a queue.  Once
assigned, the caller can break out of a queue by
pressing 1 to request a callback.  A LUA script
runs as a daemon to process the pending callbacks.

* This commit also integrates position announcements into the queue with a modified version of the sample script in the FreeSWITCH docs.  We query the callback table to include pending callbacks in the position calculation.  We also include an annoucement advising the user of the option to request a callback if there is more than 1 person ahead of them in the queue and callbacks are enabled.


md-riaz:
@markjcrane , Can you explain the call center position annoucement option is available but not the related lua script ?
image

PR #6524, opened 2023-01-04
Updated device_edit.php 'Save' to be a submit function for the form. 

harryfoster:
The reason for this is the default behaviour for 'return' or 'enter' is to delete the device entirely. This makes using a barcode scanner impossible, or user error resulting in a lot of lost time.

### DIFF
--- a/app/devices/device_edit.php
+++ b/app/devices/device_edit.php
@@ -883,7 +883,7 @@
 			echo button::create(['type'=>'button','label'=>$text['button-delete'],'icon'=>$_SESSION['theme']['button_icon_delete'],'name'=>'btn_delete','onclick'=>"modal_open('modal-delete','btn_delete');"]);
 		}
 	}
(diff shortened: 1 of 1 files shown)


harryfoster:
Import function works flawlessly, always has. We use it during the initial installations of sites.

This is for when modifications happen after the initial install. Today, someone deleted an executives handset with a ton of speed-dials on the template on the BLFs and Sidecar as they went to do this:

1. Find the MAC address of the phone they are replacing due to damage.
2. Clicked on the MAC address field at the top, used their scanner they always have plugged in for the replacement handset. The scanner always inserts a carriage return after scanning.
3. They see 'a green bar' thinking it's saved it. They then realise the device is gone after a few attempts at provisioning it. What the green bar actually stated was 'OK, device deleted'.
4. I then have to delve into a backup to grab the speed dials that have accumulated over the course of a few months.

Currently, the top bar of buttons in the device form are all of type 'button'. This means that the default submit button is selected by the browser, and is often the first in the DOM, in this case it was the delete button. Changing the safest and sanest button to type 'submit' means the browser uses save as the default functionality for a return key or enter press.

There is also no additional modal safety as the it is only called 'onclick' of the delete button and not 'onsubmit()' which would be annoying for each edit.

PR #6534, opened 2023-01-11
[dashboard] Don't show cc agent for missed calls

greenbea:


markjcrane:
This has actually been accounted use the link below to search the xml cdr import code search for **missed_call** or  **and (cc_side is null or cc_side != 'agent')**
- https://github.com/fusionpbx/fusionpbx/blob/master/app/xml_cdr/resources/classes/xml_cdr.php

greenbea:
That's for the CDR list page, not for the dashboard.

markjcrane:
If the CDR has been imported and correctly states cc calls not as a missed call then the dashboard uses the missed_call field in the CDR table then this would already be accounted for.

markjcrane:
The proper place to handle this is when the CDR is being imported into the database. The reason this is best is because it means all other places that need missed calls can do it by depending on the missed call field in CDR. A more efficient way to do it and centralizes the code that determines what is a missed call.

greenbea:
You have a point.

greenbea:
The CDR currently doesn't handle cc_side == agent

markjcrane:
Addressing this in the CDR import.
https://github.com/fusionpbx/fusionpbx/commit/e3b600eaff4bc1de8efc0be1f781d8036639e08e

PR #6536, opened 2023-01-12
Remove Ring Ready from Lua and support disabling it in dialplan

demonspork:
ring_ready is already being used in the dialplan right before the lua script is being called. I couldn't find a reason for it to be in both locations.

This change enables you to disable sending "180 ringing" whenever a ring group is called. Freeswitch will usually send a 183 session progress with early media while processing our ring group calls and we found an endpoint that breaks if you send both 180 and 183, so we had to disable sending 180. This just makes it a setting.

If we want to control ring_ready in the lua file, we have to read the database another time to get the default setting value so I opted to remove it from the lua so we aren't reading the database more than we have to.

### DIFF
--- a/app/ring_groups/app_config.php
+++ b/app/ring_groups/app_config.php
@@ -177,6 +177,14 @@
 		$apps[$x]['default_settings'][$y]['default_setting_value'] = "3";
 		$apps[$x]['default_settings'][$y]['default_setting_enabled'] = "false";
 		$apps[$x]['default_settings'][$y]['default_setting_description'] = "";
+		$y++;
+		$apps[$x]['default_settings'][$y]['default_setting_uuid'] = "50cbb1bb-3d67-4320-9f7e-0b09aa09676d";
+		$apps[$x]['default_settings'][$y]['default_setting_category'] = "ring_group";
+		$apps[$x]['default_settings'][$y]['default_setting_subcategory'] = "ring_ready";
+		$apps[$x]['default_settings'][$y]['default_setting_name'] = "boolean";
+		$apps[$x]['default_settings'][$y]['default_setting_value'] = "true";
+		$apps[$x]['default_settings'][$y]['default_setting_enabled'] = "true";
+		$apps[$x]['default_settings'][$y]['default_setting_description'] = "execute ring_ready to send 180 ringing to calling party";
 
 	//schema details
 		$y=0;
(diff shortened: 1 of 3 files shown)


greenbea:
I guess it's not an enterprise originate. How can I reproduce the scenario to get a 180 and a 183 in the same call?

demonspork:
I can reliably reproduce it with a simultaneous ring group with a couple of members. I'm pretty sure I also was able to get it with an Enterprise ring group. I won't be able to sit down and test again until next week, but I'm already running this patch in production to fix an issue with a ridiculous IP speaker that would freeze up and require a hard reboot every time it got both 180 and 183.

greenbea:
I'm not able to reproduce this. I see freeswitch sends the 183 response when calling ring_ready before originating and doesn't pass the originated 18x response to the originator. Are there any other settings you have that cause this or a specific scenario?

demonspork:
Simultaneous - Internal calls only. I stopped by the office again because I was doubting my entire yesterday of troubleshooting this. Calls from the outside aren't doing it, but calling internally to a ring group seems to cause it for simultaneous Ring Groups. I guess a simpler fix for this particular scenario would have been to change to Enterprise.

I'm almost to the point I want to make a default setting that hides Simultaneous completely so we only have Enterprise. Every time there's a problem, Simultaneous seems to be the problem.

greenbea:
If you have recording enabled, freeswitch will pre-answer and send a 180, and when the ring group rings, send a 183.

greenbea:
If the session is pre-answered ring_ready won't do anything. I had recording enabled that caused freeswitch to pre-answer, if I disable the recording I get a 180 and a 183

demonspork:
So to avoid getting both the 180 and 183 we need to predict if there will be a recording at some point _after_ the ring group. So we could handle most cases of this particular endpoint problem by checking in the ring_group lua code if the user that is being targeted by the ring group has call recording enabled and don't ring_ready before sending the branches/forks of the call that will result in a 183. 

There still could be edge cases, and that's where we could leave the new setting I already added in this PR would come in handy so you could force ring groups to always trigger ring_ready

markjcrane:
Reviewing this again I believe this is a good change that allows controlling whether ring_ready is used or not.

I'm sorry I took so long reviewing this that I was unable to accept the pull request because of version conflicts. For this reason, I committed it directly with updates to work with the latest code.

 Ring Ready support disabling it in dialplan #6536 
- https://github.com/fusionpbx/fusionpbx/commit/5b77e3d62bf8206ae7df2f791afbca85f5cfb438

PR #6540, opened 2023-01-21
user_exists: Add effective_caller_id_name and effective_caller_id_number

emaktech:
Adds 2 new variables to the user_exists dialplan:

* extension_effective_caller_id_name
* extension_effective_caller_id_number

This allows us to access these channel variables from the event socket and get the caller id name and number of the dialed extension when an inbound call comes in to that extension.


markjcrane:
I'll accept your change but I will be changing the names
-     extension_effective_caller_id_name to extension_caller_id_name
-     extension_effective_caller_id_number to extension_caller_id_number

PR #6550, opened 2023-02-14
First Step for UDR = CDR + MDR + *DR

daniel-lucio:
One of the things I do not like is that there are independent detailed records for anything. the CDR app is powerful enough to handle everything.

This first step opens the door for any other app to push a detailed record.

Developers: just make sure to build the right XML as freeswitch does when doing the POST.

Screenshot_20230214_120419


demonspork:
An alternative to this approach would be to simply have the list pull multiple record types from the DB, then when you click through on a record it loads the record type specific page, either the xml cdr or the sms or any other type of record you want to include in the unified detail record list. 

Might give us more usability and extensibility without trying to shoehorn different records into the bespoke xml_cdr tables that are built assuming SIP voice calls.

markjcrane:
UNIX philosophy for software design is Do one thing and do it well. I believe long term management and preventing bugs is easier when following this philosophy.
- https://en.wikipedia.org/wiki/Unix_philosophy
- https://klarasystems.com/articles/unix-philosophy-a-quick-look-at-the-ideas-that-made-unix/

markjcrane:
I'll consider it further as time permits.

PR #6553, opened 2023-02-14
[security] import pages xss prevention

greenbea:


greenbea:
instead of escaping `$line_field` at variable assignment we can do it just for the `echo`

markjcrane:
Yes that is another solution that would work. I went with preg_replace to ensure no risky characters could be used.

PR #6563, opened 2023-02-23
Update Install.php

jrmcclean:
Created an option for Windows directory paths.  Changed PHP_BINDIR to PHP_BINARY.  PHP_BINDIR is a value set on compile time, not on runtime. PHP_BINARY is a value set on runtime (when the script is executed) PHP_BINARY is the most accurate for this particular use case as some dev may not use the same directory for compile vs runtime.  (Tested this function on Windows and Linux ( Debian 11 and Ubuntu) and it works without issue)  Create a  variable called $php_dir  that will store PHP_BINARY info . This will enable the variable it to be changed per respective OS.  Created a variable called $phpcache_location so the cache location can be changed per respective OS)


jrmcclean:
Created an option for Windows directory paths. Changed PHP_BINDIR to PHP_BINARY. PHP_BINDIR is a value set on compile time, not on runtime. PHP_BINARY is a value set on runtime (when the script is executed) PHP_BINARY is the most accurate for this particular use case as some dev may not use the same directory for compile vs runtime. (Tested this function on Windows and Linux ( Debian 11 and Ubuntu) and it works without issue) Create a variable called $php_dir that will store PHP_BINARY info . This will enable the variable it to be changed per respective OS. Created a variable called $phpcache_location so the cache location can be changed per respective OS)

markjcrane:
I fixed the minor things I saw wrong with this but other concern is with this commit is the config file path.
$config_path = $_SERVER['DOCUMENT_ROOT'].'\resources';

Kind of doubting the best place on windows for the config.conf file would be in the web directory. Reason this is a concern is the web server would need to read the file but not deliver it to anyone if they request it in the URL. There should be a better place on windows for the config file. Maybe in the program files location?

PR #6573, opened 2023-03-08
Remove redundant SQL condition

daniel-lucio:
In this condition you already have  p.dialplay_uuid = s.dialplay_uuid  and you have p.dialplay_uuid = CONST, no need tor s.dialplay_uuid=CONST,

if a = b, and a = c, then b = c

In small databases, this could be trivial; but in significant databases, this saves a lot of processing.


markjcrane:
Rather than commenting out the line I updated your patch to delete the line.

PR #6581, opened 2023-03-17
[database] Handle NULL literal in params

yois615:


app/scripts/resources/scripts/resources/functions/database.lua
```diff
@@ -41,6 +41,7 @@ local function apply_params(db, sql, params)
 
   local str = string.gsub(sql, param_pattern, function(param)
     local v, t = params[param], type(params[param])
+    if "NULL"    == v then return 'NULL'           end
```
yois615, line 44: This was done this way to not break what's currently on line 48, although I cannot find a scenario where that actually does anything

markjcrane:
LUA uses nil rather than null for the keyword representing null. Not sure why you would want to return a 'NULL'  as a string.

markjcrane:
I'm not sure of the purpose for this.

yois615:
Example:
I have a custom LUA script to collect information from a caller.  Depending on their entries, they may not have any input for a specific value.

When INSERTing their responses into the database, because the param is `nil` it errors out with "cannot bind parameter".  If I deliberately want to leave this "NULL" in the database there is no way to do so.

By not quoting the string literal NULL, it will actually set the value in the DB as NULL instead of "NULL" (which is a string).

yois615:
Usage:

local params = {
            caller_id_name = caller_id_name,
            caller_id_number = caller_id_number,
            domain_uuid = domain_uuid,
            gender = gender or 'NULL',
            age = age or 'NULL',
            zip_code = zip_code or 'NULL'
        }

markjcrane:
Thanks for the explanation accepting your changes.

PR #6587, opened 2023-03-23
New Snom phone support plus fix for older snom model

tony1661:
Some models of phones default to setting the privacy_out parameter to "on".

This causes:

1.  **From:** header to be sent as _anonymous@anonymous.invalid_
2. The **Privacy:** header to be set as _id_

This causes some carriers to send out calls without caller ID since Freeswitch passes the **Privacy:** header to the carrier.

On newer models, the privacy_out parameter is set to off which is likely why this issue has not come up however the D715 in particular defaults this to on causing issues.

I have put this setting on all models however just to be sure it is set.


tony1661:
I've added another commit containing templates for the Snom M100 and M500 KLE (Key Line Emulation) phones.

PR #6589, opened 2023-03-24
CLI scripts support

daniel-lucio:
CLI scripts do not have the $_SESSION filled as the session starts in the php.php file. WEB pages do have this property filled as it is filled later in the PHP execution.

This patch prevents any CLI script to fail.


yois615:
I'm concerned about the security implications of this change.  How are you certain $_SESSION is always set in web connection?

PR #6609, opened 2023-04-13
Increased the Textbox width for device settings

tony1661:
This PR is to increase the width of the device settings textbox widths to make it easier to edit. Here is the before and after screenshots.

settings


tony1661:
I've also increased the box for the proxy IP address since it was very small and cutoff. Below is the before and after.

image1282

PR #6628, opened 2023-04-27
T46U: enable/disable codec settings for all accounts

kshah37:
The template for codec enable settings was taking a value as 1 if it was set in the settings. It was not taking the actual value inside it. So I have edited the templates to check the value if it's set to true.


markjcrane:
Accepting your changes. Thanks for sharing.

PR #6647, opened 2023-04-30
[device] Add emergency caller ID to device

yois615:
In a hotdesking environment, it is essential
to keep track of devices to properly
route 911 calls to the proper PSAP.

We create a field to popluate an emergency
caller ID number that will be looked up when an alternate device is set for this station.


markjcrane:
This is an interesting idea. Worth considering further as time permits.

PR #6663, opened 2023-05-04
Fix sort on call broadcast display

jonathanblack1000:
Fix default sort order


jonathanblack1000:
Sorry, I didn't realize it would put both commits together on one pull request.  I'm not sure how or if I need to fix it.  They are unrelated to each other, but I did intend to submit both.

PR #6664, opened 2023-05-05
Yealink w70b

rabbidiesel:
Fixe account 2 so it should update correctly.


markjcrane:
Had a hard time sorting out whether this was correct. So copied changes from from the current template for account.1 to account.2. Then changed the number from 1 to 2. So that they would match.

PR #6672, opened 2023-05-08
use a glob that is POSIX compliant to be more compatible with different OSes

frytimo:
The current GLOB_BRACE is not compatible with strict POSIX compliant operating systems such as Alpine Linux. This PR will allow the current FusionPBX framework to be used with such operating systems for a wider adoption.


markjcrane:
Accepted your pull request that used __dir__ as it was an even better method than this approach.

PR #6697, opened 2023-05-16
Update xml_cdr_statistics_inc.php

daniel-lucio:
On some special use cases (especially when acting as class 4), the call_direction is not set. Using a default from a default setting is better than having a // = incoming value.


markjcrane:
When adding a variable like stats_default_direction please make sure to add it to default settings in the app_config.php file.

markjcrane:
It might be better to set this during the Call Detail Record import.

PR #6713, opened 2023-05-21
Create UUID object

frytimo:
Allows a uuid to be treated as both an object and a string. When treated as a string the UUID is already validated. When typed as a param in a function there is no need for the function to do any further validation ensuring the string is validated only once.


markjcrane:
This commit uuid class implements Stringable

PR #6715, opened 2023-05-23
Update {$mac}.cfg

garryavx:
my T48s was ignoring my codec settings in default settings. after digging into the provisioning files I think I found two issues:

First: the old definitions for if a codec is enabled checks if the variable for enabling it is set, and then if it is sets it to 1, or 0 if it's not set at all

the problem here was I have a variable called  	yealink_codec_g729_enable , which takes a value...if I set that value to false, it was still setting the codec to 1. I think it was saying "well, the variable ISSET, so I should set this to 1" even though the variable was set to FALSE.

SECOND: the variable itself:
 	yealink_codec_g729_enable
was a boolean. when fusionpbx fills out the config file for the phone to download, it literally puts "false" in the field for the variable if I fill with it. yealink seems to ignore this, it requires a 1 or a 0 in the firmware my phone is on (latest public  66.86.0.15 release)

I uh...I'm not really sure where to change the default setting in the fusionpbx github, but I've just changed it on my system to be numeric and set it to 0 or 1.

I will add the T48G config file to this request as well, as I've tested on one of them as well with these changes

### DIFF
--- a/resources/templates/provision/yealink/t48s/{$mac}.cfg
+++ b/resources/templates/provision/yealink/t48s/{$mac}.cfg
@@ -249,91 +249,103 @@ account.1.bla_subscribe_period = 300
 #The payload of the specified codec.
 #account.1.codec.Y.rtpmap =
 
-account.1.codec.1.enable = {if isset($yealink_codec_pcmu_enable)}1{else}0{/if}
+#by default lets leave PCMU enabled, and stick it at the top of priority, since it is the oldest and most common codec. but if the user has specified the variables, use tht instead!em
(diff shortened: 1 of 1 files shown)


garryavx:
this probably affects the other yealink phones but really all I've got in inventory to test is the t48s/g, gotta work with what we got

garryavx:
oooh wait hang on I found a t46s lemme check that one too

garryavx:
okay, got the t46s/g as well

markjcrane:
I still need to determine if I agree with your changes haven't had time to think about this yet.

garryavx:
roger

markjcrane:
I believe this was addressed in this commit.
https://github.com/fusionpbx/fusionpbx/commit/4eb9892256f85f53b9b0f752004a897830f08c58

PR #6736, opened 2023-06-02
Add order # to parent menu items dropdown selector

chansizzle:
This is a way to help fix the issue when you have multiple parent menu and child menu items with the same name. example: should you add a custom Applications parent menu with custom permissions and you install the members' application manager, you will then have 3 "Applications" menu items to choose from when adding/editing menus. Rather than choose one and hope you have selected the correct "Application" menu item, and repeat until you get it correct, this method should reduce the blind selecting.


markjcrane:
Your idea seems okay but after thinking about it more we don't currently support Menus that are only Parent and Child currently. Decided only to show the parent menus since FusionPBX is not supporting grand children.
- https://github.com/fusionpbx/fusionpbx/commit/1f0d24a447d9ea850a47dff628f2bb63d598281f

PR #6752, opened 2023-06-12
[FAX Queue] - correct german translations

AndyX90:
The current translations are a bit unusual.


markjcrane:
Thanks for improving the German translations

PR #6758, opened 2023-06-20
Provisioning per vendor

frytimo:
Move setting for phone vendors to their own default settings category so provisioning section is more focused on global provisioning settings.


markjcrane:
I don't like this approach. Because it isn't concise and doesn't seem clean. A concise and simple category example would be yealink or provision.

I don't like this naming convention of yealink_provision either as noted last night. Many of the categories in FusionPBX match the application they come from. So that would mean that the settings for app/yealink belong in category yealink. In this case they are added to category provision because they are for provisioning the device. I know the list can get longer so default setting we could add a app_uuid and filter by that. Or the provision settings could get added to the yealink, polycom, snom, htek and other categories with another field that indicates its for another purpose such as provision settings.

PR #6763, opened 2023-06-27
Updated Snom D717 template for setting the Smartlabel display mode

mgajjar19:
By Default, Snom D717 has the smartlabel display mode set to "short". Snom D717 has an issue where for short labels it doesn't take the label that we configure in devices. Instead, it tries to look for the value in the system and if it doesn't find a contact or extension associated, it will just show the number itself.


markjcrane:
Sign up as a contributor and your pull request will be accepted.

PR #6767, opened 2023-06-30
Random password function

frytimo:
Create random password using given string as random_bytes will offer bytes instead of characters


frytimo:
generate_password function already exists.

PR #6771, opened 2023-07-11
added delayed paging

tony1661:
This PR will add a feature to the paging function of FusionPBX.

This feature allows you to add _delay=true_ to the paging dial-plan to make it so the system will record an audio clip, then disconnect the caller, then play the audio clip over paging.

This was originally requested by a customer who has a speaker beside their work station for overhead paging and was getting feedback and echo when paging overhead. This solves that.

I made sure to make it compatible with users who do not have the _delay_ variable configured in-case their dial-plan is not updated.


markjcrane:
If the blocks freeswitch then it dangerous to registrations, scaling and possibly audio. It might be better to use sched_api to schedule the command to run after a certain number of seconds

tony1661:
Hi @markjcrane ,

Thanks for your feedback. I finally had some time to implement your suggestions. Please let me know if I can make any other improvements.

PR #6773, opened 2023-07-13
Add the / character to the ALLOWED_CHARS

jacobbuscher-bt:
Add the / character to the ALLOWED_CHARS variable to fix bug with front slashes being stripped.

Previously was turning /var/lib/freeswitch into varlibfreeswitch


markjcrane:
Good point.

Can you sign up as a contributor here.

Might be unnecessary for your current pull request but would be great for future ones. After you do that I'll accept your pull request.

PR #6777, opened 2023-07-20
Update {$mac}.xml

LEMTEL-krooney:
Modify firmware upgrade xml settings so the device will actually pull firmware from server


markjcrane:
This was done in this pull request.
- https://github.com/fusionpbx/fusionpbx/pull/6778

PR #6782, opened 2023-08-02
Finalize changing origination_callee_id_name from destination_number to caller_destination in domain_variables and variables dialplan.

jacobbuscher-bt:
Finalize changing origination_callee_id_name from destination_number to caller_destination in domain_variables and variables dialplan.

This makes it so calling between two tenants in the same FusionPBX instance doesn't leak internal extension information.


markjcrane:
Why not just get rid of origination_callee_id_name?

jacobbuscher-bt:
If you remove **origination_callee_id_name** then all internal calls between two extensions on the same tenant will display as "Outbound Call".

With origination_callee_id_name set to caller_destination internal calls within a tenant, calls between tenants on the same Fusion instance, and outbound calls display properly.

PR #6784, opened 2023-08-08
Fix time conditions using 'minute-of-day' switching 1 minute early

jacobbuscher-bt:
Fix time conditions using 'minute-of-day' switching 1 minute early.

We are augmenting the value saved based on what select option is selected.

12:00 AM - 11:59 PM will now serialize as 1-1439 as expected 11:59 PM - 11:59 PM will now serialize as 1440-1439 which is unexpected behavior due to freeswitch's 'wrap around' behavior but acceptable.


markjcrane:
Additional information from FreeSWITCH documentation. Important to realize number starts with 1 instead and not 0.
minute-of-day  1-1440  minute of the day (midnight = 1, 1am = 60, noon = 720, etc)

PR #6791, opened 2023-08-17
[4.4] Make intercept feature more fault tolerant

daniel-lucio:
Same as #6788

### DIFF
--- a/resources/install/scripts/intercept.lua
+++ b/resources/install/scripts/intercept.lua
@@ -287,7 +287,7 @@
 			end
 			sql = sql .. ") ";
 			sql = sql .. "AND call_uuid IS NOT NULL ";
-			sql = sql .. "LIMIT 1 ";
+			sql = sql .. "ORDER BY created_epoch DESC LIMIT 1 ";
 			if (debug["sql"]) then
 				log.noticef("SQL: %s; params: %s", sql, json.encode(params));
 			end


markjcrane:
Anyone still using 4.4 is irresponsible and putting their customers at risk for not upgrading. This branch is not maintained for vulnerabilities. I will be removing this branch. To discourage those still using it.

daniel-lucio:
It is more irresponsible to neglect the security patches; instead, you choose to call out your users.

You should be grateful that when someone finds an issue, he/she takes the time to verify if the same patch applies to all the branches.

demonspork:
He's trying to say that 4.4 is an abandoned, unsupported, not recommended version of this product. Every software developer has to draw a line on support and patches for old versions and he has clearly, multiple times, drawn that line in the sand for you specifically.

There is a clear upgrade path to the supported versions and branches with no loss of function.

markjcrane:
When Ubuntu LTS (long term support) is five years old its end of life.  Same for some versions of Debian that get long term support. FusionPBX 4.4 released 5 April 2018 and is 5.5 years old. So it is end of life even by that point of view.

FusionPBX 4.4 is end of life. For those that don't upgrade you are putting your customers at RISK. There are multiple known vulnerabilities in it. Many that will likely be missed by anyone who tries to secure it.  From a different point of view anyone still on 4.4 is missing 5.5 years of development that makes you less competitive than those using the latest release version.

PR #6792, opened 2023-08-18
Add Date to Coverpage

digitalcrisis:
Added date to fax cover page.


digitalcrisis:
Fixed the "TO" field for the fax coversheet with date propagation.

PR #6803, opened 2023-09-14
Fix Ring Group Delay Timing (Enterprise)

emaktech:
The delays for enterprise strategy ring groups have recently been broken such that they are double the actual set values. This returns the delays back to what they should be. See the below commits for all details on this bug:

The enterprise ring group  delay timing was broken in the following commit:  https://github.com/fusionpbx/fusionpbx/commit/557284070a353e0b003368773ac151739bdd2d06

The enterprise ring group follow-me delay timing was broken in the following commit: https://github.com/fusionpbx/fusionpbx/commit/c69535413cba9ee2475f22bb7b344814aa3c7cfd

Here is the original commit where it was fixed with full explanation. Please reference this commit from 2019 for full details: https://github.com/fusionpbx/fusionpbx/commit/f44ed4370cd933820ae7c01bb90b749fa541f884

### DIFF
--- a/app/switch/resources/scripts/app/ring_groups/index.lua
+++ b/app/switch/resources/scripts/app/ring_groups/index.lua
@@ -480,7 +480,7 @@
 					d.destination_number, d.destination_timeout, d.destination_prompt,
 					CASE
 						WHEN r.ring_group_strategy = 'enterprise'
-							THEN d.destination_delay * 1000
+							THEN d.destination_delay * 500
 						WHEN r.ring_group_strategy <> 'enterprise'
 							THEN d.destination_delay
 					END as destination_delay,
@@ -592,7 +592,7 @@
 						if (follow_me_uuid ~= nil and row.is_follow_me_destination ~= "true") then
 							sql = "select d.domain_uuid, d.domain_name, f.follow_me_destination as destination_number, ";
 							if (row.ring_group_strategy == 'enterprise') then
-								sql = sql .. "f.follow_me_delay * 1000 as destination_delay, ";
+								sql = sql .. "f.follow_me_delay * 500 as destination_delay, ";
 							else
 								sql = sql .. "f.follow_me_delay as destination_delay, ";
 							end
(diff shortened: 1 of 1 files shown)


greenbea:
@emaktech Add comments to the Lua code explaining why 500ms is necessary

emaktech:
@greenbea Good idea. Comments added!

greenbea:
@emaktech It is more than you think. When you originate `user/ext@domain`, `switch_ivr_originate`  will call the `user` endpoint. The user endpoint isn't a real endpoint. It is a shortcut to get the real dialstring from the user's XML and calls again `switch_ivr_originate`, so essentially, what happens is `switch_ivr_originate` reads twice `originate_delay_start` variable and waits double the time. 

Your patch fixes the case when the endpoint is the user/ endpoint but will make the delay half the time if called with other endpoints like sofia. 

 The ultimate solution would be freeswitch to handle this in the user endpoint function. I would suggest that fusion manually get the user's dial string from the XML and call originate on it and not rely on the user (shortcut) endpoint for ring group and follow me.

emaktech:
@greenbea Thanks for the more thorough investigation here!

@markjcrane Is anybody willing to fix this as described? Because until then, the highlighted commits are a breaking change, and the values set in the GUI are literally wrong when calling extensions which, unfortunately for us (and probably most people) is the entire point of ring groups...

markjcrane:
I put a thumbs up on @greenbea reply saying I like his response.  You are welcome to call support number. I have time now if you call soon.

markjcrane:
Thanks @greenbea for the important information on this issue.

Here is a commit to fix follow me.
https://github.com/fusionpbx/fusionpbx/commit/6030843ca8c0572126ef886d142b6b5000ded344

Will need one more commit to fix ring groups.

emaktech:
Thanks @greenbea for identifying the actual root cause and @markjcrane for the fast turnaround in fixing this a better way. Closing this PR now :)

greenbea:
I sent a PR to freeswitch to fix this bug. https://github.com/signalwire/freeswitch/pull/2245

PR #6806, opened 2023-09-27
Add ability to set custom logout destination

emaktech:
Allow administrators to set a custom default setting called `logout_destination` which will redirect the user to a destination of their choosing.

We also pass the PHP session ID after logout was completed to allow custom logic to be executed based on the session ID.


markjcrane:
Not sure why you want to share the session ID that is deleted on logout. Please call to discuss.

PR #6812, opened 2023-10-16
illogical condition specified in fax_send

md-riaz:
It doesn't make logical sense for **!empty($route_array)** to also have **count($route_array) == 0**. In a typical scenario, if an array is not empty, its count should be greater than 0.

The condition **!empty($route_array)** checks if the array is not empty, meaning it has at least one element, and therefore, **count($route_array)** should be greater than 0.


markjcrane:

PR #6818, opened 2023-10-26
Update install.php

jrmcclean:
Added addition file paths and updated the code to create an array of possible config file paths and uses a loop to check for their existence. his makes the code more efficient and easier to maintain, especially if you need to add more possible paths in the future.

### DIFF
--- a/app/fax_queue/resources/service/fax_queue.php
+++ b/app/fax_queue/resources/service/fax_queue.php
@@ -104,7 +104,7 @@
 		file_put_contents($pid_file, getmypid());
 	}
 
-//get the email queue settings
+//get the fax queue settings
 	$setting = new settings(["category" => "fax_queue"]);
 
 //set the fax queue interval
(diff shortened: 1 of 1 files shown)


jrmcclean:
Added addition file paths and updated the code to create an array of possible config file paths and uses a loop to check for their existence. his makes the code more efficient and easier to maintain, especially if you need to add more possible paths in the future

markjcrane:
Your change adds ability for the config.conf in the root of the web directory. 

 __DIR__ . '/config.conf',

Problems with this approach.
1. Web server has to be setup to hide config.conf. This really is a big security problem which is why I recommend it not being in the web directory.
2. Most files should not  be place in the root of the web site. It would be better in the resources directory.
3. Need to also support ability to set a custom directory for the config.conf outside of the web directory.
4. Lua also needs to bet setup to support the same config.conf file location.

jrmcclean:
The  __DIR__ . '/config.conf is using the resource directory.  The required.php file is located in the resource folder so this code simply says to search in the same folder as the required.php.  I made several test with this and it is as prescribe.

PR #6821, opened 2023-11-01
New Devices  permissions  profile select and edit.

volga629-1:
New device select permissions  allow greater flexibility  for limited  user groups have access to devices profile in view only mode.

Screenshot 2023-11-01 at 3 28 55 PM

### DIFF
--- a/app/devices/app_config.php
+++ b/app/devices/app_config.php
@@ -308,7 +308,7 @@
                 $apps[$x]['permissions'][$y]['name'] = 'device_profile_select_edit';
                 $apps[$x]['permissions'][$y]['groups'][] = 'superadmin';
                 $apps[$x]['permissions'][$y]['groups'][] = "admin";
-		
+		$y++;
 		$apps[$x]['permissions'][$y]['name'] = 'device_profile_key_view';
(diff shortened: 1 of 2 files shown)


volga629-1:
Can you please reopen  PR 
We tested and permissions are not working 


Example device_profile_edit is use in multiply places  if you enable it will show the profile button and pencil.

Where my commit had separate permissions  for specific  select button ( pencil ) and preview of profile in view only mode. 


device_profile_edit  is use in multiply places and it interfere  between different functions . 


```
                        if (permission_exists('device_profile_edit')) {
                                echo "                  <button type='button' class='btn btn-default list_control_icon' id='device_profile_edit' onclick=\"if($('#device_profile_uuid').val() != '') window.location='device_profile_edit.php?id='+$('#device_profile_uuid').val();\"><span class='fas fa-pencil-alt'></span></button>";
                                echo "                  <button type='button' class='btn btn-default list_control_icon' onclick=\"window.location='device_profile_edit.php'\"><span class='fas fa-plus'></span></button>";
                        }
                        echo "                  <br>".$text['description-profile2']."\n";
                        echo "          </td>";
                        echo "  </tr>";
                }
```

PR #6825, opened 2023-11-03
Fix events class

frytimo:
What started as a small fix turned in to a rewrite. Created an interface to use with the events.


markjcrane:
I think this should be called event_socket to indicate more what it is for and in the app/switch/resources/classes directory. I would like to save events class name for Fusion Events.

PR #6827, opened 2023-11-03
add echo_canceller and echo_suppressor variables

denisent:


markjcrane:
The variables should have a grandstream_ prefix.

denisent:
Probably true. However, just following exactly what is setup for ht802

markjcrane:
Probably wrong as compared to the majority of the templates.

denisent:
added grandstream_ prefix

PR #6828, opened 2023-11-07
Email class improvements

frytimo:


markjcrane:
Your tab added to every line hid your changes for the file comparison. So I changed it so your changes would be more visible.

PR #6841, opened 2023-12-05
Prevent duplicate user allocation

AndyX90:
Prevent duplicate user allocation caused by https://github.com/fusionpbx/fusionpbx/pull/6837
Only assign the current user, if he is not already assigned.


AndyX90:
@markjcrane  This commit is for the (edge?)case where the editing user has no permission to view contact-permissions. Without this, there would be a user-add on every edit which leads to contacts with multiple (same) users assigned.

PR #6847, opened 2023-12-11
Snom template changes for SmartLabel and Intercom policy

mgajjar19:
Changed the template for Snom phones to set intercom policy for Paging which is by default set to always but with this change we can set a variable to determine its value.

Also for the Smart label, by default on Snom phones, it goes to short labels which don't show names for speed dials so this allows to set a variable to specify the length of the labels for Snom phones.


mgajjar19:
Updated Snom templates for Intercom policy and Smart Label.

markjcrane:
This pull request shows no changes to any files.

mgajjar19:
It was a mistake on my end. I think I already did a pull request for that earlier. Sorry for the trouble.

PR #6867, opened 2024-01-05
Updated Snom D735 Template

mgajjar19:
Snom D735 has a known issue where the lights on the phone won't illuminate when in use. It still functions correctly but without any lights. I talked with Snom support and they provided me with the string that needs to be changed in the template.

This issue started from Snom firmware version 10.1.127.10

Release Notes - https://service.snom.com/display/wiki/10.1.127.10+Release


tony1661:
Any updates on getting this merged?

PR #6869, opened 2024-01-13
Added CDR dropdown for showing CC Queue calls

tony1661:
This PR is to add a dropdown to the CDR filter options. It gives the user a way to filter for only call center calls.

image


markjcrane:
Accepting your pull request but will change the permission names to include the term search_ like other permissions for the search field in XML CDR.

PR #6870, opened 2024-01-16
Force close port

frytimo:
Create function to force a port to be closed. This requires the `--enable-debug` flag to be used when compiling freeswitch, sofia-sip, and spandsp. This also requires gdb, pidof, netstat, and lsof to be installed on the system. This will not work on a Windows system. Tested on a raspberrypi. Tested system only required lsof to be installed manually with `apt install -y lsof` as the other tools were preinstalled.

### DIFF
--- a/app/sip_status/cmd.php
+++ b/app/sip_status/cmd.php
@@ -50,6 +50,26 @@
 	$profile_name = $database->select($sql, $parameters, 'column');
 	unset($sql, $parameters);
 
+//get the port from sip profile name
+	$sql = "select sip_profile_setting_value from v_sip_profile_settings ";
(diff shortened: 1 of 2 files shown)


frytimo:
Also, I had forgotten to mention that gdb needs to be set to not require confirmation on close. To do this paste the following in the file called '**/etc/gdb/gdbinit**':
```
define hook-quit
        set confirm off
end
```
Without this setting, there is a chance that PHP will hang when trying to close ports because gdb expects an interactive shell.

demonspork:
Direct way to do this without configuration of gdb I think is the server prefix: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Server-Prefix.html

PR #6871, opened 2024-01-16
Add suggested extension ranges

chansizzle:
As a method to adhere to corporate numbering standards this PR adds suggested extension ranges for the following types of "extensions":

- call center queues
- call flows
- conference centers
- conferences (simple conferences)
- extensions
- fax servers
- IVRs
- ring groups
- time conditions

To use, simply adjust the range values and enable the variables. These are suggestions rather than forcing the "extensions" to conform to the ranges created.

The suggestions are set as placeholder text in the add & edit extension number edit fields and adds example text below the description for the field. The placeholder text displays if the field is blank.

image

Each of the 9 ranges are configured with your desired text for your corporate numbering standards via the new Ranges category in Default settings. The default settings variables are disabled by default so those new to fusionpbx do not assume the suggested ranges are required ranges.

As the default settings are text you can set the ranges to be any text you wish. i.e. "100-199,100x", "*30,*31,*88", "something else". Keep in mind that the placeholder text will get cut off if you go beyond the width of the field/text box.

image

The PR also sets these required extension number fields as required if they were not already set as required.


paulc97:
That would be a huge improvement. So that customer admins not mix the numbers and get confused after. Happens all the time

jrmcclean:
This is an amazing idea... however i believe it needs to be polished. 1) The range should act somewhat like the theme colours. If the settings is false it should default like the present code today.  Doing so will ensure  you code is backwards compatible and does not break someones system who is missing this config.  2) There is no validation. This should be added to the code to ensure it adhere to the respective range/ sub category. The point is to ensure that the admin person that is adding this new Ranges( extension, queue range.. etc.. ) adhere to the code values.  3) As a suggestion you can add code for super admin to bypass the validation.  That is my 2 cents :)

n0obHere:
Very simple but neat idea. I'm always struggling to remember and keep extension ranges consistent across multiple applications. This will be a big help. Any chance this branch will be merged into master any time soon?!?!

tony1661:
I love this! Will this get merged? @markjcrane

tony1661:
Yay it's merged!! Looking forward to trying this out

PR #6872, opened 2024-01-17
make time_condition_edit.php use time_format var for select options

paulc97:
Use the existing default Variable time_format to adjust the options of the selects of "hours of day" and "time of day" accordingly.

We have been operating FusionPBX for our customers in Germany for several years. Over time, we have noticed many small things that the main user base probably does not notice. The AM and PM times are very unusual in Germany. Therefore I hope that this change will be accepted. There is already a default variable that lets you choose the time format. In my opinion, a user who selects 24 hours and activates the variable should no longer find AM PM times on the interface. If you notice this elsewhere, I would also try to fix it.

### DIFF
--- a/app/time_conditions/time_condition_edit.php
+++ b/app/time_conditions/time_condition_edit.php
@@ -821,24 +821,48 @@
 						break;
 
 					case 'hour': //hours of day
-						for (h = 0; h <= 23; h++) {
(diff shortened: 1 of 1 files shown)


markjcrane:
You are right I had not noticed this and agree that it should use the time_format variable. So I'm accepting your pull request. Thanks for sharing it.

PR #6902, opened 2024-02-22
Fix unprotected child menu items

frytimo:
If parent menu item was protected the child items were still deleted. This pull will ensure that each child of the parent inherits the parent menu item protection.


frytimo:
Seems the SQL missed the individual item still. More work required.

PR #6907, opened 2024-02-28
disable web browser app.unavalible

cranben:


markjcrane:
The foreach needs to wrap around the entire line as done in other templates.

PR #6909, opened 2024-03-01
Update app_languages.php

tony1661:
added the official descriptions from Freeswitch's documentation for a better understanding of the parameters


markjcrane:
The problem with adding the official description is plagiarizing. In which case you need their permission. It better to read it understand it and then write your own description.

PR #6915, opened 2024-03-12
Option to Select Drop-Down in Ring Groups

frytimo:
###Select Drop-Down in Ring Groups***
Added a feature for the destinations class to filter on only applications specified for destination. This allows for all destinations to have the option to use the new dropdown select options mode. Currently this feature is implemented in ring groups but more areas could be added. 

To use this feature a default value in the **Default Settings** page under the category of **Destinations** with the name of **ring_group_select_mode** must be set to true. 

A bug was found and fixed along with other small enhancements to the destinations class and documentation was also added.

###Bug fix###
A small bug was found when selecting a voicemail destination. The destination would lose the context resulting in a display bug.

###Enhancements###

- Added constructor array parameter called _$params_ to allow for database and settings objects to be passed in
- domain_uuid can be specified. If no domain_uuid is specified the $_SESSION global is used
- user_uuid can be specified. If no user_uuid is specified the $_SESSION global is used
- **database** is now specified in the constuctor. If no database object was supplied then a new database object is created
- The new **settings** class is now used and is created in the constructor so database is not queried more than once


n0obHere:
Hello @frytimo @markjcrane, just curious about this PR, why was it closed without merge? I think it was a really good to have functionality having a destination dropdown in ring groups.

frytimo:
This was closed as it was over-engineered for the problem at hand.

PR #6917, opened 2024-03-19
initial commit of xml_cdr_service

frytimo:


frytimo:
This has been moved in to a new design removing the need for this one.

PR #6926, opened 2024-03-25
Remove dependencies on defined constants. Single line include

frytimo:
It is very important to remove the dependencies on any defined project variables. This removes those dependencies without breaking backwards compatibility as it does not remove the class definition but instead, instantiates the loader object immediately at the end of the file. This will not break any other file that still creates a new auto_loader object as PHP permits more than one auto loader registration.

The following single line can now be used:

```
require_once 'resources/classes/auto_loader.php;
```

### DIFF
--- a/resources/classes/auto_loader.php
+++ b/resources/classes/auto_loader.php
@@ -88,6 +88,6 @@ class auto_loader {
 	}
 }
 
-new autoloader();
+new auto_loader();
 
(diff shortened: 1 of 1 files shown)


frytimo (commented):
fix typo

frytimo (commented):
remove is_array check as it is no longer needed

PR #6947, opened 2024-04-11
[Feature] Database Replication Check

k-flagman:
If u have Reserve Server for your system - you have to check the synchronize two schema(database);
You can always look what is the status of comparing;

Display the status of database comparing;
1) Set the settings in default settings list "Database Replication Check":

> `node_1`: sub1.domain.com
> `node_1_password`:  Node 1 Domain Access Password (**64 length**)
> `node_2`:  sub2.domain.com
> `node_2_password:`  Node 2 Domain Access Password (**64 length**)

**The Reserve Server must have same settings;** 
You can generate the passwords with:
`openssl rand -base64 46`

========================================================================

The First Version has the popup when ones click error status;
Screenshot from 2024-04-11 11-56-14
Screenshot from 2024-04-11 11-57-18
Screenshot from 2024-04-11 11-57-20
Screenshot from 2024-04-11 12-08-02

========================================================================

Next steps:

- [ ] Thinking about displaying it on the dashboard

### DIFF
--- a/app/check_compare_db/app_config.php
+++ b/app/check_compare_db/app_config.php
@@ -1,0 +1,78 @@
+<?php
+//application details
+	$apps[$x]['name'] = "CheckCompareFor2Db";
+	$apps[$x]['uuid'] = "7b7b0be6-919c-4ab3-a1c7-beafbe40cb3d";
+	$apps[$x]['category'] = "";
+	$apps[$x]['subcategory'] = "";
+	$apps[$x]['version'] = "1.0";
+	$apps[$x]['license'] = "Mozilla Public License 1.1";
+	$apps[$x]['url'] = "http://www.fusionpbx.com";
+	$apps[$x]['description']['en-us'] = "";
+	$apps[$x]['description']['en-gb'] = "";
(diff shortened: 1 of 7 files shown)


demonspork:
This is a good idea, but is a partial implementation of what would make the most sense.

Schema is already handled by the schema defaults, so why not just execute the schema defaults check against a list of db servers instead of getting into the weeds with comparing it yourself. This approach would also make it useful for those using Logical Replication to handle schema updates across the cluster, instead of limiting the approach to only 2 servers.

demonspork:
Your solution is a very narrowly scoped solution with a very specific non standard use case. The way you have built it here partially puts it as a optional app and partially puts it into the core. 

I'm not a maintainer, but I'm willing to bet that dichotomy will put some pause on this being accepted as is. we'll wait for Mark to weigh in but I would recommend trying to have this app include all of its components in its own app folder to make it easy to contribute as a dedicated repo as an optional app instead of being combined into the core. 
see other independent repositories like the fusionpbx-app-messages repo.

PR #6958, opened 2024-04-24
Snom firmware fix

tony1661:


markjcrane:
The path for this seems odd why are is 10.1.175.10 included?

tony1661:
hi @markjcrane I am not sure. A little while back snom changed where they host their firmware files. They can now be found here:
[https://service.snom.com/display/wiki/10.1.175.10+Release](https://service.snom.com/display/wiki/10.1.175.10+Release)

Here is the format the use for the files:
xxxxx = firmware version
yyyyy = phone model number

https://downloads.snom.com/fw/xxxxx/bin/snomyyyyy-xxxxx-SIP-r.bin

The new D800 series phones do not use a bin file. Instead they use a .swu file.

markjcrane:
There is another variable called snom_firmware_url. The way its built is to add the HTTP protocol, domain name and path in that variable. The only the file name goes in the snom_firmware_<model number>  value. This follows how other vendors are setup for provisioning.

tony1661:
The issue is the path is different for each firmware so you'd need to update all phone models at the same time

tony1661:
To clarify, this is how things are setup for a D785 model of phone:

$snom_firmware_url = "https://downloads.snom.com:443/fw/"
$snom_firmware_d785 = "**10.1.152.12**/bin/snomD785-**10.1.152.12**-SIP-r.bin"

Notice how the model specific variable has the version number in the path as well as in the actual  `.bin` file.

tony1661:
Hi Fusion team,

Let me know if there is anything I can clarify

PR #6974, opened 2024-05-08
[gateways] Add domain_name, domain_uuid variables to gateways

greenbea:
When a gateway belongs to a domain and gets transferred directly to the the domain context, we want to set the domain_name and domain_uuid as channel variables for the dialplan.


markjcrane:
This seems to assume that all domains have a corresponding gateway(s). Many people have a few gateways that are used globally.

PR #6989, opened 2024-05-31
Add methods to handle device and add documentation

frytimo:
device_profile_settings and device_settings are never looked up in the database. This adds those methods with a more complete set of documentation on how the settings class works using hierarchical overriding.

### DIFF
--- a/resources/classes/settings.php
+++ b/resources/classes/settings.php
@@ -1,84 +1,147 @@
 <?php
 
 /**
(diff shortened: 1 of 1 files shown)


demonspork:
Loving all this work on a settings class! Any custom code I write will be greatly benefited by not having to include as much code to manage and manipulate various default/device/domain settings that it interacts with!

PR #6995, opened 2024-06-04
add event socket settings to config file when using installer

frytimo:
add switch.event_socket.host, switch.event_socket.port, and switch.event_socket.password to config.conf when using the core/install/install.php

### DIFF
--- a/core/install/resources/classes/install.php
+++ b/core/install/resources/classes/install.php
@@ -83,6 +83,11 @@ if (!class_exists('install')) {
 				break;
 			}
 
+			//common settings for all operating systems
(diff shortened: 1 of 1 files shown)


markjcrane:
Considering move this into default settings.

markjcrane:
I'm considering adding this to the default settings category event_socket.

frytimo:
I had originally considered that but thought creating a dependency on the database for only those settings is a lot of overhead. There could be a use case for a service that connects to the event socket without the need for the database. Such a use case might have to do with the call center module or another module. The other consideration was that the default settings allows per domain whereas this setting should not have that option.

PR #7060, opened 2024-07-22
remove splitting the process as the systemctl will do that

frytimo:
also add the ability to echo to screen when running in debug mode


demonspork:
Well this have an impact on non-systemd environments like FreeBSD?

Just a thought when i saw this, i personally don't know the answer.

PR #7070, opened 2024-07-26
generate uuid changes

tony1661:
This PR is to make the generate uuid button in the extension config default for super admins.

Based on the IRC chat, it seems many do not know about this feature.

I've also changed the language file to say "Generate UUID" instead of "Generate" to make it more obvious for new users of the feature


markjcrane:
You may also realize you can click on the arrow to make it an editable field and type uuid then press save and it will generate a UUID.

tony1661:
I do indeed know that I can type uuid in the edit field but that is only because you showed me once. Others in IRC had no idea this feature existed.

I feel like "Generate UUID" explains well to users who don't know what this is.

Adding this as a default settings will make it get more use and thus be more secure.

Up until I knew about this feature, I would create fake MAC addresses that, looking back, were very insecure

PR #7077, opened 2024-07-30
set fax failed when missing

tony1661:
This PR sets the `$fax_status = 'failed'` when the file cannot be found. Not doing this causes faxes with missing files to never get taken out of the fax queue.


tony1661:
@frytimo here is the PR we spoke about. Let me know if there should be any modifications.

PR #7082, opened 2024-08-02
Disable attaching fax to email confirmation for a domain. Under Fax c…

nbhanji:
…ategory, disable_attachment subcategory must be enabled and true


nbhanji:
This allows for faxes not to be attached to the email sent for confirmation. Our clients have request not to have fax in the email that might contain PHI information.

PR #7094, opened 2024-08-13
Yealink Wifi Settings variables

mgajjar19:
These variables are used to set the Wifi credentials and settings on the below-listed Yealink models.

Only T54S/T52S/T48G/T48S/T46G/T46S/T42S/T41S/T29G/T27G Models support these parameters.

The variables were already configured for a couple of models, I just added them for all supported models.


markjcrane:
Looks great... Your pull request has been accepted.

PR #7107, opened 2024-08-22
Localization: Add empty Georgian strings

NorwayFun:
Hello,
This is initial PR for supporting Georgian language in FreePBX.
This patch is generated using AI generated python script (not a dev myself), so it can contain mistakes.
AFAIK, there also has to be some DB changes, but it's really out of my scope.
In case this will be merged, translations will follow.

BR, Temuri


NorwayFun:
Will reopen it once I'll get files translated.

PR #7111, opened 2024-08-29
Localization: Finishing remaining files for Georgian translation

NorwayFun:
Hello,
Here's files, missing from previous PR.
Temuri


markjcrane:
I also updated user_settings.php to put ka-ge after it-it. I don't want to do it again. Please update your version then re-apply your changes to user_settings.php

PR #7152, opened 2024-10-05
Update dp750/mac.xml

ijaron:
- Use $grandstream_ring_timeout for Ring time rather then hardcode 
- Use $grandstream_dial_plan for every account rather then only first set Station name to DP750-{$mac}
- Add option for Automatic Reboot (Firmware 1.0.17.1+) Allow $grandstream_stun_server and use stun if $grandstream_nat_traversal == 1 
- $grandstream_accept_sip_from_proxy_only to set Disable Direct IP Call 
- Turn off 3CX Auto Provision
- Use $grandstream_automatic_provisioning and convert to values used in DP750 vs other grandstream hardware 
- Allow $grandstream_automatic_provisioning_minutes $grandstream_automatic_provisioning_hour and $grandstream_automatic_provisioning_day 
- use $grandstream_phonebook_download_interval rather then just disabling

### DIFF
--- a/resources/templates/provision/grandstream/dp750/{$mac}.xml
+++ b/resources/templates/provision/grandstream/dp750/{$mac}.xml
@@ -3428,7 +3428,7 @@
 {/if}
 
 <!-- Phonebook Download Interval (in minutes) -->
-<!-- Valid value range is 5-720. Default is 0 for disabled-->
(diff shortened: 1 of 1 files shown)


markjcrane:
Looks great accepting your pull request. Sorry for the delay

PR #7156, opened 2024-10-08
Fix CDR Recording Playback - Update xml_cdr.php

anthony-ricci:
Fixed CDR recording playback - casting an empty string to an int is breaking the if statement. This cast isn't required anyways because is_numeric will already find whether a variable is a number or a numeric string 

See https://php.net/manual/en/function.is-numeric.php


markjcrane:
I found a client system where is_numeric treated a number in a string as not a numeric. The client's system was PHP 7.1, which was the reason for the change.

PR #7172, opened 2024-10-27
Update voicemail script to add call_uuid with voicemail messages

md-riaz:
Hi,
I saw that there was no way for connecting voicemail messages to a call, so i added a call_uuid in voicemail_messages table.

Can anyone help me how to add the extra column in this pull request?


markjcrane:
You added the column to the database. Are you asking about adding it to the voicemail messages list in the web interface?

md-riaz:
actually there is no link between voicemail and call history in the web interface, no one could tell which call got which voicemail without looking at caller name or number.

PR #7176, opened 2024-11-01
Changed z-index for high resolutions

AdSecIT:
Video background was not showing on the login page, changing this setting to the z-index has fixed the issue in my situation.


markjcrane:
You found a solution but you didn't understand there is benefit for the design. The background_color_enabled = false if you don't want the background to cover the image. 

Why is the color on top of the video because it allows it to act like a filter when the color is semi-transparent.

PR #7184, opened 2024-11-27
add left and right padding to the dashboard applets

chansizzle:
This helps when applying themes. The header of the applets need to be padded.

image


markjcrane:
This has been fixed. Instead of padding the title background should expand 100% the width of the widget.

PR #7189, opened 2024-12-02
Create 465_transfer_to_me.xml

denisent:
The purpose of this dialplan is to allow the user to transfer their existing call back to themselves so they can answer the call with a different device. Transfer from desk phone to software phone to home office phone, etc.

This dialplan has been designed to be global and there will be no 'parking lot' collisions as each extension across all domains will have their own uniquely identified parking lot/park ID if the form of "<extension>-<domain_name>". Example: 101-customer45.pbxdomain.com, where extension = 101 and the domain = customer45.pbxdomain.com.

Following the example above, extension 101 will Transfer or Blind Transfer a call to *63 (which spells *me). Their extension 101 will hangup and display something to the effect of "transfer complete" (phone make & model dependent). After 5 seconds all of the user's extension 101 devices will ring. The user answers the call with their desired registered extension 101.

The park_lot variable will display as the example above in the CDR details.

### DIFF
--- a/app/dialplans/resources/switch/conf/dialplan/465_transfer_to_me.xml
+++ b/app/dialplans/resources/switch/conf/dialplan/465_transfer_to_me.xml
@@ -1,0 +1,24 @@
+<extension name="transfer_to_me" context="global" continue="false" uuid="ccf52820-f471-406b-83c9-cbbbe597513e" global="true" enabled="true" order=465>
+	<condition field="destination_number" expression="^\*63$" break="never"/>
+	<condition field="${sip_h_Referred-By}" expression="sip:(.*)@.*" break="never">
+		<action application="set" data="referred_by_user=$1" inline="true"/>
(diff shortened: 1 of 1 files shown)


tony1661:
@markjcrane would this be more useful as a bind_digit_action so that mid call, we can dial *63 to transfer the call?

I really like this idea

denisent:
@tony1661 If you program a key on your phone you can transfer with 1 button press. BLF type, Speed Dial type, Transfer type, or others.

tony1661:
> @tony1661 If you program a key on your phone you can transfer with 1 button press. BLF type, Speed Dial type, Transfer type, or others.

True, however we can already do this currently by transferring to our own extension.

This PR doesn't add something that wasn't there, it just changes it I guess.

I do like the idea though but maybe the *63 mid call would be good?🙂

Thoughts?

PR #7193, opened 2024-12-06
New Feature: SMS/MMS to Email

anthony-ricci:
This PR adds the ability to have SMS and MMS notifications via email. It uses Email Templates to control what is populated in the email. Email notifications are sent to the "User" configured in the Destination.


markjcrane:
The app/messages and app/providers should be removed from this pull request.

PR #7202, opened 2024-12-22
Revert "Update voicemail script to add call_uuid with voicemail messages"

md-riaz:
Reverts fusionpbx/fusionpbx#7172

I saw that the main voicemail_messages table primary uuid is the same as the xml_cdr table primary uuid, so there is no need extra call_uuid column as the voicemail_messages table primary uuid is a direct match to the call uuid


markjcrane:
You are right I forgot about that made it many years ago. Since it was the last commit I can remove it without another commit.

PR #7205, opened 2024-12-29
Fix regex to match country code OR leading zeroes.

realJoshByrnes:
Currently the regex when a country code and phone number are provided matches invalid sequences, and doesn't match numbers where leading zeroes are provided.

eg. Consider the Australian Number (02) 5550 1234

A VoIP provider may send one of the following three phone number formats, which are all valid:
* \+ 61 2 5550 1234
* 61 2 5550 1234
* 02 5550 1234

Currently, the default regular expression that FusionPBX provides would allow you to match one (format with 61) or the other (format with leading 0). Users have stated on the forums that they have created both entries for each number.

The default regular expression would also match +255501234 (a Tanzania phone number in this case, but could be anywhere in the world depending on the number).

The new regex ensures that the optional plus is only present before the country code, and also allows the matching of a preceding zero.

Caveat: The new regex, while not matching arbitrary numbers like the old one, does match where a leading zero should not be used (eg. North America).

### DIFF
--- a/app/destinations/destination_edit.php
+++ b/app/destinations/destination_edit.php
@@ -394,16 +394,16 @@
 						foreach($destination_numbers as $destination_number) {
 
 							//convert the number to a regular expression
-								if (isset($destination_prefix) && !empty($destination_prefix)) {
+								if (isset($destination_prefix) && strlen($destination_prefix) > 0) {
 									$destination_numbers['destination_prefix'] = $destination_prefix;
 								}
-								if (isset($destination_trunk_prefix) && !empty($destination_trunk_prefix)) {
+								if (isset($destination_trunk_prefix) && strlen($destination_trunk_prefix) > 0) {
 									$destination_numbers['destination_trunk_prefix'] = $destination_trunk_prefix;
 								}
-								if (isset($destination_area_code) && !empty($destination_area_code)) {
+								if (isset($destination_area_code) && strlen($destination_area_code) > 0) {
 									$destination_numbers['destination_area_code'] = $destination_area_code;
 								}
-								if (isset($destination_number) && !empty($destination_number)) {
+								if (isset($destination_number) && strlen($destination_number) > 0) {
 									$destination_numbers['destination_number'] = $destination_number;
 								}
 								$destination_number_regex = $destination->to_regex($destination_numbers);
--- a/app/destinations/resources/classes/destinations.php
+++ b/app/destinations/resources/classes/destinations.php
@@ -22,6 +22,7 @@
 
 	Contributor(s):
 	Mark J Crane <markjcrane@fusionpbx.com>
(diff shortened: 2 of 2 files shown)


realJoshByrnes:
I should note, that this patch is specifically for the destinations app.

markjcrane:
The leading zero is likely the trunk prefix. Were you aware there is a field in destinations for the trunk prefix? To see it, you should assign the trunk_prefix permission to the superadmin, and admin groups, along with other equivalent custom groups.

Also to accept your change please sign up as a contributor here.

realJoshByrnes:
@markjcrane I wasn't aware that there was a trunk prefix option (not sure why it would be hidden by default, since most countries use one), and that should definitely be set correctly (to `0` in my case).

Unfortunately, even with the trunk prefix set correctly to `0`, the resulting regular expression remains unchanged: `^\+?(?:61)?(255501234)$`.

* The current commit is still an improvement on matching (as it is)
* Now that I know that trunk prefix exists, I can improve the matching here. I'll add another commit once I've done some more testing

Also, CLA is complete.

realJoshByrnes:
# Current

There's a few things that seem to have been overlooked, here's some examples.

'N', 'X' and/or 'Z' can only currently be used in the phone number, and only when the area code and trunk prefix

* Perfect
  * Only number set
  ````regex
  ^(55501234)$
  ````
* Missing start of line match (^) in international format (could start with anything)
  * Country code, trunk prefix, area code and number
  ````regex
  (\+?61255501234$|^0255501234$|^255501234$|^55501234$)
  ````
  * Country code, trunk prefix and number
  ````regex
  (\+?6155501234$|^055501234$|^55501234$)
  ````
  * Country code, area code and number
  ````regex
  (\+?61255501234$|^255501234$|^55501234$)
  ````
* Too lax, could match another country (eg. +55501234)
  * Country code and number
  ````regex
  ^\+?(?:61)?(55501234)$
  ````
* Doesn't match at all (only matches number)
  * Area code and number
  * Trunk prefix and number
  * Area code, trunk prefix and number
* No number present
  `PHP Warning:  Undefined variable $destination_regex in...

markjcrane:
I'm not interested in an almost complete rewrite. It's very risky because the changes are more dramatic. Please just make the adjustments required.  Indentation changes should be their own commit because these make it look like everything changed.

- Replacing the empty() function is a good idea because it treats 0 as empty.
- Adding the ^ makes it more exact

realJoshByrnes:
I only re-wrote one function (that had a bunch of issues as pointed out above). The code is simplified, and more accurate than the original.
I added one function, which supplemented the previous function.
Your choice to use it (or not). I made a contribution based on where I had seen issues, which took some effort not being too familiar with the codebase.

Not sure about the indentation changes, It's not showing to me on on VS Code diff or GitHub diff, happy to fix that, but I can only see it showing actual code changes.

PR #7208, opened 2025-01-07
Fix: Changed 'http_enabled' comparison from string 'false' to boolean…

chrisswy:
Fix: Corrected 'http_enabled' comparison in xml_cdr class
The condition was always true due to comparing a boolean setting with the string "false". This change uses the boolean false for the comparison.


markjcrane:
One more try to see if you want to sign up as a contributor.

markjcrane:
Closing this pull request as the person that submitted didn't sign up as a contributor.

PR #7209, opened 2025-01-09
fix xml_cdr import failing to move a zero byte record to failed folder

frytimo:
When an XML CDR record is zero bytes that is recorded in the /var/log/freeswitch/xml_cdr directory, the xml_cdr class would fail to move the file. This causes the record files to eventually build up to where the files can exceed the import limit. This adjustment moves the sanity checking for `filesize` to before the import attempt and checks for over limit and zero bytes. If those conditions match the file is moved to the "failed" folder.


frytimo:
Looks like something went wrong when trying to import the patch from testing machine. Closing pull request to have another one that will have the updated file before the pull request is submitted.

PR #7211, opened 2025-01-10
add xml statistics mouse over text and use languages variables

chansizzle:


chansizzle:
One of the mouse hovers
Screenshot from 2025-01-09 17-42-01

markjcrane:
When the translations are not English leave them an empty value if you are not going to fill them in.

chansizzle (commented):
updated the languages file to remove English language from non-English as directed

PR #7214, opened 2025-01-17
Make extension concurrent outbound limit a default setting

denisent:


markjcrane:
We are using a setting under category: extension called limit_max. Different name but does the same thing.
- https://github.com/fusionpbx/fusionpbx/commit/e95ea23ce94ca86ef74ef45f9e862d32c9dcdfea

PR #7221, opened 2025-01-18
Add a voicemail deletion queue

eliweaver732:
## Purpose
Instead of immediately deleting a voicemail simply mark the status of the message as _deleted_. Users can listen to "deleted" messages and undelete/forward them etc. 
There is a script that can be run from a cronjob that deletes any messages deleted longer ago than the new default setting, **deletion_queue_retention_hours**,  which I added to the voicemail app. 
This behaviour is helpful to prevent accidental deletions of a voicemail.

## Changes
1. Added default settings **use_deletion_queue** and **deletion_queue_retention_hours** to the **voicemail** category. 
2. Added a **deleted** phrase for **voicemail_listen_file_options**.
3. Added option 3 in the main menu to listen to deleted messages.
4. Added ability for **menu_messages** to accept and query messages by any status.
5. Added ability to pass any status to **message_saved** to set the message status to anything.
6. Added a message count phrase for deleted messages in the main menu.
7. **message_saved** updates field **update_date** to record when a message was saved or "deleted".
8. Added script **delete_messages** which deletes messages if the interval between **update_date** and now is greater than **deletion_queue_retention_hours**.

## Usage Notes
**delete_messages** needs to be run from a cronjob. Use this command: _fs_cli -x "luarun app/voicemail/resources/scripts/delete_messages.lua"_ I run this cronjob hourly. 

I could also make this run everytime a user logs into his mailbox and only on that mailbox if there is interest.

### DIFF
--- a/app/switch/resources/conf/languages/en/vm/sounds.xml
+++ b/app/switch/resources/conf/languages/en/vm/sounds.xml
@@ -104,6 +104,25 @@
     </input>
   </macro>
 
+  <macro name="voicemail_deleted_message_count">
+    <input pattern="^(1):(.*)$" break_on_match="true">
+      <match>
+        <action function="play-file" data="voicemail/vm-you_have.wav"/>
+        <action function="say" data="$1" method="pronounced" type="items"/>
+        <action function="play-file" data="voicemail/vm-$2.wav"/>
+        <action function="play-file" data="voicemail/vm-message.wav"/>
+      </match>
+    </input>
+    <input pattern="^(\d+):(.*)$">
+      <match>
+        <action function="play-file" data="voicemail/vm-you_have.wav"/>
+        <action function="say" data="$1" method="pronounced" type="items"/>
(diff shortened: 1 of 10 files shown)


markjcrane:
I like the use of message_status = deleted as the way to mark deleted voicemails

nktech1135:
I've wanted something like this for a long time. This could work as is, but i do have one thought on the deletion script, Wouldn't that fit better in the fusionpbx-maintenance script than in it's own script. That way it would execute with the rest of the maintenance tasks.

nktech1135:
There is merit to making this a per domain or per user setting, but triggering it on log in doesn't really feel like the right place to do it. If we do make it run based on voicemail authorization status i'd rather trigger it on hangup or log out status.
Is that a possibility?

markjcrane:
This is a good improvement. Accepted it today and now working on adding the phrases changes to the other languages.

PR #7223, opened 2025-01-20
multiple changes to the and for the yealink t58W,and images for yealinks in general.

rabbidiesel:
Added the variable for Bluetooth so that it won't be disabled every time the phone is provisioned. Added what is needed for the wallpaper and firmware.


markjcrane:
I updated the path to use /app/yealink/resources/images/ for the wallpaper directory.

PR #7230, opened 2025-01-24
use lowercase matching domain name

frytimo:


frytimo:
When the domain name has upper and lower case letters. The domain UUID results in not found in app/provision/index.php

PR #7239, opened 2025-02-03
Create sweep function for event guard

frytimo:
I have added a new feature to allow the logs to be "swept" clean. The basic idea is to remove duplicate IP addresses from the logs. This ignores whether or not the IP address was blocked from a failed authorization request or an attempt at registering to the IP address of the server. The work is done using the database in an SQL query. I have not been able to test performance on a large dataset but I would be happy to if anyone is interested in sharing their own event_guard table data.
This can be a heavy work for the database server if it has a lot of records so use with caution.
If the browser shows a timeout error for large datasets, the query is still running but the browser is no longer waiting.


markjcrane:
Reason this hasn't been accepted is that I would rathe use app defaults to clear the duplicates. And ultimately even better prevent the duplicates in the first place.

PR #7257, opened 2025-02-14
Vm deleted queue

eliweaver732:
remove_deleted_messages function had a bug that caused infinite loops if there was a message to delete. It was triggered on login to the voicemail which caused the call to hang and required a restart of freeswitch to remedy.


markjcrane:
It appears this pull request is missing files.

PR #7278, opened 2025-02-26
add call status to emergency logs page

chansizzle:
I found it was important to know if an emergency call was successful. This pull request adds the status of the call to the emergency logs page so we can know if the call was successful or not.

Firefox_Screenshot_2025-02-26T01-15-52 158Z


markjcrane:
This will work, but it doesn't account for the possibility of the CDR record being deleted before the emergency logs.

chansizzle:
The left join ensures the record displays with the call status as blank if the CDR has been deleted. 

image


It could be changed to "if call status is null then "cdr deleted" or a similar message.

PR #7279, opened 2025-02-27
Leave vm as new

rudyweiler:
Add the ability to press 9 to skip to the next message without changing the message state, allowing us to keep a message in new status, and then the forward change is to move the forward to email option from opt 9 to a submenu of opt 8.


markjcrane:
The description for this pull request was copied from an IRC conversation.

"Add the ability to press 9 to skip to the next message without changing the message state, allowing us to keep a message in new status, and then the forward change is to move the forward to email option from opt 9 to a submenu of opt 8."

rudyweiler:
Hello, Checking in to see what the status of this pull request it. Thanks

markjcrane:
Main main hesitation on accepting this is that the numbers were changed. Affecting current behavior rather than only adding to it.

rudyweiler:
Thanks for getting back to me. I understand your concern. Would there a better key available for us to implement the "leave as new" feature?

PR #7280, opened 2025-02-27
fix email category misspelling

chansizzle:
Sending an email with the email templates will never work if the email template category is misspelled.

Search this URL for 'emergency' to confirm singular versus plural misspelling

https://github.com/fusionpbx/fusionpbx/blob/f3986290eaf3e3b65b792da803e21ff4969c0113/core/email_templates/app_defaults.php#L28


markjcrane:
I'll accept it but I'm also going to change it. The reason is no other feature uses category: plugin.

PR #7294, opened 2025-03-07
add new active calls permissions

chansizzle:
Add new permissions to the active calls page so the fields can be removed from an admin permissions group to simplify the page, hiding more technical fields they probably will not find helpful.

These are all of the fields that can now be hidden, including eavesdrop and hangup, which were already wrapped in permissions. This pull request wraps **Profile**, **Application**, **Read/Write Codec**, and **Secure** fields in permissions.
Untitled


chansizzle:
A screenshot of these new permissions removed from an admin permission level group. A simplified view for less technical experienced users.

Firefox_Screenshot_2025-03-07T00-46-12 394Z

markjcrane:
Add the new permissions to the app_config.php file

chansizzle:
somehow the first file edited did not get included in the pull request. :face_exhaling:  This pull request has the permissions. 

https://github.com/fusionpbx/fusionpbx/pull/7295

PR #7301, opened 2025-03-11
Add CDR Details, Call Recordings Play, and Call Recordings Download to emergency logs

chansizzle:
This pull request adds 2 new fields to the emergency logs application. There is a hyperlink to open the CDR details page in a new browser tab. There is also a set of call recordings Play and Download buttons field. Both of the new fields and buttons are only available to the user if they have the associated permissions assigned to their account.

If the CDR record has been deleted, then there will be no CDR details hyperlink. If there is no call recording, then there will be no call recording play and download buttons.

Firefox_Screenshot_2025-03-11T16-40-06 573Z


markjcrane:
A good idea. However, recommend taking design queues from the Call Detail Records list. 
- The column CDR Details is not needed.
- Use the Status column to link to the Details. 
- The recording should be before Call Status.

chansizzle:
Combining the Call Status and the CDR Details fields is a good idea. The new changes place the Recording field before the Call Status column and the Call status row text has been changed to include the hyperlink to the CDR Details page.

Code for the Recording was copied from the Call Recordings page with small adaptations for the UUID.

Firefox_Screenshot_2025-03-11T23-26-36 734Z

markjcrane:
What branch did you copy the recording code from? Master or 5.3?

chansizzle:
master. I only touch master

PR #7312, opened 2025-03-13
fix typo in setting vs settings variable

frytimo:


frytimo:
This was fixed with commit b5592e84c79d7255e6a2842771525d8cf29ff593.

PR #7316, opened 2025-03-14
Caller id dashboard card

n0obHere:
Corrected the column names for caller ID name and number on the dashboard caller ID card, to match whats on the extensions page


markjcrane:
I used the labels that were used to keep the column names short.

n0obHere:
@markjcrane But doesn't the column name 'Destination' for caller ID number sounds incorrect.

PR #7324, opened 2025-03-18
bug fix for voicemail edit transcription

tony1661:
Voicemail transcription in voicemail_edit.php is not visible despite having the correct permissions. This is a fix for that.


markjcrane:
The settings object was added to require.php so its already available

PR #7357, opened 2025-04-24
Update music_on_hold.php

graycom-git:
Changes made stops global MOH categories being visible to non SuperAdmins


markjcrane:
Its better to add a permission and use a permission than a group. Then assign that permission to the superadmin group.

PR #7359, opened 2025-04-27
Fix: Ensure proper absolute_codec_string for fax_queue

hacc1225:
The `absolute_codec_string` parameter for FreeSWITCH was being incorrectly formatted, causing issues with fax transmissions, particularly in regions preferring the PCMA codec.

Without proper quoting, the comma-separated codec list `PCMU,PCMA` was being parsed incorrectly by FreeSWITCH. The second codec `PCMA` was misinterpreted as a separate dial string parameter, resulting in `absolute_codec_string` being set to only `PCMU`. This led to negotiation failures when the receiving end required PCMA.

This commit resolves this by enclosing the codec list within *escaped* single quotes `\'PCMU,PCMA\'`. **Crucially, without this escaping, FreeSWITCH would not correctly interpret the quoted string.** The literal quotes are necessary to ensure that `PCMU,PCMA` is treated as a single value.

By implementing this change, we guarantee that both PCMU and PCMA are correctly included in the codec negotiation, resolving the fax transmission failures in PCMA-dominant regions.

### DIFF
--- a/app/fax_queue/resources/job/fax_send.php
+++ b/app/fax_queue/resources/job/fax_send.php
@@ -396,7 +396,7 @@
 			//$dial_string .= "fax_retry_sleep=180,";
 			$dial_string .= "fax_verbose=true,";
 			//$dial_string .= "fax_use_ecm=off,";
-			$dial_string .= "absolute_codec_string=PCMU,PCMA,";
+			$dial_string .= "absolute_codec_string=\'PCMU,PCMA\',";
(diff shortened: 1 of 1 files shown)


markjcrane:
Please sign up as a contributor here and your pull request will be accepted.

hacc1225:
> Please sign up as a contributor here and your pull request will be accepted.
> 
>     * https://github.com/fusionpbx/open-source

OK, I have already done that.

PR #7363, opened 2025-05-06
Add friendly names for system services

n0obHere:
- Adds a $service_friendly_names array at the top of the details block
- For each service, tries to find a match in the array and sets the $display_name
- If no match found in the array, it falls back to prettifying the name automatically (replace underscore with space and capitalize the first letter of each word) so future services don't break the display


n0obHere:
Closed in favor of [7367](https://github.com/fusionpbx/fusionpbx/pull/7367)

PR #7389, opened 2025-06-18
Add a domain dropdown that allows you to only enter the subdomain and…

eliweaver732:
… select the parent domain from the dropdown. The list contains all the parent domains already in FusionPBX. You can add more parent domains by simply selecting --manual entry-- and entering the complete domain name. 

The purpose is to prevent misspellings when long or multiple domain names are in use and to save time when adding a domain.


markjcrane:
I haven't decided yet on this. I think it would be better to make it an optional feature. Another option could be to check the spelling against other domains that already exist.

However, notice that your code was probably not up to date when this change was made. Your code replaced this in favor of $_SESSION (now considered the old way to do it). The new code uses the $settings object.

$settings->get('theme', 'button_icon_export'),

eliweaver732:
Fair enough. 
I've committed several changes as required. I've implemented a default setting that can be used to turn this option on. Not sure where to add the default setting so it gets added to the database via upgrade so I skipped that step.

Even if you do not accept this commit I will not be offended. It works well for us and I can patch future versions with our specific changes. Please close this PR if you don't think this is a positive benefit to the community.

PR #7687, opened 2026-01-06
Reset cache after deleting domains

frytimo:
Added new code to:
- Flush all cache after deleting domains (filesystem, apcu).


markjcrane (commented):
Re-add the persistformvar if statement
- Save this change for the next commit.
- By doing this in two separate commits, it makes it easier to see what was changed.

PR #7728, opened 2026-01-29
Fix date-time conditions to respect domain timezone by converting to/from UTC

rabbidiesel:
Fixes date-time conditions to properly convert between the domain timezone and UTC.

Issue:
Date-time conditions triggered at the wrong times because FreeSWITCH evaluates them in UTC, but the user input was not being converted.

Solution:
    Convert user timezone → UTC when saving
    Convert UTC → user timezone when displaying
    Use settings object cascading for timezone (user → domain → global → default)
    Handle both 12h and 24h formats

Testing:
Set time condition for 9:00 AM in America/New_York timezone - should trigger at 9:00 AM local time, not UTC.


rabbidiesel (commented):
Changed to settings

PR #7854, opened 2026-04-14
Change provisioning to persistant

Ivane120:
Set app/phone to Provision on every startup


markjcrane:
What is the goal of this change?

Ivane120:
This option sets the linphone app to pull the provision every time it opens instead of just one time

PR #7928, opened 2026-04-28
Update HT814 provisioning template

Ivane120:


markjcrane:
This and some other pull requests remove this. Removing this is a behavior change and can disrupt how it works for people using it currently. This change doesn't look like it is backwards compatible.

    
    {if isset($grandstream_firmware_path) && isset($firmware_version)}
    <P192>{$grandstream_firmware_path}/gxw4004/{$firmware_version}</P192>
    {elseif isset($grandstream_firmware_path)}
    <P192>{$grandstream_firmware_path}/gxw4004</P192>
    {else}
    <P192>{$domain_name}{$project_path}/app/provision</P192>
    {/if}

    
    
    {if $grandstream_config_server_path=="none"}
    <P237></P237>
    {elseif isset($grandstream_config_server_path)}
    <P237>{$grandstream_config_server_path}</P237>
    {elseif isset($domain_name)}
    <P237>{$domain_name}{$project_path}/app/provision</P237>
    {/if}

Ivane120:
Okay, that's fine. I have noticed some do have the wrong model "gxw4004" instead of "ht814" as it was probably copied from another provision file.

markjcrane:
I noticed that today too so the model number does need to be fixed.

markjcrane:
I updated this template with this commit.
- https://github.com/fusionpbx/fusionpbx/commit/fd5eef4f16d7bb11c897713b189356545e1b63ad

PR #7938, opened 2026-04-29
Added Service Manager feature

tony1661:
Added the ability to see which FusionPBX systemd services are installed as well as see their status.

Status is colour coded for:
- Active
- Inactive
- Failed
- Activating
- Deactivating

### DIFF
--- a/app/service_manager/app_config.php
+++ b/app/service_manager/app_config.php
@@ -1,25 +1,4 @@
 <?php
-/*
-	FusionPBX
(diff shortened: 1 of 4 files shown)


markjcrane:
This is a great idea and is something that is needed. I looked at the code for a bit and decided that it would take longer to rewrite it. 

A similar feature has been committed and can be found in the menu at Advanced -> Services in the master branch and it will be part of the next release.

PR #7951, opened 2026-05-04
Fix an error when upgrade cli is used for first time with services

frytimo:


frytimo:
Fixed some PHP errors and warnings as well. Just variables that needed declaring before use and a variable that didn't have the '$' in front of it.

PR #7955, opened 2026-05-05
Add fanvil v6x templates

rabbidiesel:


markjcrane:
You need to make a new updated branch then added your changes to it that way you create a clean pull request. This way it will just show your changes. Rather than a bunch of conflicts and changes from other commits. Please try again.

PR #7960, opened 2026-05-07
Ring Groups: Add CSV import and export

rabbidiesel:
Adds Import and Export buttons on the ring groups list, modeled on the destinations import/export. New permissions ring_group_import, ring_group_upload and ring_group_export gate the feature. Import supports a two-step CSV mapping flow with Add or Delete actions, and multiple destinations per ring group via a configurable separator. Export emits a CSV with all ring_group columns plus a synthetic destinations column joined with '|'.


rabbidiesel:
Added confirm prompt

PR #8012, opened 2026-06-02
Add T38 variables to Inbound routes

eliweaver732:
### What
When an inbound destination has it's `Fax` type checkbox activated, we need to add t38 variables to the inbound dialplan. Currently we have to do this manually and resaving a destination wipes this for us.

### How
When the type is Fax for a destinations it adds these 2 variables to the inbound route in the `public` context:
* fax_enable_t38=true
* t38-passthrough=once


eliweaver732:
Closeing this as we no longer need it. We can do this another way.

PR #8045, opened 2026-06-20
Restrict call recording download to the current domain

9cb14c1ec0:
xml_cdr::download() looked up the recording by xml_cdr_uuid alone, with no domain_uuid filter, so a user with xml_cdr_view could download a recording belonging to another domain by supplying its uuid. Add the domain_uuid restriction unless the user has xml_cdr_all, matching the existing behavior in xml_cdr_details.php.


markjcrane:
It's extremely unlikely anyone could guess a UUID version 4 ID. However, someone may find a way to obtain the UUID and abuse it. Accepting this change.

PR #8046, opened 2026-06-20
update missing paths for php systemd file

frytimo:
Special considerations for the php-fpm.service file:
- It has to exist in the /usr/lib/systemd/system folder.
- We only change the version of the file that matches the currently running PHP version.
- If the file exists but is missing the ReadWritePaths directive then we add it with the required paths and in the correct section.
- If the file exists and has the ReadWritePaths directive but is missing any of the required paths then we add the missing paths to the end of the existing ReadWritePaths directive.
- Sections can be in any order and there can be multiple sections. So we make sure to add the directive in the [Service] section.
- An undefined section can exist before the first detected section. These are usually comments but can also be something else.

### DIFF
--- a/core/upgrade/upgrade.php
+++ b/core/upgrade/upgrade.php
@@ -583,14 +583,45 @@
  */
 function update_php_fpm(settings $settings) {
 
+	//
+	// Special considerations for the php-fpm.service file:
+	//   - It has to exist in the /usr/lib/systemd/system folder.
+	//   - We only change the version of the file that matches the currently running PHP version.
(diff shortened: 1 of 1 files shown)


markjcrane:
Sorry, I went back in a little different direction and used the original code. Then updated it with more error handling, a preg_replace for the entire line to ensure all ReadWritePaths are included in the PHP FPM service file.
https://github.com/fusionpbx/fusionpbx/commit/efea1bafb0cf23801aeb108dbcb57deaae3eb62d

PR #8048, opened 2026-06-22
Fix file upload path traversal issues

9cb14c1ec0:
Wrap basename() around ['ulfile']['name'] in 6 CSV import files to prevent directory traversal via malicious filenames.

Affected files:
- core/contacts/contact_import.php
- app/voicemails/voicemail_imports.php
- app/extensions/extension_imports.php
- app/access_controls/access_control_import.php
- app/destinations/destination_imports.php
- app/bridges/bridge_imports.php

The affected functionality requires admin and/or superadmin permissions to reach, but even in the case of superadmin we want to protect against a compromised superadmin account escalating into raw system access.


9cb14c1ec0:
Actually, not sure this makes sense.  Superadmin can inject shell commands through xml dialplans, so trying to harden the superadmin accounts for account compromise situations probably isn't viable.

core/contacts/contact_import.php is the only one that can be reached by non-superadmin accounts, so your call if this is worth merging or not.

markjcrane:
FreeSWITCH has an option to disable the shell commands. So even if they are in the dialplan, they won't run.

PR #8053, opened 2026-06-23
Add dialed number to call recording transcription email

elvinlh:
This updates the default call recording transcription email template to include destination_number.

For outbound calls, the existing transcription email shows the internal caller ID but not the external number that was dialed. Adding the dialed number makes outbound call transcription emails easier to identify.

Change:
- Adds Dialed Number to the call recording transcription email template body.

Example:
Caller ID OFFICE 109
Dialed Number 4805862496

### DIFF
--- a/app/call_recordings/app_defaults.php
+++ b/app/call_recordings/app_defaults.php
@@ -52,7 +52,8 @@ if ($domains_processed == 1) {
 	$array['email_templates'][$x]['template_body'] .= "	<body>\n";
 	$array['email_templates'][$x]['template_body'] .= "	Caller ID \${caller_id_name} <a href=\"tel:\${caller_id_number}\">\${caller_id_number}</a><br />\n";
 	$array['email_templates'][$x]['template_body'] .= "	<br />\n";
-	$array['email_templates'][$x]['template_body'] .= "		Date \${start_date}<br />\n";
+    $array['email_templates'][$x]['template_body'] .= "        Dialed Number <a href=\"tel:\${destination_number}\">\${destination_number}</a><br />\n";	
+    $array['email_templates'][$x]['template_body'] .= "		Date \${start_date}<br />\n";
 	$array['email_templates'][$x]['template_body'] .= "		Time \${start_time} \${end_time}<br />\n";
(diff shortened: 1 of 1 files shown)


elvinlh:
Just following up on this PR.

This is a small default email template change for call recording transcriptions. It adds `${destination_number}` to the transcription email body so outbound call transcription emails can show the external number that was dialed.

Please let me know if any changes are needed.

elvinlh:
I received feedback that using the label “Dialed Number” may not be ideal because `${destination_number}` may display differently for inbound calls.

Would “Destination Number” be a better label so the template works for both inbound and outbound call transcription emails?

Proposed line:

```php
$array['email_templates'][$x]['template_body'] .= "        Destination Number <a href=\"tel:\${destination_number}\">\${destination_number}</a><br />\n";
```

My understanding is:

Inbound call:

* Caller ID = outside caller
* Destination Number = number/extension/DID the caller reached

Outbound call:

* Caller ID = internal extension/user placing the call
* Destination Number = outside number that was called

Please let me know if this is the preferred approach, or if there is a better variable/label to use for both inbound and outbound calls.

markjcrane:
Please sign up as a contributor here. So that I can accept your pull request. It needs to be done today or tomorrow, or I will write the changes myself to complete the feature.

You are right; the destination_number is probably not the right variable. I changed the variable to caller_destination. I made that variable because it seemed logical to follow caller_id_name and caller_id_number.

This also required a code update, which I did here.
https://github.com/fusionpbx/fusionpbx/commit/2a029cd4eb2910a3daaff1d1c334a6fdd7974c6b

elvinlh:
I submitted the contributor agreement PR here:

fusionpbx/open-source#335

I understand that caller_destination is the better variable for this use case instead of destination_number. Please let me know if anything else is needed from me.

markjcrane:
Looks good. I have accepted your pull request.

elvinlh:
Okay great, when will it be deployed or active?

elvinlh:
Thank you for accepting the pull request and updating the implementation to use caller_destination.

Our PBX provider said they will not deploy this change until it is included in a stable FusionPBX release.

Do you know whether this change will be included in the next stable release, and is there any estimated timing for that release?

PR #8062, opened 2026-06-27
Replace shell command pipelines with PHP curl in transcribe.php

9cb14c1ec0:
Refactor Google V1/V2, Azure, and Custom transcription providers to use the PHP curl library instead of calling curl with `shell_exec()`.  This eliminates several shell command injection vulnerabilities (where user-configurable settings like `transcribe_language`  were concatenated raw into shell commands) that could lead to privilege escalation.


markjcrane:
Accepted your change; however, this file is no longer used. At some point in 2024, it was replaced by the app/transcribe feature located here.
[https://github.com/fusionpbx/fusionpbx-app-transcribe](https://github.com/fusionpbx/fusionpbx-app-transcribe)

The new code uses a transcribe PHP class instead of a transcribe function.

PR #8064, opened 2026-06-29
Add Advanced Call Distribution (ACD) app

teknoprep:
## Advanced Call Distribution (ACD)

A lightweight, multi-tenant call-queue / contact-distribution app for FusionPBX that drives FreeSWITCH directly from PostgreSQL + Lua. It is an alternative to `mod_callcenter`: no queue state is held in FreeSWITCH memory — all state (queues, members, login status, live sessions) lives in the database, so login/logout is instant and everything is inspectable in SQL.

### Features
- **Queues** whose extension is selectable as a destination everywhere (inbound routes, IVRs, time conditions, ring groups, …).
- **Tiered ringing** of logged-in agents with a configurable tier-advance time.
- Agents reached on **extension + follow-me**, including Microsoft Teams users. First to answer wins; the rest stop ringing.
- **Per-member Busy Handling**: Always Ring / Never Ring (Any Call) / Never Ring (Queue Calls Only).
- **Self-service login/logout** feature codes: `*86` (all queues) and `*86<extension>` (single queue).
- **Live dashboard**: waiting callers, connected calls with talk time, and ringing agent legs.

### FreeSWITCH modules
`mod_lua`, `mod_dptools` (incl. `intercept`), `mod_commands`, `mod_loopback`, `mod_sofia`, `mod_local_stream`, `mod_dialplan_xml`, `mod_event_socket`, `mod_sndfile`.

### Self-installing
- `app_config.php`: schema (`v_acd_*`), permissions (`acd_*`), default settings, and the queue **destination type**.
- `app_menu.php`: menu entry.
- `app_defaults.php`: creates the per-domain `*86` feature-code dialplan (idempotent).
- Per-queue dialplan entries are generated automatically when a queue is saved.

### Layout
- Web app: `app/acd/`
- FreeSWITCH Lua: `app/switch/resources/scripts/app/acd/` (installs to `scripts/app/acd/`)


frytimo:
Great work! In my humble opinion, the ability to have the call center work without the module is really good. The time and attention to detail you have given for the needs of the agent are well thought out for real-world work flows.

Upon code review, I noticed in the code there are many `new database` commands. This is no longer standard in the framework as it causes a new connection to the database with each request of a new database object. This can cause issues because it can reach the maximum number of database connections much faster. A direct replacement is to use the `database::new()` command instead which uses a singleton pattern to re-use the same object and connection. One other thing I noticed is the `app_uuid` and `app_name` properties of the acd object are set to private but should now be declared as a constant in the object. This allows the database transactions to automatically detect the app used when changing table data and is now standard in the framework.

teknoprep:
Just updated based on what you asked for

markjcrane:
This would need to be committed to its own repo. If we add it to a new repo called fusionpbx-app-acd, would you commit it there?

PR #8068, opened 2026-07-01
Add W60B Music on Hold provisioning settings

cranben:
Adding SIP hold settings to the W60B template so cordless handsets use PBX-hosted Music on Hold instead of local handset hold behavior.


cranben:
Deployed to a customer system and tested successfully. Verified that callers now receive PBX-hosted Music on Hold when calls are placed on hold from Yealink W60B cordless handsets.

PR #8123, opened 2026-08-07
Clean up provisioning template

Ivane120:
Mapped variables


Ivane120:
Current file seems to be old format, Updated to newer template

PR #8153, opened 2026-08-27
Update xml_cdr.php

fsilvestre:
ob_clean() causes race conditions while fopen read file. Storing ob_clean() return guarantee that finished and fopen will work without interference


markjcrane:
I appreciate your willingness to contribute.

I checked with multiple Large Language Models to get additional insight. It might be a good idea for you to do so as well and see if you still think the code change is needed.

The ones I checked were
Nemotron, Ministral, Qwen, and Gemma

My response here is not meant to discourage. I'm only reviewing the code change.

PR #8163, opened 2026-08-28
Expand paging and intercom application

denisent:
## Summary

- Add paging mode types: one-way (page), two-way (intercom)
- Adds announcement options, announce recordings or sounds
- Add a configurable 400-499 extension range default that is disabled by default


markjcrane:
A few Guidelines for contributing

**Application Notes**

I'm not certain join sound is a good feature for paging. The reason why is that you will hear one for every person that is added to the paging/intercom. I had a lot of complaints about this in the past. FreeSWITCH would play each sound one at a time. So if you have 30 people, it plays the join tone 30 times before people can talk after the call is answered.

**Development notes**

The app_config.php file should look like other FusionPBX app_config files that include MySQL and SQLite.

In table and in the interface, and in the code, the order of the variables used in the code should be enabled second-to-last and description last.

The files app_config.php, app_menu.php, app_defaults.php don't have any mentions of contributors. Git does show what lines were contributed and who did them. This is project-wide; if I find them, they are removed, including in the current codebase.

The app_languages.php file should be translated and not set to English for every language. Because AI has improved at translations, this is easy to do.

Look at other files in FusionPBX at the top, just below the license, to find the contributors list. They will not be accepted in a different location than the standard way we are already doing it.

The database connection is already there by default; this isn't needed.
```
$database = new database; 
```

However, if it were to be added, then it should be done this way.
```
//connect to the database
	$database = database::new();
```
	
The database save method adds insert_date and insert_user automatically for you when a record is added. It also updates the update_date and update_user for you without extra code.

The dialplan_enabled boolean, other boolean fields save to the database array as a string, and PostgreSQL converts it to automatically to the correct boolean value. So dialplan_enabled 'true' and 'false' is the correct value.

This should be replaced
```
$array['dialplans'][0]['dialplan_enabled'] = ($paging_group_enabled === true || $paging_group_enabled == 'true' || $paging_group_enabled == 't' || $paging_group_enabled == '1') ? 'true' : 'false';
```
Corrected version: the value should be 'true' or 'false' as a string
```
$array['dialplans'][0]['dialplan_enabled'] = $paging_group_enabled
```

Temporary permission should be removed after the database->save method. Temporary permissions are for tables like the dialplan where a user will not have that permission, such as dialplan_add and dialplan_update. The following are not needed.
```
$p->add('paging_group_add', 'temp');
$p->add('paging_group_edit', 'temp');
$p->add('paging_group_destination_add', 'temp');
$p->add('paging_group_destination_edit', 'temp');
$p->add('paging_group_destination_delete', 'temp');
```

Deleting a record should follow how it's done in other code in FusionPBX.

The update command should instead use the database save method that uses the array. The database save method can be used for both insert and update, and so the SQL insert and update should not be used directly.

The \t should be replaced with a tab.

This is what I have found with a short review; there is probably more.

PR #8183, opened 2026-09-04
update system permissions to use unique UUIDs

chansizzle:
Updated UUIDs for various permissions in app_config.php.


markjcrane:
The menu_uuid is supposed to reference an existing menu. Are you pointing these to existing menus?

PR #8184, opened 2026-09-04
Update system_status_sofia_status_profile permission to use unique UUID

chansizzle:


markjcrane:
The menu_uuid is only useful if it points to an existing menu.

PR #8186, opened 2026-09-08
Fix PHP error and deprecated warnings for themes class

frytimo:
Updated the class with the following changes:
- Removed check for class exists as the auto_loader class handles this automatically
- Replaced the $parameters variable with an empty array as PHP gives a warning for undefined variable (not initialized)
- Add the database and tables properties to the object as this causes a warning in PHP and a fatal error in PHP 8.5
- Added minimal documentation to the functions to match current class standard
- Removed closing tag on class to match current class standard
- Changed the "new database" syntax to use existing database reference in the object instead
- Removed setting app_uuid and app_name when calling database->save in delete method to match other methods


frytimo:
Closed - Files are supposed to be separate pull requests for different issues. Creating new pull request for the individual changes.

PR #8225, opened 2026-09-21
FIFO: remove stale members from mod_fifo when members or the queue extension change or are deleted

apurboislam:
## Problem

When a FIFO queue member's contact is changed (e.g. `user/101@example.com` → `user/100@example.com`) and saved, the new member is added to mod_fifo but the old one is never removed. The queue keeps ringing the old extension, even though the database and the UI only show the new one.

`reloadxml`, `fifo reparse`, reloading mod_fifo and restarting FreeSWITCH don't clear it.

### Steps to reproduce

1. Create FIFO queue `3001` with member `user/101@example.com` (enabled) and save.
2. Edit the queue, change the member to `user/100@example.com`, and save.
3. In fs_cli run:
```
   fifo list_verbose 3001@example.com
```
4. Both `user/101@example.com` and `user/100@example.com` are listed as outbound members. Dialing 3001 still rings 101.

The same leak happens when:
- a member row is deleted with the **Delete** checkbox,
- the queue's **Extension** is changed (members stay registered under the old `<extension>@domain` queue), and
- a queue is deleted from the FIFO list page.

## Cause

`app/fifo/fifo_edit.php` only sends `fifo_member add` / `fifo_member del` for the rows in the submitted form. It never compares against the previously saved contacts or extension, so a changed contact never gets a `del`. The member delete action in `fifo_edit.php` and `fifo::delete()` in `app/fifo/resources/classes/fifo.php` only delete database rows.

mod_fifo keeps members added with `fifo_member add` as dynamic members (`static=0`) in its `fifo_outbound` table. `load_config()` in `mod_fifo.c` only deletes `static=1` rows on reload and on startup, unless `fifo reparse del_all` or `delete-all-outbound-member-on-startup=true` is used. So stale members stay there until someone removes them by hand.

## Fix

- **`fifo_edit.php` (save):** before saving, load the old `fifo_extension` and `member_contact` values from the database and send `fifo_member del` for every old contact that's no longer in the form. If the extension changed, send it for all old contacts. Members that didn't change aren't touched, so their mod_fifo statistics aren't reset. The existing add/del loop then runs as before.
- **`fifo_edit.php` (member delete action):** send `fifo_member del` for the checked members before deleting the rows.
- **`fifo::delete()` (`classes/fifo.php`):** send `fifo_member del` for all members of the deleted queues before deleting the rows.

All commands are only sent when the event socket is connected, and they use the same originate string as the add command (`{fifo_member_wait=nowait}` + `member_contact`), because mod_fifo finds members by an MD5 of that string. `member.lua` (the agent login/logout feature code) already adds and removes members with matching strings and is unchanged.

## Testing

Tested on a live system:

1. Created queue 3001 with member 101 and confirmed it appears in `fifo list_verbose 3001@example.com`.
2. Changed the member to 100 and saved: only 100 is listed.
3. Changed the extension to 3002 and saved: `3001@example.com` has no members and `3002@example.com` has 100.
4. Deleted member 100 with the checkbox: it's gone from mod_fifo.
5. Deleted the queue from the list page: its members are gone from mod_fifo.

## Workaround for existing stale members

This PR prevents new stale members. Members that are already stuck can be removed once in fs_cli:

```
fifo_member del <ext>@<domain> {fifo_member_wait=nowait}user/<member_ext>@<domain>
```


markjcrane:
Please sign up as a contributor here, and I will accept your pull request.

PR #8227, opened 2026-09-21
Fix clear destination ringback when empty

frytimo:


frytimo:
Accidental duplicate PR

PR #8232, opened 2026-09-22
Fix OpenID login buttons rendering twice on the login form

rabbidiesel:
Commit f1849bd8a added a second copy of the OpenID banner loop inside the login form next to the passkey button, but unlike the passkey button it had no inside/outside guard, and the original copy below the form was kept. Every enabled OpenID provider (Google, Microsoft, Okta) was therefore shown twice on the login page.

Make both banner blocks follow the same login -> passkey_position switch the passkey button already uses, so the providers render once, inside the form by default or below it when the position is set to outside.


markjcrane:
Update login settings options position
- Rename setting from passkey_position to options_position
- Rename banners to login_options
- Rename banner to login_option
https://github.com/fusionpbx/fusionpbx/commit/c9742dfda7002bc0043aa9546a54a99f52b625a0

This is what I did instead. Now there is a setting called  login options_position. This makes it so openid is part of the login options and allows the position to work for passkey and the other options.

PR #8234, opened 2026-09-23
Fix call center agent contact for gateway agents with custom variables

paulobautz:
## Problem

In `callcenter.conf.lua`, when an agent contact **already has a `{...}` variable block** and uses **`sofia/gateway`**:

1. **With `call_timeout` in the block**, the code uses `tmp_first` / `tmp_last`, which are never defined:
   ```
   callcenter.conf.lua:290: attempt to concatenate a nil value (global 'tmp_last')
   ```
   The error aborts the `dbh:query` agents loop. **Every agent returned after that row is left out of the generated `callcenter.conf`**, and they don't load on `reload mod_callcenter`. Which agents disappear depends on row order, so the effect looks random.

2. **Without `call_timeout`**, mixed single and double quotes write the literal text into the contact:
   ```
   ...,domain_name="..domain_name..",domain_uuid="..domain_uuid..",...
   ```

## Fix

Use `first` / `last`, which are set a few lines above in the same branch, and fix the quoting. Line 287 also drops the duplicated `sip_h_caller_destination`.

## Test

The contact-building block was extracted and run with 6 contact shapes (no `{}` local/gateway, `{}` local with/without `call_timeout`, `{}` gateway with/without `call_timeout`):
- **Before:** the `{}`+gateway without `call_timeout` case writes the literal `"..domain_name.."`, and the `{}`+gateway with `call_timeout` case raises the nil error above.
- **After:** all 6 build valid contacts. `luac -p` is clean.

Found in production on 5.6: two gateway agents with `{group_confirm_key=1,...,call_timeout=30}sofia/gateway/...` silently removed all agents created after them from the XML.


markjcrane:
Please sign up as a contributor here and your pull request will be accepted.

PR #8238, opened 2026-09-23
Fix duplicate banner block for open_id_connect

frytimo:


markjcrane:
Update login settings options position
- Rename setting from passkey_position to options_position
- Rename banners to login_options
- Rename banner to login_option
https://github.com/fusionpbx/fusionpbx/commit/c9742dfda7002bc0043aa9546a54a99f52b625a0

This is what I did instead. Now there is a setting called login options_position. This makes it so that OpenID is part of the login options and allows the position to work for passkey and the other alternate options.

